PR #8, opened 2023-07-26
feat: add `command` operation

fzdwx:


sxyazi:
It doesn't seem to wait for the command to finish, nor output the results back. I'm wondering what real-world usage it would be? 🤔

fzdwx:
提供一种可以不退出 `yazi` 并执行 shell script 的方式

---

Provide a way to execute shell scripts without exiting `yazi`.

sxyazi:
一般是执行哪些命令？

---
What commands would be executed in general?

fzdwx:
比如 `code .` `idea .`, or `lazygit`...  后面也可以出一个 popup view 来显示类似 `lazygit` 这种 tui 的 程序

sxyazi:
这些用法和现有的 `open_with` 功能重叠，只是操作对象是目录，而非普通文件，并且 `open` 有更完善的功能，比如运行中的任务会出现在 `tasks` 列表，可以取消它们；比如打开 `lazygit`、`nvim` 这种需要交互的应用。我想应该将 `open` 的作用范围扩大到涵盖目录。

---
These usages overlap with the existing `open_with` command, but the difference is that the target are directories instead of regular files. Also, `open` provides more comprehensive features, such as displaying running tasks in the tasks list, and the ability to cancel them. And, it can be used to launch interactive applications like `lazygit` or `nvim`. I believe the scope of `open` should be expanded to include directories.

fzdwx:
Yes, `open` should include dir

sxyazi:
Done by https://github.com/sxyazi/yazi/pull/9

PR #18, opened 2023-08-03
Add pdf preview with pdftotext

Eric-Song-Nop:
Add functionality to preview pdf as text with `pdftotext`, I am not sure how it behaves on a system without `pdftotext`. The command line arguments are borrowed from ranger.


sxyazi:
Thanks for the PR!

I'm also thinking of adding PDF preview functionality to yazi, via `pdftoppm` -- also called `poppler`. I think it's a much cooler tool than `pdftotext`, which allows displaying the cover of a PDF as an image and is much more expressive.

image

(The image taken from [Telescope-media-files.nvim](https://github.com/nvim-telescope/telescope-media-files.nvim))

Displaying images is not an issue, yazi aims to provide a more modern interactive experience with multiple image protocols built-in and covering most terminals. And with the pre-caching mechanism, rendering PDF images will not take much time. 😆

PR #26, opened 2023-08-06
fix: tasks number format

fzdwx:


sxyazi:
Thanks! I didn't even catch this..

PR #30, opened 2023-08-07
feat: allow to customize status section separator

g-plane:


sxyazi:
Great and thanks!

- Can you rename `section_separator` to `separator` to keep the name as short as possible?
- What do you think about renaming `left` and `right` with `opening` and `closing` to preserve the original semantics?

g-plane:
Done.

sxyazi:
LGTM, thanks!

PR #31, opened 2023-08-07
feat: add name to opener and show opener's args

XYenon:


sxyazi:
I think the `name` may be confused with the `command name`, can you change it to `display`, or something else?

XYenon:
> I think the `name` may be confused with the `command name`, can you change it to `display`, or something else?

I feel `display` confused too. What about `display_name` or `description`?

sxyazi:
`display_name` should make sense. 😆

sxyazi:
Thanks, merged!

PR #32, opened 2023-08-07
refactor: avoid hard coding tmp dir

g-plane:
Avoid hard coding `/tmp`, which makes cross-platform friendly.


sxyazi:
Thanks, merged!

PR #33, opened 2023-08-08
chore: rename bin

g-plane:
We can use `./target/{debug, release}/yazi` now.


sxyazi:
Thanks, that makes sense. I'm wondering if that will affect the current packaging of Nix? @XYenon

XYenon:
> Thanks, that makes sense. I'm wondering if that will affect the current packaging of Nix? @XYenon

Yeah I can remove `mv $out/bin/app $out/bin/yazi`.

sxyazi:
Okay, merged!

PR #35, opened 2023-08-08
refactor: remove `static mut` usage

g-plane:
`static mut` should be avoided because of UB. This PR refactors this with `OnceCell`.


sxyazi:
Should be fixed by https://github.com/sxyazi/yazi/pull/63, closing then.

PR #37, opened 2023-08-08
feat: add args support for `shell`

fzdwx:
e.g:  `{ on = [ "l", "g" ], exec = "shell lazygit --block" } `

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -134,6 +134,7 @@ dependencies = [
  "tracing",
  "tracing-appender",
  "tracing-subscriber",
+ "unicode-width",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


fzdwx:
Yes, perhaps we can add an "args" to represent the parameters of lazygit, Can you give some suggestions?

`{ on = [ "l", "g" ], exec = "shell lazygit --block" , args = [xxx]}`

sxyazi:
I'm also still thinking about how to handle it, with the current idea being that it could be configured in the following format:

```toml
exec = [ "shell --block", [ "lazygit", "foo", "--bar" ] ]

# or
# exec = [ "shell", [ "lazygit", "foo", "--bar" ], "--block" ]

# or
# exec = [ "shell --block", [ "lazygit foo --bar" ] ]
```

where `[ "lazygit", "foo", "--bar" ]` as a whole is the second argument to `shell`. This changes the format of all `exec` support, not just `shell` commands. But I'm not sure what the flaws are with this, and it may need more discussion.

PR #39, opened 2023-08-09
feat: add interactive cd

XYenon:


core/src/manager/tab.rs
```diff
@@ -74,6 +74,25 @@ impl Tab {
 		true
 	}
 
+	pub fn cd_interactive(&mut self, interactive: bool, target: PathBuf) -> bool {
+		if interactive {
+			let path = target.to_str().unwrap_or_default().to_string();
```
sxyazi, line 79: Why not use `to_string_lossy` here? It's better to move the `target` to the async block since it has ownership.

sxyazi (changes requested):

core/src/manager/tab.rs
XYenon: Updated.

PR #41, opened 2023-08-09
feat: show current dir name in tab

g-plane:
This PR introduces a new option `max_width` in `[tab]` section for controlling the max width of current directory name in tab. Default value is `0` which means directory name won't be displayed in tab. This matches current behavior.

Here is an example for setting value to `10`:

image

The first tab's directory is "raffia" whose length is less than the threshold we set; the second tab's directory is "swc-plugin-react-remove-properties" whose length is greater than the threshold, so it's truncated and only the first 10 chars are displayed. (Should we add ellipses for this case?)

### DIFF
--- a/app/src/header/tabs.rs
+++ b/app/src/header/tabs.rs
@@ -19,11 +19,18 @@ impl<'a> Widget for Tabs<'a> {
 			tabs
 				.iter()
 				.enumerate()
-				.map(|(i, _)| {
+				.map(|(i, tab)| {
+					let mut text = format!(" {} ", i + 1);
+					if THEME.tab.max_width > 0 {
+						text
+							.push_str(&tab.current_name().chars().take(THEME.tab.max_width).collect::<String>());
+						text.push(' ');
+					}
+
 					if i == tabs.idx() {
-						Span::styled(format!(" {} ", i + 1), THEME.tab.active.get())
+						Span::styled(text, THEME.tab.active.get())
 					} else {
-						Span::styled(format!(" {} ", i + 1), THEME.tab.inactive.get())
+						Span::styled(text, THEME.tab.inactive.get())
 					}
 				})
 				.collect::<Vec<_>>(),
--- a/app/src/header/tabs.rs
+++ b/app/src/header/tabs.rs
@@ -1,5 +1,8 @@
+use std::ops::ControlFlow;
+
 use config::THEME;
 use ratatui::{buffer::Buffer, layout::{Alignment, Rect}, text::{Line, Span}, widgets::{Paragraph, Widget}};
+use unicode_width::UnicodeWidthStr;
 
 use crate::Ctx;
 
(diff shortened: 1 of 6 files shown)


core/src/manager/tab.rs
```diff
@@ -258,6 +258,10 @@ impl Tab {
 		};
 		true
 	}
+
+	pub fn current_name(&self) -> &str {
+		self.current.cwd.file_name().and_then(|name| name.to_str()).unwrap_or_default()
```
sxyazi, line 263: Using the `cwd` itself instead of the default might be better when can't get the filename.

app/src/header/tabs.rs
```diff
@@ -19,11 +19,18 @@ impl<'a> Widget for Tabs<'a> {
 			tabs
 				.iter()
 				.enumerate()
-				.map(|(i, _)| {
+				.map(|(i, tab)| {
+					let mut text = format!(" {} ", i + 1);
+					if THEME.tab.max_width > 0 {
+						text
+							.push_str(&tab.current_name().chars().take(THEME.tab.max_width).collect::<String>());
```
sxyazi, line 26: The number of chars and the display width are different, one Unicode char may take up 2 cells. it's better to use the `unicode-width` crate here.

config/preset/theme.toml
```diff
@@ -1,6 +1,7 @@
 [tab]
-active   = { fg = "#1E2031", bg = "#80AEFA" }
-inactive = { fg = "#C8D3F8", bg = "#484D66" }
+active    = { fg = "#1E2031", bg = "#80AEFA" }
+inactive  = { fg = "#C8D3F8", bg = "#484D66" }
+max_width = 0
```
sxyazi, line 4: I think `max_width` should represent the overall width of the tab (with default 1), not just the dir name, since it seems not related to the dir name by looking

sxyazi (changes requested):

config/preset/theme.toml
g-plane: I agree. Do you have any suggestions about the naming?

sxyazi: I prefer to keep this name, but change the behavior to fit it, so that it can cover what is added in the future not only the dir name currently.

core/src/manager/tab.rs
g-plane: Do you mean return `self.current.cwd`?

sxyazi: Like this https://github.com/sxyazi/yazi/blob/866ed1704459b91b7e6ea220e4aadcbd1397ba6d/config/src/pattern.rs#L23

app/src/header/tabs.rs
```diff
@@ -19,11 +22,33 @@ impl<'a> Widget for Tabs<'a> {
...
 				.iter()
 				.enumerate()
-				.map(|(i, _)| {
+				.map(|(i, tab)| {
+					let mut text = format!(" {} ", i + 1);
+					if let Some(dir_name) = tab.current_name() {
+						text.push_str(dir_name);
+					}
+
+					let threshold = THEME.tab.max_width.max(3) - 1;
```
sxyazi, line 31: It seems that `threshold` doesn't need to contain 2 spaces, as the spaces are `padding`, not `content`.

Currently, the spaces should always be present, and the user can be allowed to configure their own padding characters later on.

config/preset/theme.toml
```diff
@@ -1,6 +1,7 @@
 [tab]
-active   = { fg = "#1E2031", bg = "#80AEFA" }
-inactive = { fg = "#C8D3F8", bg = "#484D66" }
+active    = { fg = "#1E2031", bg = "#80AEFA" }
+inactive  = { fg = "#C8D3F8", bg = "#484D66" }
+max_width = 3
```
sxyazi, line 4: Should be set to `1`, i.e. `padding` is not included.

sxyazi (changes requested):

sxyazi:
Thanks for your work, merged!

PR #48, opened 2023-08-12
feat: add preset command to the `shell`

fzdwx:
previous issues #37 

support style:

```
{ on = [ "l", "g" ], exec = "shell --block lazygit" },


{ on = [ "l", "g" ], exec = "shell --block 'lazygit -foo bar'" },
```

### DIFF
--- a/app/src/executor.rs
+++ b/app/src/executor.rs
@@ -106,7 +106,10 @@ impl Executor {
 			}
 			"create" => cx.manager.create(),
 			"rename" => cx.manager.rename(),
-			"shell" => cx.manager.shell(exec.named.contains_key("block")),
+			"shell" => cx.manager.shell(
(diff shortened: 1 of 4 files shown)


sxyazi:
I think it would be better to always show the input so that the user can finish the final editing of the template code.

When the user does not need to edit, explicitly pass the `--confirm`, i.e. `shell "echo 123" --confirm`

fzdwx:
> I think it would be better to always show the input so that the user can finish the final editing of the template code.
> 
> When the user does not need to edit, explicitly pass the `--confirm`, i.e. `shell "echo 123" --confirm`

Good idea.

fzdwx:
cc @sxyazi

sxyazi:
Thanks, merged!

PR #49, opened 2023-08-12
[FEAT] Directory at top option

Eric-Song-Nop:
`sort_dir_first: bool` option added and supported to manager


core/src/files/files.rs
```diff
@@ -116,29 +116,40 @@ impl Files {
...
+				} else if !a.meta.is_dir() && b.meta.is_dir() {
+					return std::cmp::Ordering::Greater;
+				}
+				std::cmp::Ordering::Equal
+			});
+		}
+
+		if self.sort.reverse {
+			self.items.reverse();
+		}
```
sxyazi, line 151: Can you put both `dir_first` and `reverse` in the `cmp` function to make sure that it will only be 1-pass processed?

config/docs/yazi.md
```diff
@@ -14,6 +14,11 @@
   - `true`: Reverse order
   - `false`: Normal order
 
+- sort_dir_first: Display files in reverse order
+
+  - `true`: Directory at first
+  - `false`: Sort According to `sort_by`
```
sxyazi, line 20: Does "According to `sort_by` only" sound better?

config/preset/yazi.toml
```diff
@@ -2,6 +2,7 @@
 sort_by      = "modified"
 sort_reverse = true
 show_hidden  = false
+sort_dir_first = true
```
sxyazi, line 5: It's better to move this line up one, so that the `sort_*` is in the same group.

sxyazi (changes requested):

core/src/files/files.rs
Eric-Song-Nop: I am afraid that the code might be a little more ugly of I put everything in cmp, maybe I should find another way to make it one pass.

config/preset/yazi.toml
Eric-Song-Nop: Of course!

core/src/files/files.rs
sxyazi: Let me try it now!

Eric-Song-Nop: Quick note, I just found that cmp::Ordering has a reverse() method, like `std::cmp::Ordering::Less.reverse()` |-> `cmp::Ordering::Greater`, I guess you would need it.

Eric-Song-Nop: I have updated my codes

sxyazi:
Thanks, I've simplified the code, merged!

PR #50, opened 2023-08-12
feat: bulk rename

TD-Sky:


shared/src/temp_path.rs
```diff
@@ -0,0 +1,36 @@
...
+	let mut name = String::with_capacity(FILE_PREFIX.len() + ID_LEN + suffix_len);
+	name.push_str(FILE_PREFIX);
+	rand::thread_rng().sample_iter(&Alphanumeric).take(ID_LEN).for_each(|c| name.push(c as char));
+
+	if !ext.is_empty() {
+		name.push('.');
+		name.push_str(ext);
+	}
+
+	let tmp = std::env::temp_dir();
```
sxyazi, line 23: Instead of `temp_dir`, you can use `BOOT.cache_dir`, which is automatically created at booting.

shared/Cargo.toml
```diff
@@ -8,5 +8,6 @@ anyhow      = "^1"
 crossterm   = "^0"
 libc        = "^0"
 parking_lot = "^0"
-ratatui  		 = { version = "^0" }
+rand        = "^0"
```
sxyazi, line 11: How about using the md5 with the current time? That way don't need to introduce an extra crate

core/src/manager/folder.rs
```diff
@@ -115,12 +115,11 @@ impl Folder {
 	pub fn select(&mut self, idx: Option<usize>, state: Option<bool>) -> bool {
 		let len = self.files.len();
 		let mut apply = |idx: usize, state: Option<bool>| -> bool {
-			if state.is_none() {
+			let Some(state) = state else {
```
sxyazi, line 118: Good idea! 👍🏻

sxyazi (changes requested):

shared/Cargo.toml
TD-Sky: Simply use the timestamp, no need for md-5.

PR #67, opened 2023-08-17
feat: show symlink

XYenon:


app/src/manager/folder.rs
```diff
@@ -80,7 +80,13 @@ impl<'a> Widget for Folder<'a> {
 					self.file_style(v)
 				};
 
-				ListItem::new(format!(" {icon} {}", readable_path(k, &self.folder.cwd))).style(style)
+				let display_path = readable_path(k, &self.folder.cwd);
+				let display_full = if MANAGER.show_symlink && v.is_link {
+					format!(" {icon} {} -> {}", display_path, v.link_to.clone().unwrap_or_default().to_string_lossy())
```
sxyazi, line 85: Seems there is an unnecessary clone?

config/preset/yazi.toml
```diff
@@ -3,6 +3,7 @@ sort_by        = "modified"
 sort_reverse   = true
 sort_dir_first = true
 show_hidden    = false
+show_symlink   = false
```
sxyazi, line 6: `show_symlink = true` is just fine for most cases

app/src/manager/folder.rs
XYenon: It will be an error without clone.
image

sxyazi: I'll fix it.

PR #83, opened 2023-08-24
Organize keys for logical behavior

asim215:
- [x] Organize keys on sorting: 1) lower key always normal sort; 2) upper key always _reverse_ sort.
- [x] Add comment to improve clarity on _close_ command.
- [x] ~Change on _redo_ on more logical "U" instead of Ctrl+r (similar behavior with helix). Less keys to type.~

### DIFF
--- a/config/preset/keymap.toml
+++ b/config/preset/keymap.toml
@@ -176,5 +176,5 @@ keymap = [
 
 	# Undo/Redo
 	{ on = [ "u" ], exec = "undo" },
-	{ on = [ "U" ], exec = "redo" },
+	{ on = [ "<C-r>" ], exec = "redo" },
 ]
--- a/config/preset/keymap.toml
+++ b/config/preset/keymap.toml
@@ -3,6 +3,7 @@
 keymap = [
 	{ on = [ "<Esc>" ], exec = "escape"  },
 	{ on = [ "q" ], exec = "quit" },
+	# close current tab and exit if it is last tab
 	{ on = [ "<C-q>" ], exec = "close" },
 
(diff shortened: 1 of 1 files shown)


sxyazi:
Ohhh that's intentional.

It seems that for time, size, reverse order is more common, and lowercase (without Shift) is easier. Let me know your usage habits.

The reason for using `<C-r>` instead of `U` for undo is that it can be used in insert mode, and to keep it consistent with Vim keybindings.

asim215:
> It seems that for time, size, reverse order is more common, and lowercase (without Shift) is easier. Let me know your usage habits.

I sometimes have to really pay attention to menu items for understand which key makes reverse ordering. For me reverse ordering more intuitive to uppercase letters. Because similar in vim: 1) t, f - forward, T, F - backward; 2) n - forward search results, N - backward; 3) o - down/forward, O - up/backward.

I see. We can replace to "U" in insert mode (helix) or left as it is now "Ctlr+r" (vim).

sxyazi:
> I sometimes have to really pay attention to menu items for understand which key makes reverse ordering. For me reverse ordering more intuitive to uppercase letters. Because similar in vim: 1) t, f - forward, T, F - backward; 2) n - forward search results, N - backward; 3) o - down/forward, O - up/backward.

Make sense

> I see. We can replace to "U" in insert mode (helix) or left as it is now "Ctlr+r" (vim).

Nope, in insert mode can only use `C-r`, but not `U`, which is typed as the character 'U'.

asim215:
Restored redo. Is it looks good for merge?

Issue #94, opened 2023-08-31
Support global config file in /etc


linsui:
IIUC, yazi doesn't support config files in /etc because `dirs` is used, as all other rust cli apps. Could you please support that on Linux?

sxyazi:
> as all other rust cli apps

AFAIK, neither ranger nor joshuto supports `/etc`, and I'm not sure there's any reason to need it.

If you're looking for a global configuration, that's already built into Yazi at build time, and doesn't need to be copied to the `/etc` manually.

linsui:
joshuto supports `$JOSHUTO_CONFIG_HOME` though which can be used to set the location of the config file. It's not ideal to me but can cover the uasge.

I'd like have global config in /etc because I want to manage it in my NixOS configuration and keep my home clean.

sxyazi:
Oh I see, maybe I can add a `YAZI_CONFIG_HOME` like joshuto dose, which is more flexible than `/etc` that's only valid for Linux.

BTW, Yazi has a [home-manager module](https://nix-community.github.io/home-manager/options.html#opt-programs.yazi.enable), but I haven't used NixOS in a while, and not sure if it works for `/etc` as well lol

linsui:
Thanks!

I don't use home-manager so I'd like to config yazi globally.

PR #97, opened 2023-08-31
feat: new `YAZI_CONFIG_HOME` env variable

sxyazi:
An implementation of the feature request: https://github.com/sxyazi/yazi/issues/94#issuecomment-1700582829

Close https://github.com/sxyazi/yazi/issues/94


linsui:
Thanks!

Issue #81, opened 2023-08-22
feature request for "find" aka jump to matching file/folder


teto:
thanks for creating this ! I used to use ranger but was tempted by joshuto for the rust bonus but either choice had issues and I feel yazi has strong foundamentals and can become the best choice for me:
- (image & text) preview by default !
- it was the fastest when opening my 200GB /nix/store that joshuto and ranger have difficulty with. I believe it could be even better and refresh the listing dynamically (I still had to wait ~8 sec) but it's much better than the other two.

The feature I miss the most currently is "search" or "filter". In ranger/joshuto, I could type `/` to start typing a folder name and go to it. Maybe it's possible in yazi but I couldn't find how ?

NB: nitpicks: PageUp and PageDown are not bound byu default and when the prompt to quit appears asking for "yes or no", I think it should quit on "y" press while right now I had to press "Enter" too.

Gajus84:
> 
> The feature I miss the most currently is "search" or "filter". In ranger/joshuto, I could type `/` to start typing a folder name and go to it. Maybe it's possible in yazi but I couldn't find how ?
> 

There are two possibilities, provided you have the requirements installed. FD and ripgrep (rg). Those are executed with "s" and "S". 

You can consult (https://github.com/sxyazi/yazi/tree/main/config/preset) for default keycodes.

I hope this is of help.

sxyazi:
Hi, thanks for the appreciation! Yazi integrates fzf and rg as the "search" feature; what you're looking for is called "find", which is documented in [Feature requests](https://github.com/sxyazi/yazi/issues/51), and I'll implement it sometime in the near future.

It's true that there's a lot of optimization potential in Yazi, such as dynamically loading file metadata, putting file sorting into a separate blocking threads pool, etc. It's a matter of finding the performance bottlenecks and optimizing for them, but I'll probably need to install a Nix before I can do that 🤣

Regarding the last point, does "yes or no" refer to the prompt triggered by quitting while there are still tasks left to complete?

teto:
> what you're looking for is called "find", which is documented in https://github.com/sxyazi/yazi/issues/51, and I'll implement it sometime in the near future.

cool ! let me know if you need testers 

> Regarding the last point, does "yes or no" refer to the prompt triggered by quitting while there are still tasks left to complete?

Yes that's exactly it. I am not sure what/how I created the tasks but the message mentioned those.

sxyazi:
You can press `w` key (the default) to see the tasks in progress

[...]

PR #104, opened 2023-09-03
feat: find

sxyazi:
An implementation of the feature request: https://github.com/sxyazi/yazi/issues/81

Close https://github.com/sxyazi/yazi/issues/81

---

https://github.com/sxyazi/yazi/assets/17523360/425aa023-cdb0-4d77-991d-c68472a5a7f4

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -177,9 +177,9 @@ dependencies = [
 
 [[package]]
 name = "base64"
-version = "0.21.3"
+version = "0.21.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "414dcefbc63d77c526a76b3afcf6fbb9b5e2791c19c3aa2297733208750c6e53"
+checksum = "9ba43ea6f343b788c8764558649e08df62f86c6ef251fdaeb1ffd010a9ae50a2"
 
 [[package]]
 name = "bincode"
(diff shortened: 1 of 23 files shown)


teto:
what software do you use to capture the video ? I like that but what would be even better is to update the selected item on each keypress way rather than waiting for the Enter keypress (similar to how ranger does)

sxyazi:
> what software do you use to capture the video

https://github.com/wulkano/Kap

> I like that but what would be even better is to update the selected item

Do you mean the "filter" feature of ranger? It seems a duplicate of the existing "search".

teto:
for people passing by, I answered here https://github.com/sxyazi/yazi/issues/81#issuecomment-1707372346

sxyazi:
Hey @teto, I believe I have implemented it in the way of Ranger, please give it a try.

The video has been updated to demonstrate it.

teto:
Thanks I've just tested and that's exactly what I wanted <3 and I think it'sobjectively better than the initial behavior .

NB: Unrelated question but how could I map C-w to erase the last word. I tried doing so while testing the feature but  was not sure how.

sxyazi:
> NB: Unrelated question but how could I map C-w to erase the last word. I tried doing so while testing the feature but was not sure how.

`C-w` and these keybindings are included in the "Emacs readline keybindings for the Input component" feature, which has been documented in the Feature requests. I will find some time to implement it.

c02y:
Sorry, I cannot activate the four `find` commands (`find/find --previous/find_arrow/find_arrow --previous`) using the builtin shortcuts like `/`.

`yazi 0.1.4-1`
`Arch Linux`

sxyazi:
@c02y This feature will be shipped with Yazi v0.1.5, you can build it from code to use it ahead

PR #106, opened 2023-09-04
fix: mime of javascript

XYenon:


sxyazi:
Thank you!

`application/javascript` is obsolete, but `file(1)` still uses it. It would be better to use a glob of `*/javascript` here to handle both `text/javascript` and `application/javascript`.

PR #109, opened 2023-09-05
fix: set cursor block after closing input prompt from insert mode

auvred:
I've noticed that after closing the input prompt from the insert mode, the cursor shape doesn't change back to block

How to reproduce:
1) open input prompt (e.g try to rename some file)
2) press enter to close it (from insert mode)
3) see that cursor shape is now bar (I guess it should be block as before opening `yazi`)

<details>
<summary>Before</summary>
<br>

https://github.com/sxyazi/yazi/assets/61150013/fb33b774-e1a9-4490-8ce8-be203a4c7b73

</details>

<details>
<summary>After</summary>
<br>

https://github.com/sxyazi/yazi/assets/61150013/22c7fa80-be8f-409d-93c0-b4164ec8ab51

</details>

(Not sure that I placed `Term::set_cursor_block()` to the right place)

### DIFF
--- a/core/src/input/input.rs
+++ b/core/src/input/input.rs
@@ -3,7 +3,7 @@ use std::ops::Range;
 use anyhow::{anyhow, Result};
 use config::keymap::Key;
 use crossterm::event::KeyCode;
-use shared::CharKind;
+use shared::{CharKind, Term};
 use tokio::sync::oneshot::Sender;
 use unicode_width::UnicodeWidthStr;
 
@@ -43,6 +43,7 @@ impl Input {
 				cb.send(if submit { Ok(self.snap_mut().value.clone()) } else { Err(anyhow!("canceled")) });
 		}
 
+		let _ = Term::set_cursor_block();
 		self.visible = false;
 		true
 	}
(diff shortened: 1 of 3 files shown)


sxyazi:
Thank you for your PR and detailed explanation!

It seems that for users whose default cursor is not a block, `set_cursor_block` doesn't make sense.

I think it's best to change `set_cursor_block` to `set_cursor_default` and use `SetCursorStyle::DefaultUserShape` instead of `SetCursorStyle::BlinkingBlock` here.

auvred:
You're right! It makes more sense, I updated it

auvred:
Okay I'll update this one too, but I'm not sure that we should show default shape in normal mode, if user's default shape is bar, then this might be confusing (same shape in both normal and insert modes)?

sxyazi:
Oh! I didn't think of that...

Keeping it as a `block` in normal mode should make sense, and it should restore to the `default` upon exiting.

sxyazi:
Thank you, merged!

adrian5:
I think this issue persists when confirming `There are … tasks running, sure to quit? (y/N)` to exit the program. I didn't look at the code, but maybe that code path avoids the fix.

PR #117, opened 2023-09-06
perf: load large folders in chunks

sxyazi:
Optimized for large folders, loading files in chunks. Now even folders with 100,000 files can open in one second.

---

Using a directory with 100,000 files as a test, before:

https://github.com/sxyazi/yazi/assets/17523360/55b19cd6-8893-4824-95e8-9842db13a864

After:

https://github.com/sxyazi/yazi/assets/17523360/831c310e-e399-4c71-abab-239625f5f37d


teto:
amazing !

Issue #118, opened 2023-09-06
Strange error "failed to connect to a Wayland server"


DreamMaoMao:
https://github.com/sxyazi/yazi/assets/30348075/e4d3aa68-1a85-485a-bd7a-b74882582b78

I use the platform is xorg(dwm), why yazi occur such fail message.

DreamMaoMao:
But there is nothing wrong with the operation, i  don't know why it reported this error, and I am not using wayland

DreamMaoMao:
this error will occur when i use delet commad like(dd cb ce)你 input field.

sxyazi:
What is your terminal, and are you using Überzug++? Please paste the output of `echo $TERM, $TERM_PROGRAM, $XDG_SESSION_TYPE` here.

DreamMaoMao:
![image](https://github.com/sxyazi/yazi/assets/30348075/4fb631c6-ad7f-4b2e-bd72-d96cea004e06)
my terminal is konsole

sxyazi:
Thanks for reporting, https://github.com/sxyazi/yazi/pull/119 should fix the issue, please give it a try.

PR #119, opened 2023-09-06
fix: doesn't redirect the stderr of the clipboard command to null

sxyazi:
Close https://github.com/sxyazi/yazi/issues/118


DreamMaoMao:
it work

Issue #105, opened 2023-09-03
Ctrl-z doesn't send yazi to foregroud


teto:
sry for yet another report but that's because I like the program :) 

Usually when I hit `Ctrl-Z`, it sends the program in background (it sends SIGSTOP), I go back to a shell prompt, I can run a few commands and then I can resume the program by typing `fg`. 
Hitting `Ctrl-Z`does nothing with 0.1.4 , and it is not mapped in my ~/.config/yazi/keymap.toml .

I would like `Ctrl-Z` to pause, which may imply for yazi to react to the STOP signal

sxyazi:
Seems that `SIGTSTP` is the signal that should be handled, and `SIGCONT` for resuming?

teto:
ha indeed, I looked for the signal quickly to give you some hint but looked like I was wrong :'(

teto:
Note: same remark for Ctrl-C that usually kills the command

sxyazi:
Okay, I will address it as soon as I finish refactoring.

teto:
thanks a lot for your hard work.

PR #120, opened 2023-09-07
feat: suspend process (`Ctrl-Z`)

sxyazi:
An implementation of the feature request: https://github.com/sxyazi/yazi/issues/105

Close https://github.com/sxyazi/yazi/issues/105

---

https://github.com/sxyazi/yazi/assets/17523360/d0b2c3d5-f2d6-4ae9-8f71-c074c8d85ea8


teto:
that's weird, I've recompiled from scratch 230362d6c78c6b3921254f32035d31b1fff804f8 and I cant suspend (nixos). Would be curious if someone else can try

sxyazi:
@teto Hey, this is a new command, if you're not using the default config, you'll need to add it to your config file:

https://github.com/sxyazi/yazi/blob/230362d6c78c6b3921254f32035d31b1fff804f8/config/preset/keymap.toml#L7

teto:
ha I see. I am not familiar with how it is implemented usually but this mapping works for 99% of the software out of the box, I feel like this should be maybe a (compiled) default.

teto:
Thank you very much for the explanation, I supposed that yazi was catching everything; I guess in most cases, zsh catches the Ctrl-Z and is the one sending the signal.

> this is based on Yazi's conservative expectation of user config, rather than bringing uncertain "surprises" to users

I dont really mind either way but in this case the uncertain surprise is to have "Ctrl-Z" not working IMO. The number of times I had a misfunctioning program that I stopped with Ctrl-C is huge. Let's say for some reason yazi goes into an infinite loop, having Ctrl-C registered automatically makes sense

sxyazi:
> I supposed that yazi was catching everything; I guess in most cases, zsh catches the Ctrl-Z and is the one sending the signal

Yes, when a TUI app is running in raw mode, it needs to handle all user keypresses by itself. However, when running in cooked mode, keys like `Ctrl-Z` and `Ctrl-C` are typically handled by the shell.

> Let's say for some reason yazi goes into an infinite loop, having Ctrl-C registered automatically makes sense

Given the reasons mentioned above, this doesn't seem to hold true. Since `Ctrl-C` needs to be handled by the TUI app, not the shell, there might not be an opportunity to process `Ctrl-C` when the app is consuming 100% CPU.

> in this case the uncertain surprise is to have "Ctrl-Z" not working IMO

Are you meaning hardcoding the registration of `Ctrl-Z` for all users in the code? If that's the case, it seems like an impact on potential user keybindings. Some users (like me) may not frequently use the "suspend" and could have `Ctrl-Z` bound to other actions, considering that this key is close to Ctrl :)

teto:
> Are you meaning hardcoding the registration of Ctrl-Z for all users in the code? If that's the case, it seems like an impact on potential user keybindings. Some users (like me) may not frequently use the "suspend" and could have Ctrl-Z bound to other actions, considering that this key is close to Ctrl :)

since yazi doesn't merge the default config with user's config, what I suggest is something we do in neovim too: when the program starts, it maps Ctrl-z to suspend before loading the system/user configuration: users are thus free to remap Ctrl-Z.

Issue #108, opened 2023-09-05
[Bug] open a folder with Chinese name and trigger jump fzf to search and select a file, nothing is visible.


DreamMaoMao:
This problem is very strange, and will only happen the first time, when you can not see file list in the current path, press the left arrow key, then press the right arrow key, the file in the current path can be saw again .

After that, I repeated the first operation (jump fzf to search and select a file in the same path). After the selection, I can see  file list in the path.

DreamMaoMao:
press left key and right key:

https://github.com/sxyazi/yazi/assets/30348075/f18881eb-14ea-4b01-a59b-50810827a513

DreamMaoMao:
secode time search:

https://github.com/sxyazi/yazi/assets/30348075/b58d008b-e66a-4414-8e42-b7d6349abbcc

sxyazi:
Thank you for your report. I will investigate it.

sxyazi:
I cannot reproduce this issue on the latest HEAD. Please try the latest code to see if the issue has gone.

https://github.com/sxyazi/yazi/assets/17523360/1a097ea5-b484-4e75-906a-8d9dc607e994

sxyazi:
Do they have any patterns, and how can I reproduce them?

DreamMaoMao:
I found the Rules that trigger bugs. I have a path that is
~/down

And then I have keymap config:
```
{ on = [ "g", "d" ],       exec = "cd ~/down",   desc = "Go to the downloads directory" },
```

The down path create method is:

```
ln -s /media/wrq/UUI/down  ~/down
```


This problem occurs when I use the yazi gd command to open ~/down and then jump fzf. Other directories don't have this problem

That is to say, only the Chinese folder in the soft link directory has this problem

sxyazi:
Sounds like it might be related to symbolic links. I will investigate it further.

DreamMaoMao:
There are other  bugs about link path. After opening link path, whether it is deleted, created, or renamed, it will automatically trigger a jump to the previous level of directory

sxyazi:
> There are other bugs about link path. After opening link path, whether it is deleted, created, or renamed, it will automatically trigger a jump to the previous level of directory

Can you record a GIF for this?

DreamMaoMao:
[...]

PR #125, opened 2023-09-08
fix: recognize symlink directories as files

sxyazi:
Try to fix https://github.com/sxyazi/yazi/issues/108


sxyazi:
It seems unrelated to the previous changes. Previously, we only modified the logic for the watcher, which involved path conversion during cd.

Could you please file a new issue, and let's conclude this one?

DreamMaoMao:
ok

sxyazi:
> now fzf selection will jump to the directory after the link instead of the current directory

I can't wait for a new issue, so I've already fixed it XD, try https://github.com/sxyazi/yazi/pull/126

PR #141, opened 2023-09-12
Swap description for search commands

knutwalker:
The help overview showed "Search files by content using ripgrep" for `seach fd` and "Search files by name using fd" for `search rg` when it should be the other way around.


sxyazi:
Oh thanks! I had them backwards.

PR #150, opened 2023-09-13
Add fallback mime guessing in case file(1) is not available

ndtoan96:
Relying on `file(1)` for mime guessing is quite ok for unix-like os. But on Windows, only few people add it to PATH. This pull request add a fallback mechanism in case `file(1)` is not in PATH.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -379,6 +379,7 @@ dependencies = [
  "futures",
  "indexmap 2.0.0",
  "libc",
+ "mime_guess",
  "natord",
  "notify",
  "parking_lot",
@@ -1049,6 +1050,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5a634b1c61a95585bd15607c6ab0c4e5b226e695ff2800ba0cdccddf208c406c"
 dependencies = [
  "autocfg",
+]
+
+[[package]]
+name = "mime"
+version = "0.3.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6877bb514081ee2a7ff5ef9de3281f14a4dd4bceac4c09388074a6b5df8a139a"
+
+[[package]]
+name = "mime_guess"
+version = "2.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4192263c238a5f0d0c6bfd21f336a313a4ce1c450542449ca191bb657b4642ef"
+dependencies = [
+ "mime",
+ "unicase",
 ]
 
 [[package]]
@@ -2003,6 +2020,15 @@ name = "typenum"
 version = "1.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "497961ef93d974e23eb6f433eb5fe1b7930b659f06d12dec6fc44a8f554c0bba"
+
+[[package]]
+name = "unicase"
+version = "2.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f7d2d4dafb69621809a81864c9c1b864479e1235c0dd4e199924b9742439ed89"
+dependencies = [
+ "version_check",
+]
 
 [[package]]
 name = "unicode-bidi"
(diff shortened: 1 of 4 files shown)


sxyazi:
Thank you for your pull request, but I'm afraid I may not be able to merge it for the following reasons:

- `mime_guess` only guesses MIME types based on file extensions, and it doesn't examine the file contents, which is quite limited. In contrast, `file(1)` checks both the file extension and the file contents.
- The [Windows installation guide](https://github.com/sxyazi/yazi/wiki/Windows-Installation-Guide) for Yazi already includes adding `file` to the PATH as a requirement.
- For users who prefer not to manually add `file` to their PATH, they can also use `scoop install file`.
- This would introduce additional complexity in handling user issues, as we would need to determine whether they have used the fallback method.

ndtoan96:
Here is my thought on the topic:
- Preview is a non-critical functionality, it's ok to guess the wrong mime type, it does not crash the app, does it?
- Guessing mime type from file extension already covers 90% use case out there. No sane person would create a jpeg file and put text in it, even if there's a case like that, would the app really need to support it?
- Adding `C:\Program Files\Git\usr\bin` to `PATH` just to get a preview is a hassle. Installing `scoop` so that we will not add it manually is even more of a hassle. At least that's my case, and I'm pretty sure many others think the same.

sxyazi:
I would like to clarify that the mimetype is crucial for Yazi, because it is not only used for preview and style rendering but, most importantly, Yazi relies on the correct mimetype to open files.

I want to ensure consistency in Yazi's behavior across platforms, and not using `file` on Windows would break this consistency. Please consider the following:

```toml
[open]
rules = [
  { "mime": "application/json", "use": "json" }
]
```

This is effective in `.prettierrc`, which is a common configuration file for [Prettier](https://prettier.io/) used in web development.

However, `mime_guess` obviously cannot correctly infer it as JSON, which would disrupt the user's JSON opening rules.

---

I believe most users have Git installed since Yazi is hosted on Git. All they need to do is add a variable, so I'm not sure why it would be a hassle.

ndtoan96:
I still don't think adding a fallback is a bad option. Almost all the cases where `mime_guess` cannot guess the correct mime type (including your example) are config text files with custom extension. In those cases, the return mime type is `text/plain`, which 99% will use the same opener as `json` or `xml` or whatever. I mean we almost always use a text editor to edit those files. And if the text editor is smart enough (vscode for example), it can then recognizes the correct type anyway. Even if it does not, user can always set the type once it opens.

sxyazi:
Would you be interested in adding it as a platform-specific feature? i.e., adding `[target.'cfg(target_os = "windows")'.dependencies]` build options for `mime_guess`.

Seems it makes sense for Windows users to conventionally differentiate file types based on file extensions, while Linux/macOS users are more accustomed to distinguishing file types based on file content and permissions.

I think this is the best compromise currently for both sides :)

ndtoan96:
Well, I just found out there is a mime type `application/x-msdownload` which includes `.bat` and `.dll` from `mime_guess` point of view. This is a problem since one is text file, another is binary file. So yeah, I will reconsider this issue.

There's one point I want to make clear, it's not that Windows users like to differentiate file types based on file extensions. The root issue is that users normally expect the app's basic functionalities to just work without having to do extra step or install extra software. Hence, guessing mime type from content is fine, but I would prefer something built-in.

I will try to experiment with [mime-sniffer](https://github.com/flier/rust-mime-sniffer) and [mime-classifier](https://github.com/demurgos/mime_classifier), both seem to be unmaintained though (maybe they just work and there's no issue to maintain to begin with since both are extracted from well tested applications)

sxyazi:
In fact, before this, I had already exchanged opinions with several Yazi contributors, and we unanimously agreed that using "file" for Windows users is the best approach. It not only provides consistency in behavior but also offers reliability in functionality. AFAIK, it is currently the most comprehensive among all detectors, with a database of over 8M, which obviously cannot be pre-built in Yazi.

I agree that it would be better if Yazi didn't rely on external software, but the reality is quite challenging. I have previously researched crates like `mime-sniffer`, and the conclusion is that it's best to avoid them. Apart from their lack of maintenance, there are some subtle differences in behavior compared to `file`. I believe this introduces a significant pitfall for the sake of convenience.

PR #151, opened 2023-09-13
Fall back to text if json provider is not available

ndtoan96:
Currently, if `jq` is not available, there's no preview for json file. We should fall back to text in that case.


sxyazi:
Great idea! can you add that fallback inside the `Provider::json` function?

Issue #145, opened 2023-09-12
multiple openers for a single rule


Linus789:
Would it be possible to add multiple openers for a rule? For example with
```toml
[opener]
text = [
	{ exec = 'nano "$1"', block = true, display_name = "Edit with nano" },
]
browser = [
	{ exec = 'chromium "$@"', display_name = "Open in Chromium" },
]

[open]
rules = [
	{ name = "*.html", use = ["browser", "text"] },
]
```
it would show
```
Open in Chromium
Edit with nano
```
as options.

This would reduce the overhead of duplicating entries.

Linus789:
Would you accept a PR for this issue?

sxyazi:
Of course! I really like this proposal and want to see it happen, just not get around to completing it...

PR #154, opened 2023-09-13
feat: multiple openers for a single rule

Linus789:
issue: #145

### DIFF
--- a/config/docs/yazi.md
+++ b/config/docs/yazi.md
@@ -81,6 +81,8 @@ rules = [
 
 	# { mime = "application/json", use = "text" },
 	{ name = "*.json", use = "text" },
+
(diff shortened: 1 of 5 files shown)


sxyazi:
Thank you!

Seems like it could break the existing user config, as they would need to switch to arrays. Is it possible to support both arrays and strings, similar to the keymap rules:

https://github.com/sxyazi/yazi/blob/9c6b3c9d1d1763074341042ef0f74ab45bb56da1/config/preset/keymap.toml#L31-L32

Here is the implementation code:

https://github.com/sxyazi/yazi/blob/9c6b3c9d1d1763074341042ef0f74ab45bb56da1/config/src/keymap/exec.rs#L53-L88

sxyazi:
Thank you, I made some simplifying. Merged!

PR #155, opened 2023-09-13
feat: sort ignore case

Linus789:


sxyazi:
Thank you for the PR!

Which do you think is more common, case sensitive and case insensitive? i.e. is it necessary to set case insensitivity as the default?

Linus789:
I can only speak for myself, but I prefer it to be case insensitive. And what default do you mean? The yazi.toml one, or the ones in the keymap?

sxyazi:
Yep, I agree with that as well. It seems that sorting by ASCII ordering doesn't make much sense, and case insensitivity makes more sense.

> The yazi.toml one, or the ones in the keymap?

Both.

core/src/files/sorter.rs
```diff
@@ -32,7 +32,7 @@ impl FilesSorter {
 
 		match self.by {
 			SortBy::Alphabetical => items.sort_unstable_by(|a, b| {
-				if self.ignore_case {
+				if self.sensitive {
```
Linus789, line 35: Shouldn't it be the inverse?

sxyazi: Yep, thank the reminder!

core/src/files/sorter.rs
```diff
@@ -65,12 +75,16 @@ impl FilesSorter {
 		indices.sort_unstable_by(|&a, &b| {
 			let promote = self.promote(entities[a].1, entities[b].1);
 			if promote != Ordering::Equal {
-				promote
-			} else if self.reverse {
-				natord::compare(&entities[b].0, &entities[a].0)
+				return promote;
+			}
+
+			let ordering = if self.sensitive {
```
Linus789, line 81: Here too

sxyazi:
Thank you, merged!

Issue #146, opened 2023-09-12
open: ignore case for name globbing


Linus789:
If I have for example this
```toml
[opener]
browser = [
	{ exec = 'chromium "$@"', display_name = "Open in Chromium" },
]

[open]
rules = [
	{ name = "*.html", use = ["browser"] },
]
```
and want to open a file called `test.HTML` it won't work. But it will work for `test.html`. Is there an easy way around that?

Edit: it seems to also apply to the theme icons.

sxyazi:
`*.[hH][tT][mM][lL]` looks silly but works... maybe I can add case insensitivity if needed.

sxyazi:
Hi, I just noticed that `*.@(html|HTML)` works for your case.

I wanted to know if you're still interested in adding a case-insensitive option. If so, I will add it now.

sxyazi:
Sorry, ignore what I said below. The Rust `glob` library doesn't support the `@()` syntax. I will add it now :)

Linus789:
> `*.@(html|HTML)`

Doesn't seem to work for me.

> I wanted to know if you're still interested in adding a case-insensitive option. If so, I will add it now.

Yes, I'm still interested. But I don't see a good way to add it for the theme icons.

PR #156, opened 2023-09-14
feat: make the glob expr case insensitive by default, and prepend `\s` to make it sensitive

sxyazi:
An implementation of the feature request: https://github.com/sxyazi/yazi/issues/146

Close https://github.com/sxyazi/yazi/issues/146


Linus789:
Thanks! :)

PR #160, opened 2023-09-16
feat: tab switch wraparound

ctem:
This allows tab switching to wrap back to the opposite extremity. E.g., if 5 tabs are open:
- `tab_switch 1` (<kbd>]</kbd> by default) from tab 5 will wrap to tab 1
- `tab_switch -1` (<kbd>[</kbd> by default) from tab 1 will wrap to tab 5

### DIFF
--- a/core/src/manager/tabs.rs
+++ b/core/src/manager/tabs.rs
@@ -33,7 +33,11 @@ impl Tabs {
 	}
 
 	pub fn switch(&mut self, idx: isize, rel: bool) -> bool {
-		let idx = if rel { self.absolute(idx) } else { idx as usize };
+		let idx = if rel {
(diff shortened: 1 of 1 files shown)


core/src/manager/tabs.rs
```diff
@@ -77,6 +77,17 @@ impl Tabs {
...
+	#[inline]
+	fn modulo(&self, rel: isize) -> usize {
+		let idx = (self.idx + rel as usize) % self.items.len();
+
+		if rel > 0 {
+			idx
+		} else {
+			(idx + self.items.len()) % self.items.len()
+		}
+	}
```
Linus789, line 89: Any reason to not use the `rem_euclid` function?

sxyazi:
LGTM, thank you!

PR #164, opened 2023-09-16
Plain text log

ndtoan96:
Avoid ansi color escape in log file.


sxyazi:
Hi, this is expected behavior as I have plans to implement a simple log viewer within Yazi, which requires these color codes

ndtoan96:
Oh, is that so. Yeah, that makes sense.

PR #174, opened 2023-09-17
fix: inconsistent `Shift` key behavior on Unix and Windows

### LINKED ISSUE #166: Some keymap ("shift" character) does not work on Windows Terminal
`~` or `?` in keymap config file does not work on Windows Terminal, instead we have to use `<S-~>` or `<S-?>`.

I have tracked down the issue: on Windows Terminal when I press `~`, crossterm returns key event with code `Char(~)` and SHIFT modifier, but with current implementation (https://github.com/sxyazi/yazi/blob/902c0942244f2e1753899ed0aa1d4faf5b01c7eb/config/src/keymap/key.rs#L58C7-L58C7), SHIFT is only enabled for uppercase characters.

Since I don't use linux or mac, I don't know how it behaves there. I'm not sure how to fix it.

ndtoan96:
Fix #166

### DIFF
--- a/config/src/keymap/key.rs
+++ b/config/src/keymap/key.rs
@@ -32,16 +32,17 @@ impl Default for Key {
 
 impl From<KeyEvent> for Key {
 	fn from(value: KeyEvent) -> Self {
-		/*
-		On Linx and Mac:
-			shift + alphabet => uppercase alphabet + SHIFT
-			shift + non alphabet => shifted non alphabet + NULL
-		On Windows:
-			shift + alphabet => uppercase alphabet + SHIFT
-			shift + non alphabet => shifted non alphabet + SHIFT
-		So we detect (non alphabet + SHIFT) and change it to (non alphabet + NULL) for consistent
-		behavior between OSs.
-		 */
+		// For alphabet:
+		//   Unix    :  <S-a> => Char("A") + Shift
+		//   Windows :  <S-a> => Char("A") + Shift
+		//
+		// For non-alphabet:
+		//   Unix    :  <S-`> => Char("~") + NULL
+		//   Windows :  <S-`> => Char("`") + Shift
+		//
+		// So we detect `Char("`") + Shift`, and change it to `Char("~") + NULL`
+		// for consistent behavior between OSs.
+
 		let shift = match (value.code, value.modifiers) {
 			(KeyCode::Char(c), _) if c.is_ascii_uppercase() => true,
 			(KeyCode::Char(_), m) if m.contains(KeyModifiers::SHIFT) => false,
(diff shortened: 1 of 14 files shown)


config/src/keymap/key.rs
```diff
@@ -32,11 +48,9 @@ impl Default for Key {
 
 impl From<KeyEvent> for Key {
 	fn from(value: KeyEvent) -> Self {
-		let shift = matches!(value.code, KeyCode::Char(c) if c.is_ascii_uppercase());
```
sxyazi, line 35: Why do we need to remove the Shift check here?

ndtoan96: Any char keycode now is compared without the shift modifier so it doesn't matter if uppercase characters have or don't have shift modifier. I remove it to save some execution time. Also, for it to be consistent with other characters like `~` and `?`.

ndtoan96:
What is your autoformat setting, I use the default one which comes with rust-analyzer, seems like it is different from yours.

sxyazi:
uh, I use rustfmt nightly and https://github.com/sxyazi/yazi/blob/main/rustfmt.toml as its config

sxyazi:
I tested it and it recognized `<S-a>` as `a`.

ndtoan96:
In my opinion, a keymap `<S-a>` does not make sense, we can just use `A`. If you want to support `<S-a>` style, then we will have to think about all other cases, for example `<C-S-/>` vs `<C-?>`. And if we take different keyboard layouts into the picture (French or Germania one for example), things will get quite complicated because we don't know which character is the "shifted" one.

sxyazi:
I think not supporting `<S-a>` should result in nothing happening; let it fail silently, just like Yazi's current behavior, rather than recognizing it as `a`.

ndtoan96:
Can you help me check again? I decided to work around the inconsistency for now because it kinda breaks the app on Windows. I'll come back to this issue later if the crossterm team decides to address it.

config/src/keymap/key.rs
```diff
@@ -32,13 +32,28 @@ impl Default for Key {
...
+		// For non-alphabet:
+		//   Unix    :  <S-`> => Char("~") + NULL
+		//   Windows :  <S-`> => Char("~") + SHIFT
+		//
+		// So we detect `Char("~") + SHIFT`, and change it to `Char("~") + NULL`
+		// for consistent behavior between OSs.
+
+		let shift = match (value.code, value.modifiers) {
+			(KeyCode::Char(c), _) if c.is_ascii_uppercase() => true,
+			(KeyCode::Char(_), m) if m.contains(KeyModifiers::SHIFT) => false,
```
sxyazi, line 48: Seems these two lines can be merged into `(KeyCode::Char(c), _) => c.is_ascii_uppercase()`?

ndtoan96: Yes, that's right

sxyazi: Done

PR #181, opened 2023-09-18
Fix image preview broken on zellij with wezterm

Eric-Song-Nop:
Zellij only supports sixel, and can detect zellij env with `ZELLIJ_SESSION_NAME`.

Also I ran `cargo fmt`

### DIFF
--- a/adaptor/src/adaptor.rs
+++ b/adaptor/src/adaptor.rs
@@ -29,6 +29,7 @@ pub enum Adaptor {
 impl Adaptor {
 	pub(super) fn detect() -> Self {
 		let vars = [
+			("ZELLIJ_SESSION_NAME", Self::Sixel),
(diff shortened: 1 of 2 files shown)


sxyazi:
Thank you!

peteringram0:
Is this available in 0.1.4 please? I still have issues with images inside Zellij if so.

Eric-Song-Nop:
No, it’s not included in 0.1.4

sxyazi:
This is a new feature that will be shipped with Yazi v0.1.5. You can try it out by building Yazi from Git in advance.

BTW, during the last test, there are some issues with the graphics support for Zellij. It occasionally glitches, and likely an issue with Zellij itself, since tmux works fine with the same protocol. See https://github.com/zellij-org/zellij/issues/2576

PR #187, opened 2023-09-19
[feat]: Byword editing

Eric-Song-Nop:
Provide basic by word editting like in default bash shell: <C-w> backspace a word, <C-,> or <C-.> for move back/forward a word


Eric-Song-Nop:
For now, can use `Ctrl-W` to delete back a word

PR #201, opened 2023-09-21
Fix LICENSE file

conradojordan:
Another small contribution.

Fixing the LICENSE file, that was named as `LICENCE` instead of `LICENSE`.
Also added a link to the LICENSE file in the `README.md` file. Feel free to make any changes to the PR


sxyazi:
Thanks! I will keep the old name "LICENCE" but add it to the README.

PR #205, opened 2023-09-22
feat: add flake.nix

### LINKED ISSUE #189: add flake.nix to be able to test the master version easily
hi,

nix is a package manager with a focus on build reproducibility. It has a nice feature called flake.nix that once committed to the git ttree allows to run `nix run github:sxyazi/yazi` would let anyone build (if not in nix cache) and run directly the master version. 
One can also specify a ref/rev with  `nix run github:sxyazi/yazi?rev=ae313bc` for instance.

As the yazi software evolves quite fast, I would love to be able to test the latest version without having to pull/cargo run (I also need to run `nix develop nixpkgs#yazi` to get the required dev tools). 

It's more convenient when the flake is at the root of the repo but it's possible to put it in a subfolder, as in neovim's case; `nix run 'github:neovim/neovim?dir=contrib'`

@sxyazi would you mind a PR adding this to the repo ? @XYenon any interest in this as you submitted yazi to nixpkgs ?

XYenon:
Close #189

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 target/
 
 .DS_Store
+result
+result-*
+.devenv*
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,256 @@
+{
+  "nodes": {
+    "devenv": {
+      "inputs": {
+        "flake-compat": "flake-compat",
+        "nix": "nix",
+        "nixpkgs": [
+          "nixpkgs"
+        ],
+        "pre-commit-hooks": "pre-commit-hooks"
+      },
+      "locked": {
+        "lastModified": 1695365659,
+        "narHash": "sha256-VsbuaBJ7Hx3aqdPPRT2Gp+TmAvtJtcffQGChws9VHZM=",
+        "owner": "cachix",
+        "repo": "devenv",
+        "rev": "1f21193496c515e55c556176e6aa13db9fdad1b4",
+        "type": "github"
+      },
+      "original": {
+        "owner": "cachix",
+        "repo": "devenv",
+        "type": "github"
+      }
+    },
+    "flake-compat": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1673956053,
+        "narHash": "sha256-4gtG9iQuiKITOjNQQeQIpoIB6b16fm+504Ch3sNKLd8=",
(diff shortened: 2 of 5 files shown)


sxyazi:
Thank you!

teto:
This project feels magical, issues get solved almost before being created <3
First good news this works ! 

A question, would it be better to show the git version here ?
```
nix run github:sxyazi/yazi -- --version
yazi 0.1.4
```
or maybe the commit in the version ?
```
ls -l result
lrwxrwxrwx 1 teto users 52 09-22 14:31 result -> /nix/store/ci5kbaxjndg6jj6ikvcmm1kra8sibjq1-yazi-git
```

Other question, last time I packaged a rust flake, changing the lockfile was a pain because it required updating a cargoDepsHash in the buildRustPackage call, this doesn't seem to be the case anymore ? (which is a good news).

thanks to you both

XYenon:
> A question, would it be better to show the git version here ?
> 
> ```
> nix run github:sxyazi/yazi -- --version
> yazi 0.1.4
> ```
> 
> or maybe the commit in the version ?
> 
> ```
> ls -l result
> lrwxrwxrwx 1 teto users 52 09-22 14:31 result -> /nix/store/ci5kbaxjndg6jj6ikvcmm1kra8sibjq1-yazi-git
> ```

I have created [a pr to add this](https://github.com/sxyazi/yazi/pull/206). And I want to know are there any other optimizations? 

> Other question, last time I packaged a rust flake, changing the lockfile was a pain because it required updating a cargoDepsHash in the buildRustPackage call, this doesn't seem to be the case anymore ? (which is a good news).

Yeah, I found `buildRustPackage` can read `Cargo.lock` directly. I hope other languages can also have this ability.

XYenon:
> I just tried to launch a nix shell and got
> 
> ```
>        error: devenv was not able to determine the current directory.
>        Make sure Nix runs with the `--impure` 
> ```
> 
> I have launched hundreds of different nix shells and it's the first time I am asked for the impure flag (admittedly it's the first time I see one using devenv). I think we lose the nixpkgs evaluation cache + reproducibility properties. Isn't there some way we can get rid of devenv ?

I have refactor it to avoid impure, could you give a review?
https://github.com/sxyazi/yazi/pull/293

PR #211, opened 2023-09-23
feat: highlight matching words on finding

PhotonQuantum:
https://github.com/sxyazi/yazi/assets/18749973/17ff3dd5-c4c3-4d30-8767-08dd6e34a1c6

## TODOs

1. Can we integrate the highlight style into the theme system in the future? 

2. We currently perform regex matching on all displayed filenames during every re-rendering, which may affect performance. Therefore, it would be beneficial to add a cache to optimize this process.

### DIFF
--- a/app/src/manager/folder.rs
+++ b/app/src/manager/folder.rs
@@ -50,7 +50,7 @@ impl<'a> Folder<'a> {
 	}
 
 	#[inline]
-	fn hovered_style(&self, file: &File) -> Style {
+	fn item_style(&self, file: &File) -> Style {
 		let mimetype = &self.cx.manager.mimetype;
 		THEME
 			.filetypes
@@ -115,7 +115,7 @@ impl<'a> Widget for Folder<'a> {
 				} else if hovered {
 					THEME.selection.hovered.get()
 				} else {
-					self.hovered_style(f)
+					self.item_style(f)
 				};
 
 				let mut spans = Vec::with_capacity(10);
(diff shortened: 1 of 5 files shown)


sxyazi:
> Can we integrate the highlight style into the theme system in the future?

Sure.

> We currently perform regex matching on all displayed filenames during every re-rendering, which may affect performance. Therefore, it would be beneficial to add a cache to optimize this process.

After some reflection, I've come to realize that Yazi's precaching mechanism was designed for exactly this purpose. It allows us to offload certain time-consuming tasks (IO / CPU) to the background and span across multiple threads. Perhaps in the future, we can utilize precaching to optimize it further, but for now, let's keep it simple!

sxyazi:
Thank you! I have made some simplifications:

- Using `Cow` to avoid memory allocating
- Removed unnecessary `FolderLocation` as we don't need to obtain the current location (is-a), but only need to check if it currently contains `with_find` (has-a). This allows for more flexible combinations of various functionalities.
- Merged `short_path` and `short_path_parts` into a single function and returned a deterministic result.

PhotonQuantum:
> * Removed unnecessary `FolderLocation` as we don't need to obtain the current location (is-a), but only need to check if it currently contains `with_find` (has-a). This allows for more flexible combinations of various functionalities.

I decided not to use `is_find` so that I could mimic the `incsearch` behavior. It's helpful to see how the pattern matches as you type.

sxyazi:
Not quite sure what `mimic the incsearch behavior` means here.

In this case,`is_find` has the same effect as `self.location == FolderLocation::Current`, and currently, it's `true` only when the location is *current*.

PhotonQuantum:
Sorry for the confusion earlier. I thought 'is_find' would only be true if the search request was committed, but now I understand that it is set to true as soon as a search is initiated. Everything is working correctly now!

sxyazi:
Thanks, merged! 🎉

PR #212, opened 2023-09-23
feat: include ignored files on search when hidden files are shown

PhotonQuantum:
```
  -I, --no-ignore
          Show search results from files and directories that would otherwise be ignored by
          '.gitignore', '.ignore', '.fdignore', or the global ignore file. The flag can be
          overridden with --ignore.
```

### DIFF
--- a/core/src/external/fd.rs
+++ b/core/src/external/fd.rs
@@ -17,7 +17,7 @@ pub fn fd(opt: FdOpt) -> Result<UnboundedReceiver<File>> {
 	let mut child = Command::new("fd")
 		.arg("--base-directory")
 		.arg(&opt.cwd)
-		.arg(if opt.hidden { "--hidden" } else { "--no-hidden" })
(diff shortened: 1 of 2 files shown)


sxyazi:
Currently, `opt.hidden` only represents Yazi's "hidden mode", which simply checks if a file starts with a `.`. It does not take into account `.gitignore`.

Therefore, to ensure consistency in this behavior, should it not be `--no-ignore` instead of `--ignore` when `!opt.hidden` is true?

sxyazi:
Thank you!

Merge first, and later, we will create `fd` and `rg` as independent plugins to provide users with finer control over the parameters.

PhotonQuantum:
> Therefore, to ensure consistency in this behavior, should it not be `--no-ignore` instead of `--ignore` when `!opt.hidden` is true?

It's really confusing but fd `--hidden` means including hidden files while `--ignore` means *excluding* ignored files in fd and rg.

PR #213, opened 2023-09-23
feat: scroll half/full page with `arrow` percentage supported, and new Vi-like `<C-u>`, `<C-d>`, `<C-b>`, and `<C-f>` keybindings added

TD-Sky:
Now users can use command like `arrow 50%`, `arrow -50%` to move cursor in the distance calculated by percentage of page rows.

In a nutshell,
* `arrow -50%` to scroll up half page.
* `arrow 50%` to scroll down half page.
* `arrow 100%` to scroll down one page.
* `arrow 100%` to scroll up one page.

Users can customize the number, but I think the above is enough. See `config/preset/keymap.toml` in this PR.

### DIFF
--- a/app/src/executor.rs
+++ b/app/src/executor.rs
@@ -61,7 +61,7 @@ impl Executor {
 
 			// Navigation
 			"arrow" => {
-				let step = exec.args.get(0).and_then(|s| s.parse().ok()).unwrap_or(0);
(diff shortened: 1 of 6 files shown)


TD-Sky:
Wait a minute to merge. I want to change some codes.

sxyazi:
Okay.

PR #214, opened 2023-09-23
fix: handle shell arguments on Windows

### LINKED ISSUE #192: Path(s) are added automatically to shell command on Windows
https://github.com/sxyazi/yazi/assets/33489972/63042c3e-c655-44b2-9982-93882f18f11b

This issue has been reported via Discord but I'll also leave it here to keep track.

ndtoan96:
Fix #192

### DIFF
--- a/core/src/external/shell.rs
+++ b/core/src/external/shell.rs
@@ -63,13 +63,7 @@ pub fn shell(opt: ShellOpt) -> Result<Child> {
 #[cfg(target_os = "windows")]
 mod cmdexpand {
 	use anyhow::{anyhow, Result};
-	use nom::branch::alt;
-	use nom::bytes::complete::{is_not, tag, take_while1};
-	use nom::character::complete::{anychar, char, digit1, space0, space1};
-	use nom::combinator::recognize;
-	use nom::multi::{many0, many1};
-	use nom::sequence::{delimited, pair, preceded, tuple};
-	use nom::IResult;
+	use nom::{branch::alt, bytes::complete::{is_not, tag, take_while1}, character::complete::{anychar, char, digit1, space0, space1}, combinator::recognize, multi::{many0, many1}, sequence::{delimited, pair, preceded, tuple}, IResult};
 
 	enum CommandPart<'a> {
 		Space(&'a str),
--- a/core/src/external/shell.rs
+++ b/core/src/external/shell.rs
@@ -4,9 +4,9 @@ use anyhow::Result;
 use tokio::process::{Child, Command};
 
 pub struct ShellOpt {
-	pub cmd:    OsString,
-	pub args:   Vec<OsString>,
-	pub piped:  bool,
+	pub cmd: OsString,
+	pub args: Vec<OsString>,
+	pub piped: bool,
 	pub orphan: bool,
 }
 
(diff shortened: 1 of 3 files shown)


sxyazi:
I've made a simple test with:

```rust
cmdexpand::expand_cmd(r#"  a "b \"%1\""  "#, &["c", "d"])
```

And got:

```
  a "b \"
```

Is this expected?

ndtoan96:
I'm aware that it has problem with escaped quote, but I spent yesterday fixing it in my crate and making documentation so haven't ported it here. Will find time to do it later.

ndtoan96:
Just realized this expand thing only works with no-space argument. I tried to open a text file `"D:\a b.txt"` by select the file and run `%EDITOR% %1`. Even though it expands correctly to `%EDITOR% "D:\a b.txt"`, the editor actually open 2 files `"D:\a` and `b.txt"`.

Pretty sure it's because the way `cmd /C` works, argument should be feed one by one instead of all in one.

sxyazi:
I have to sleep, will investigate it tomorrow. =)

sxyazi:
I just tested it, and it indeed doesn't work as expected:

```bash
cmd /C "nvim 'rust fmt.toml'"
```

This results in opening two files (`'rust` and `fmt.toml'`), and `cmd` doesn't correctly recognize the quotes. I'm thinking of two simple workarounds now:

- Use PowerShell instead of cmd.
  I found that PowerShell can work correctly:

  ```bash
  powershell -Command "& nvim 'rust fmt.toml'"
  ```

  However, this requires PowerShell to be present on the user's system, it requires a minimum of Windows 8. In fact, I haven't tested whether Yazi works correctly on Windows 7 and earlier systems yet...

- Parse only the first-level placeholders, and, if they exist, pass them as cmd parameters:

  From:

  ```bash
  cmd /C "nvim 'rust fmt.toml'"
  ```

  To:

  ```rust
  Command::new("cmd").args(["/C", "nvim", "rust fmt.toml"])
  ```

 It was similar to Yazi's current behavior, but the difference is that this time we can use `nom` to parse more accurately which first-level placeholders are available.

Apart from these, I'm not sure if there are better solutions. Any ideas?

ndtoan96:
I prefer the second approach. Reason is that user can use environment variable in the form `%SOME_VAR%`, which is nicer, and more well known than `$env:SOME_VAR` in powershell.

sxyazi:
Oh, that makes sense, It's the first time I've seen this kind of peculiar format 🤣

ndtoan96:
I thought it would be easy enough to make a manual parser. The code doesn't look pretty but it works (at least if users don't do anything too crazy).😅

core/src/external/shell.rs
```diff
@@ -49,3 +50,166 @@ pub fn shell(opt: ShellOpt) -> Result<Child> {
...
+		#[test]
+		fn test_double_quote() {
+			let args = parse_cmd_to_args("echo \"abc ' 'xyz\" \"%1\" %2 %3", &["111", "222"]);
+			assert_eq!(args, vec!["echo", "abc ' 'xyz", "111", "222", ""]);
+		}
+
+		#[test]
+		fn test_escaped() {
+			let args = parse_cmd_to_args("echo \"a\tbc ' 'x\nyz\" \"\\%1\" %2 %3", &["111", "22  2"]);
+			assert_eq!(args, vec!["echo", "a\tbc ' 'x\nyz", "%1", "22  2", ""]);
```
sxyazi, line 194: This test case seems wrong; the fourth element should be `"\\%1"` instead of `"%1"`?

sxyazi: Got it!

sxyazi:
I've made some simplifications, merging now. Thanks!

PR #218, opened 2023-09-24
feat: add overlays to flake.nix

musjj:
Having overlays makes it easier to install on NixOS systems.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -22,11 +22,11 @@
 
         devShells.default = import ./nix/shell.nix { inherit pkgs inputs; };
       }) 
-    // rec {
-      overlays = {
-        default = overlays.yazi;
+    // {
+      overlays = rec {
+        default = yazi;
         yazi = final: prev: {
-          yazi = final.callPackage ./nix/yazi.nix { };
+          yazi = self.packages."${final.system}".yazi;
         };
       };
     };
--- a/flake.nix
+++ b/flake.nix
@@ -29,6 +29,5 @@
           yazi = final.callPackage ./nix/yazi.nix { };
         };
       };
-      overlay = overlays.default;
     };
 }
--- a/flake.nix
+++ b/flake.nix
@@ -21,5 +21,14 @@
         formatter = pkgs.nixpkgs-fmt;
 
         devShells.default = import ./nix/shell.nix { inherit pkgs inputs; };
-      });
+      }) 
+    // rec {
+      overlays = {
+        default = overlays.yazi;
+        yazi = final: prev: {
+          yazi = final.callPackage ./nix/yazi.nix { };
+        };
+      };
+      overlay = overlays.default;
+    };
 }


sxyazi:
cc @XYenon @teto

Could you please review this PR, I've completely forgotten my limited Nix knowledge lol 🤣

flake.nix
```diff
@@ -21,5 +21,14 @@
...
-      });
+      }) 
+    // rec {
+      overlays = {
+        default = overlays.yazi;
+        yazi = final: prev: {
+          yazi = final.callPackage ./nix/yazi.nix { };
+        };
+      };
+      overlay = overlays.default;
```
teto, line 32: if you dont need it, let's remove the "overlay" since it's deprecated.

teto:
you can install `yazi.packages.${system}.yazi`, how does the overlay make it easier, aka you can just import the overlay without modifying the rest of your config ?

flake.nix
musjj: Didn't know there was a deprecation. I though it was just a unofficial semi-common community practice. But it looks like you're right: https://github.com/NixOS/nix/blob/master/doc/manual/src/release-notes/rl-2.7.md

flake.nix
```diff
@@ -21,5 +21,14 @@
         formatter = pkgs.nixpkgs-fmt;
 
         devShells.default = import ./nix/shell.nix { inherit pkgs inputs; };
-      });
+      }) 
+    // rec {
+      overlays = {
+        default = overlays.yazi;
+        yazi = final: prev: {
+          yazi = final.callPackage ./nix/yazi.nix { };
```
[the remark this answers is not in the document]
musjj: You can't re-use the `yazi` inside `eachDefaultSystem` because it's already instantiated into a specific architecture. An overlay should not be dependent on a specific architecture and accept whatever `nixpkgs` that was passed to it.

musjj:
In flakes, you actually can't access the current system architecture AFAIK. So you have to use workarounds by passing custom args from `flake.nix` into your configuration, or just hardcoding it. Most other repo flakes I've seen provide overlays which seems like a cleaner and more convenient solution.

sxyazi:
I'm cleaning up PRs, looks like this PR is just one step away.

@teto, would you like to give it a review? If it's ready, feel free to ping me anytime to get it merged! 😉

flake.nix
```diff
@@ -21,5 +21,13 @@
         formatter = pkgs.nixpkgs-fmt;
 
         devShells.default = import ./nix/shell.nix { inherit pkgs inputs; };
-      });
+      }) 
+    // rec {
```
XYenon, line 25:
```suggestion
    // {
      overlays = rec {
        default = yazi;
        yazi = final: prev: {
          yazi = self.packages."${final.system}".yazi;
        };
      };
    };
```

flake.nix
```diff
@@ -21,5 +21,13 @@
...
+      overlays = {
+        default = overlays.yazi;
+        yazi = final: prev: {
+          yazi = final.callPackage ./nix/yazi.nix { };
+        };
+      };
+    };
```
XYenon, line 32:
```suggestion
    // {
      overlays = rec {
        default = yazi;
        yazi = final: prev: {
          yazi = self.packages."${final.system}".yazi;
        };
      };
    };
```

musjj: Thanks, didn't know that you can obtain the system information this way! I thought the only way to obtain it is from `builtins.currentSystem` which is impure and not allowed in flakes.

XYenon (approved):

flake.nix
```diff
@@ -21,5 +21,13 @@
         formatter = pkgs.nixpkgs-fmt;
 
         devShells.default = import ./nix/shell.nix { inherit pkgs inputs; };
-      });
+      }) 
+    // {
+      overlays = rec {
```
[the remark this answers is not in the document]
musjj: It's just a convention I usually see in flakes. The `packages` already in the flake is an example of this pattern (it provides both `yazi` and `default`).

sxyazi:
Thank you all, merged! 🎉

PR #161, opened 2023-09-16
feat: new theme system

sxyazi:
## TODOs:

Features:

- [x] Progress
- [x] Highlighting
- [x] Short path

Issues:

- [ ] https://github.com/sxyazi/yazi/issues/220
- [ ] https://github.com/sxyazi/yazi/pull/221, https://github.com/sxyazi/yazi/issues/149
- [x] https://github.com/sxyazi/yazi/issues/110#issuecomment-1728665761
- [x] Closes #204
- [x] Closes #246

The PR is bigger than anticipated and has taken more time.

There are still some unfinished works, additional PRs will be implemented shortly to prevent this one from continuing to expand.

Issue #149, opened 2023-09-13
default theme is not friendly for light terminal background


tkkcc:
yazi 0.1.4 on archlinux, no custom config
alacritty terminal with light background
open yazi and type `g`, the keymap prompt text is unrecognizable:
![image](https://github.com/sxyazi/yazi/assets/17373509/a8a9aa4d-f756-4646-bc63-8ba3be62551f)

sxyazi:
Have you made a light theme for Yazi?

tkkcc:
no yazi config, but I have custom $LS_COLORS, is it related?

```txt
[...]

PR #221, opened 2023-09-26
fix: collect and fix all hard coded themes and color

Eric-Song-Nop:
- [x] Which
- [x] Tasks
- [x] A color in manager/folder (not sure what is that)
- [x] help
- [x] find arrow
- [x] header
- [x] input
- [x] select  

Plz let me know if there is more.
#149 #161


sxyazi:
Excellent work!

Eric-Song-Nop:
still need to fix ansi color problem

Eric-Song-Nop:
Fix ansi

sxyazi:
Sorry for taking so long to address this PR.

I will merge it once https://github.com/sxyazi/yazi/pull/161 is completed, since pr-161 breaks some things, such as removing the `.get()` method from `THEME.xxx.get()`. It shouldn't be too far off. 🚀

PR #228, opened 2023-09-30
Fix Windows build and add github action

### LINKED ISSUE #227: Build breaks on Windows
Commit f9eeac6
```
error[E0515]: cannot return value referencing temporary value
   --> core\src\manager\finder.rs:125:3
    |
116 |           let b = { name.to_string_lossy().as_bytes() };
    |                     ---------------------- temporary value created here
...
125 | /         Some((
126 | |             String::from_utf8_lossy(&b[..range.start]),
127 | |             String::from_utf8_lossy(&b[range.start..range.end]),
128 | |             String::from_utf8_lossy(&b[range.end..]),
129 | |         ))
    | |__________^ returns a value referencing data owned by the current function
```

ndtoan96:
- Fix #227 
- Add github action


sxyazi:
This is so fast! I'm still struggling with Rust ownership, and you've already fixed it, unbelievable! 🙀

BTW, I tried so many ways and couldn't make it work. I'm curious if there's any way to handle it without moving `as_bytes` to the outer scope?

ndtoan96:
I whink we can call `to_string_lossy` in upper scope, then change the parameter type to `Cow<'a, str>`

sxyazi:
Merge first, this piece of code will soon be replaced by Lua. 🤣

PR #230, opened 2023-09-30
feat: backward/forward

ndtoan96:
~~Backward/forward only works with big jump (cd, zoxide...).~~

~~This pull request also includes an improvement of #228.~~

https://github.com/sxyazi/yazi/assets/33489972/4942549d-6f56-4347-a45a-900aedade018


sxyazi:
Why does backward/forward only work for cd? I think it should be expanded to apply to all navigation, and this wouldn't affect the current behavior of cd, just widen the scope.

ndtoan96:
> Why does backward/forward only work for cd? I think it should be expanded to apply to all navigation, and this wouldn't affect the current behavior of cd, just widen the scope.

For practical reason. Just like in text editor, you only need backward/forward when you make a jump (imagine how it would be if you go backward and the cursor move up one line).

sxyazi:
I haven't seen any file managers that implement forward/backward only for the "jump"... I can't imagine what harm reducing the granularity of operations would cause; on the contrary, if the granularity is too large, it would be limiting when users actually need it.

ndtoan96:
I also use other file manager and I think this is the better way to do backward/forward.

Imagine if I'm in folder `a`, I want to check something in `b` so I jump to it. Now I realize the thing I'm looking for is not in `b` so I navigate down to `b/1/2/3/4/5`. Now, after checking the thing I need, I want to go back to `a` to continue the work. With normal backward, I have to press `H` 5 times up to folder `b` and then one more time to go back to `a`. When I am already back to `a`, I realize I need to check one more thing then I have to forward 6 times to go back to `b/1/2/3/4/5`.

In the same situation, with this implementation, when I'm in `b/1/2/3/4/5` I need only press `H` once to go to `a`, then `L` once to go back to `b/1/2/3/4/5`.

Someone would argue "what if I want to go back one folder?", just use `h` in that case.

sxyazi:
Have you ever considered: In the current directory `abc`, there are two subdirectories, `foo` and `bar` separated by `n` files. Here is the navigation path:

```
abc -> abc/foo -> abc -> abc/bar
                                  -> abc/foo
```

- The current cursor is on `abc/foo`, and I press `l` to enter it
- After some operations, I press `h` to return to the parent directory `abc`
- I press `j` continuously `n` times to move the cursor to `abc/bar`, and press `l` to enter it
- At this point, I need to return from `abc/bar` to `abc/foo`
  - Current implementation: I have to return to the parent directory again and press `k` for `n` times, until the cursor is on `abc/foo`, and then press `l` to enter it.
  - My implementation: I only need to press 2 times `H`; and I can also press `L` twice to go back again.

ndtoan96:
I don't run into that situation often but I can see where you're coming from. How about this: keep the current implementation and add an `add_to_jump_stack` (temporary name) command, binding it to `m` or some other key.

So when you're at `abc/foo`, press `m` to add it to the stack. After that when you're in `abc/bar`, press `H` to return to `abc/foo`.

ndtoan96:
On second thought, that kind of functionality may better be done in a bookmark or jump list plugin.

Ok, I'll change the behavior to that of normal file manager.

app/src/app.rs
```diff
@@ -119,9 +119,9 @@ impl App {
 		let manager = &mut self.cx.manager;
 		let tasks = &mut self.cx.tasks;
 		match event {
-			Event::Cd(url) => {
+			Event::Cd(url, backstack_push) => {
```
sxyazi, line 122: `backstack_push` seems to be useless anymore?

ndtoan96: No, it is still necessary. Because when we `cd` via back and forward, we don't want the path to be pushed to stack.

sxyazi:
I made some changes, removed the unnecessary `backstack_push` to simplify the code, could you give it a review?

ndtoan96:
Oh, that's a clever way to guard against backward/forward pushing to stack. It looks fine to me.

PR #233, opened 2023-09-30
fix: a logic error in `Step`

### LINKED ISSUE #231: Bug: cd exit with code 101
While implementing the backward/forward, I've noticed one bug.

Steps to reproduce:
1. Run yazi at `D:\projects\debug\yazi`
2. `j` to move cursor to `nix` folder
3. `g <space>` and enter `D:\projects\debug\yazi\target`
4. App crash, exit code 101

Tested on 5584ba48442c780d5a2b2c49419b19ee746c0b89

Update: I am able to reproduce the issue on Linux too.

Update: version v0.1.4 does not have this issue.

https://github.com/sxyazi/yazi/assets/33489972/5346c261-5930-47dd-8809-d54d6b5ac704

ndtoan96:
fix #231

### DIFF
--- a/core/src/manager/folder.rs
+++ b/core/src/manager/folder.rs
@@ -84,7 +84,7 @@ impl Folder {
 		self.set_page(false);
 
 		if self.cursor < self.offset + 5 {
-			self.offset = self.offset.saturating_sub(old - self.cursor);
+			self.offset = self.offset.saturating_sub(old.saturating_sub(self.cursor));
 		}
 
 		old != self.cursor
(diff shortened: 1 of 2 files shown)


sxyazi:
Thank you for the PR.

But I can't reproduce the issue, and not quite clear on why "prev" would result in `old < self.cursor`, any tips here?


https://github.com/sxyazi/yazi/assets/17523360/ac637c2b-06c7-4ebe-92e3-e1902370772b

ndtoan96:
Supposed you have two subfolders in the current folder
```
-->a
   b
```
You put the cursor at the second one (but don't enter it)
```
   a
-->b
```
Then cd to the first one

sxyazi:
I reproduced it.

Thanks for the report, it was a logic error of `Step` not a overflow error.

PR #240, opened 2023-10-02
feat: added new options to the `find' command for smart-case/case-insensitive finding

### LINKED ISSUE #239: Case insensitive find by default
I'm thinking of implement the `fd` strategy for `find`, meaning if the query has uppercase character then we use case-sensitive regex, else case-insensitive regex should be enabled by default.

ndtoan96:
Close #239

### DIFF
--- a/app/src/executor.rs
+++ b/app/src/executor.rs
@@ -154,12 +154,11 @@ impl Executor {
 			"find" => {
 				let query = exec.args.get(0).map(|s| s.as_str());
 				let prev = exec.named.contains_key("previous");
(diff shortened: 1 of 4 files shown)


sxyazi:
Well done! I'd like to hide it behind `find --smart` and change the default `/` and `?` to it.

This will be advantageous for us to later add something like `find --insensitive` without breaking the default (sensitive) behavior.

ndtoan96:
I added `--smart-case` and `--ignore-case` flags and changed keymap to use `--smart-case`.

sxyazi:
Thank you!

PR #241, opened 2023-10-02
feat: support environment variable in `cd` path

### LINKED ISSUE #196: Note on 'cd' navigation on windows
The variable `~` is not expanded by `cd`, and environment variables like `USERPROFILE` (equivalent of `$HOME` on *nix systems) are not expanded either.

Therefore, to my knowledge, one needs to hardcode paths for `cd` as follows:
```toml
{ on = [ "g", "h" ],       exec = "cd c:/users/username",             desc = "Go to the home directory" },
```
or (note four backslashes):
```toml
{ on = [ "g", "h" ],       exec = "cd c:\\\\users\\\\username",             desc = "Go to the home directory" },
```

ndtoan96:
Fix #196 

This pull request implements the below features:
- Expand `~` on Windows
- Support using environment variables in cd


sxyazi:
Why do we need to use `echo` here? In Unix implementations, when there are special symbols in the user's path (such as `;`, which is a valid filename), it can potentially lead to ACE (Arbitrary Code Execution).

For example, `foo;rm bar.txt` would result in `echo foo;rm bar.txt`.

ndtoan96:
Well, `echo` is the simplest way to expand path. It's better than writing another parser.

I think ACE can be avoided by wrapping the path with quote.

sxyazi:
Hmm, seems unnecessary to make it so complicated. Just expanding the system preset variables should suffice?

This is just a simple replace, like `replace("$USERPROFILE", || env::get("USERPROFILE"))`, without worrying about user variables.

ndtoan96:
On Windows, there're more presets than you might think. In my knowledge alone, there are %USERPROFILE%, %ALLUSERSPROFILE%, %TEMP%, %TMP%, %APPDATA%, %LOCALAPPDATA%, %PROGRAMDATA%, %PROGRAMFILES%, %COMMONPROGRAMFILES%,...

And when you install some programs, they add more to the list (%GO_HOME%, %PYTHONPATH%, %OneDrive%,...)

sxyazi:
I found 97 variables at https://learn.microsoft.com/en-us/windows/deployment/usmt/usmt-recognized-environment-variables, and I think users might not be aware of most of them. Implementing all of them would be laborious and pointless.

Is it possible to implement only a portion of them or have Yazi set a unified name that can work across different systems, similar to [what dirs does](https://github.com/dirs-dev/dirs-rs#features)? Alternatively, can we directly use dirs, which is currently a dependency of Yazi?

ndtoan96:
One user may not use all the variables, but a combination of many users may cover a whole lot of them.

On the contrary, I don't think calling `echo` is something complicated. It's no different than calling `file` for mimetype or `lsar` for archive review. Even better since we don't have to install anything.

sxyazi:
- If it relies on the `cmd`, `sh`, and `echo` commands, it should be placed in the `external/` directory as a separate file. Since it relies on `cmd` and `sh`, it should be handled through `shell.rs` rather than creating a new one, and it should be asynchronous. Yazi encapsulates all external commands that need to be called multiple times as asynchronous because it's not just a system call but involves the entire process lifecycle of creation and destruction. This would add complexity to `expand_path`.
- I'm not a security expert, I can't be certain if the current Unix implementation is absolutely secure. Currently, the visible issue is that only `"` is escaped, and `\` itself is not escaped, which is also a valid part of a filename.
- In the Windows implementation, why is `.replace("\\\"", "\"")` needed and is this an undefined behavior? Also, I would prefer using `OsString` to get the result as it can be directly converted to `PathBuf` without the need for `String::from_utf8_lossy`.
- I'm not sure if implementing complete variable expansion is worthwhile. I lean towards implementing only the common ones (in the `dirs`'s list), and leaving more complex cases to implement through plugins (Lua can read env variables).

ndtoan96:
I think this should be built-in instead of relying on plugin, because I don't expect my file manager to be less useful than a `cd` command.

If you don't want to use external program call, then how about just using regex to expand it?

sxyazi:
How about the proposal to make `echo` asynchronous mentioned above?

BTW, I'm not sure how about Windows, but in Unix the `cd` command "indeed" doesn't support expanding paths; it relies on the shell's expansion capabilities, and Yazi's goal is not to reimplement a shell, that might seem reasonable although...

ndtoan96:
Currently `shell.rs` is in `core` module. If we want to call `echo` via `shell.rs` then we have to move it to `shared` module. What do you think?

sxyazi:
This is a real issue, I'll try to make it done later

shared/src/fns.rs
```diff
@@ -5,12 +5,45 @@ use tokio::fs;
...
+
+	// expand the environment variable by calling the "echo" command, in linux case,
+	// this also expands the '~' path
+	#[cfg(target_os = "windows")]
+	let expanded_path = match std::process::Command::new("cmd").args(&["/C", "echo"]).arg(p).output() {
+		Ok(output) if output.status.success() => Some(
+			String::from_utf8_lossy(&output.stdout)
+				.trim_end()
+				.trim_matches('"')
+				.replace("\\\"", "\"")
```
sxyazi, line 18: When do quote marks appear? I tested several cases but none of them does this.

ndtoan96: When there is `"` in the file name. You'll see it if you try to run:
```rust
std::process::Command::new("cmd")
        .arg("/C")
        .arg("echo")
        .arg(std::path::Path::new("a\"b.txt"))
        .output();
```

sxyazi:
```
a\"b.txt
```

Seems the quote marks only appear in the middle, not on both sides. Is `trim_matches('"')` here meaningless?

ndtoan96: That is for another case where there is space in the file name.

sxyazi: Oh I see, thanks

sxyazi:
I switched to the regex, would you like to give it a review?

ndtoan96 (commented):
Review finished.

sxyazi:
Thank you!

PR #245, opened 2023-10-05
feat: add new options to `quit` command for flexible cwd-file setting

### LINKED ISSUE #243: [Feature Request] Option to quit without writing cwd-file
## Background
Currently, users cannot choose whether quitting yazi with changing the current working directory. According to the [documentation](https://yazi-rs.github.io/docs/usage/quick-start/#changing-working-directory-when-exiting-yazi), users need to decide it before the launch of yazi, which is inconvenient in real life.

## Potential Solution
Add a dedicated keybinding for this type of quitting. After this action, yazi will write the $PWD before it's launched. Users can input "q" to write the current directory in yazi to the cwd-file normally.

XOR-op:
Close #243

### DIFF
--- a/app/src/app.rs
+++ b/app/src/app.rs
@@ -25,8 +25,8 @@ impl App {
 
 		while let Some(event) = app.signals.recv().await {
 			match event {
-				Event::Quit(write_cwd_file) => {
-					app.dispatch_quit(write_cwd_file);
+				Event::Quit(no_cwd_file) => {
+					app.dispatch_quit(no_cwd_file);
 					break;
 				}
 				Event::Key(key) => app.dispatch_key(key),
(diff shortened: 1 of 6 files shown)


app/src/app.rs
```diff
@@ -41,9 +41,10 @@ impl App {
 		Ok(())
 	}
 
-	fn dispatch_quit(&mut self) {
+	fn dispatch_quit(&mut self, write_cwd_file: bool) {
 		if let Some(p) = &BOOT.cwd_file {
-			let cwd = self.cx.manager.cwd().as_os_str();
+			let cwd =
+				if write_cwd_file { self.cx.manager.cwd().as_os_str() } else { &BOOT.cwd.as_os_str() };
```
sxyazi, line 47: If the sole purpose is to avoid triggering `cd` not creating a `cwd-file` is a better choice compared to writing `cwd` into `cwd-file`, and `ya` wrapper doesn't execute `cd` when it detects the file not exist. Any other reasons?

app/src/app.rs
```diff
@@ -25,8 +25,8 @@ impl App {
 
 		while let Some(event) = app.signals.recv().await {
 			match event {
-				Event::Quit => {
-					app.dispatch_quit();
+				Event::Quit(write_cwd_file) => {
```
sxyazi, line 28: It's better to keep the same name with `no-cwd-file`

sxyazi:
Thanks for your PR; I made some reviews.

app/src/app.rs
```diff
@@ -200,7 +202,7 @@ impl App {
 						use std::os::unix::ffi::OsStrExt;
 						std::fs::write(p, paths.as_bytes()).ok();
 					}
-					return emit!(Quit);
+					return emit!(Quit(true));
```
sxyazi, line 205: Should be `false`

XOR-op: Fixed as well as another occurance

core/src/manager/manager.rs
```diff
@@ -113,7 +113,7 @@ impl Manager {
 		if self.tabs.len() > 1 {
 			return self.tabs.close(self.tabs.idx());
 		}
-		self.quit(tasks)
+		self.quit(tasks, true)
```
sxyazi, line 116: hmm and this one

XOR-op: Huh sorry I missed that

sxyazi:
I made some simplifications, and merged it now!

PR #249, opened 2023-10-08
avoid adding search to backstack

### LINKED ISSUE #248: Search result is lost after going back to it
Step to reproduce:
1. Search something with `s`
2. Use `h` to go to parent of a file
3. Use `H` to go back to the search result -> It's empty

I haven't look into the search functionality yet, but I think we can either store the search result or perform the search again.

ndtoan96:
Fix #248 

This PR remove search result from backstack.

### DIFF
--- a/core/src/manager/tab.rs
+++ b/core/src/manager/tab.rs
@@ -132,7 +132,9 @@ impl Tab {
 		}
 
 		// Backstack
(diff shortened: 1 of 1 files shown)


sxyazi:
Thank you. use `is_regular()` is better than `!is_search()`

PR #255, opened 2023-10-11
bugfix: cannot cd if there is whitespace prefix in path

ndtoan96:
How to reproduce:
1. `g <space>`
2. enter ` C:\<some path>` (notice the prefix whitespace)
3. cannot cd


sxyazi:
I generally prefer to do that [only where necessary](https://github.com/sxyazi/yazi/blob/15c34fed5c006c7b9b9152ffe6da32735ae77773/core/src/manager/tab.rs#L149), rather than modifying common functions, as this can better limit the scope of impact.

PR #264, opened 2023-10-13
fix: some colors not readable in light mode

sxyazi:
This PR is a follow-up to the https://github.com/sxyazi/yazi/pull/221


Brixy:
Just if I’m not too late to the party:

Input windows do not completely cover preview images, see screenshot below. Yet, they do cover preview text completely.

screenshot

magnetophon:
> Input windows do not completely cover preview images, see screenshot below. Yet, they do cover preview text completely.

Same here when using WezTerm.
When I use Allacritty with ueberzug, the image covers the input window.

Maybe you should open a separate issue for this, as it's not really related to the colors issue.

sxyazi:
This is a known issue specific to the Kitty graphics protocol. Other protocols like iTerm2 and Sixel do not have this problem. I haven't found the time to fix it yet, perhaps it requires a more subtle eraser...

PR #282, opened 2023-10-16
feat: shell completions & auto releasing

TD-Sky:
Now yazi has shell completion scripts. I package them into the released assets.

The publish actions will be triggered after you create a new tag. Here is a real action:

https://github.com/TD-Sky/yazi/actions/runs/6535605738

And here are the released assets:

https://github.com/TD-Sky/yazi/releases/tag/v0.1.5


app/Cargo.toml
uncenter, line 1: These changes are just formatting? Probably shouldn't be included in this PR.

TD-Sky: I commit it by mistake, sorry! The completion scripts should be generated under `app`  in old plan, so I changed `app/Cargo.toml`.

uncenter: No worries! I was just taking a look at the PR and it caught my eye. Nice work!

TD-Sky:
> if the completions can be generated, I wonder if it's worth committing them ? i.e., they can get out of date

It is not difficult at all to update them.

sxyazi:
Thank you!

PR #293, opened 2023-10-21
refactor(nix): add shell completion and avoid impure

XYenon:


flake.nix
```diff
@@ -2,16 +2,30 @@
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
     flake-utils.url = "github:numtide/flake-utils";
-    devenv = {
-      url = "github:cachix/devenv";
-      inputs.nixpkgs.follows = "nixpkgs";
+    rust-overlay = {
```
teto, line 5: we need the rust overlay because yazi relies on a more recent rust than what is nixpkgs ?

teto (approved):
thanks a lot, that worked .

flake.nix
XYenon: `rust-src` is needed to make rust-analyzer works, is there any way to get it from nixpkgs?

teto: rust-analyzer seems to be in nixpkgs but I dont have much experience with the whole process so I trust you :)

sxyazi:
Thanks, let me merge it now!

Issue #302, opened 2023-10-22
Add config support for coordinate and size adjustment for ueberzugpp image preview


15cm:
Background: https://github.com/sxyazi/yazi/issues/207

PR #304, opened 2023-10-22
Add config support for coordinate and size adjustment for ueberzugpp image preview

15cm:
Fix https://github.com/sxyazi/yazi/issues/302

Example usage in yazi.toml for a 2.0 scale monitor on Hyprland:
```
[preview.ueberzug]
scale_down_factor = 2.0
x_offset = 0.5
y_offset = 0.5
width_offset = -0.5
height_offset = -0.5
```

### DIFF
--- a/yazi-adaptor/src/ueberzug.rs
+++ b/yazi-adaptor/src/ueberzug.rs
@@ -1,8 +1,15 @@
+use std::cmp::max;
 use std::{path::PathBuf, process::Stdio};
+use tracing::debug;
 
 use anyhow::Result;
 use ratatui::prelude::Rect;
-use tokio::{io::AsyncWriteExt, process::{Child, Command}, sync::mpsc::{self, UnboundedSender}};
+use tokio::{
+	io::AsyncWriteExt,
+	process::{Child, Command},
+	sync::mpsc::{self, UnboundedSender},
+};
+use yazi_config::PREVIEW;
 
 use crate::Adaptor;
 
@@ -31,6 +38,19 @@ impl Ueberzug {
 		Ok(tx)
 	}
 
+	fn adjust_rect(mut rect: Rect) -> Rect {
+		let cfg = &PREVIEW.ueberzug;
+		rect.x =
+			max((rect.x as f64 / cfg.scale_down_factor + cfg.x_offset) as i32, 0) as u16;
+		rect.y =
+			max((rect.y as f64 / cfg.scale_down_factor + cfg.y_offset) as i32, 0) as u16;
+		rect.width =
+			max((rect.width as f64 / cfg.scale_down_factor + cfg.width_offset) as i32, 0) as u16;
+		rect.height =
+			max((rect.height as f64 / cfg.scale_down_factor + cfg.height_offset) as i32, 0) as u16;
+		return rect;
+	}
+
 	fn create_demon(adaptor: Adaptor) -> Result<Child> {
 		Ok(
 			Command::new("ueberzug")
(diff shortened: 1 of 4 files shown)


15cm:
Not sure if the doc site https://yazi-rs.github.io is auto generated or maintained somewhere else. It would be helpful if there is a CONTRIBUTING.md that provide minimum instructions for things like doc maintenance.

yazi-adaptor/src/ueberzug.rs
```diff
@@ -44,7 +57,10 @@ impl Ueberzug {
 
 	async fn send_command(child: &mut Child, cmd: Option<(PathBuf, Rect)>) -> Result<()> {
 		let stdin = child.stdin.as_mut().unwrap();
-		if let Some((path, rect)) = cmd {
+		if let Some((path, tmp_rect)) = cmd {
+			debug!("ueberzug rect before adjustment: {:?}", tmp_rect);
+			let rect = Self::adjust_rect(tmp_rect);
+			debug!("ueberzug rect after adjust adjustment: {:?}", rect);
```
15cm, line 63: typo, "adjust adjustment" should be "adjustment", fixing

15cm: Done

sxyazi:
Why did we need to adjust 0.5 cells here?

15cm:
Two reasons:
1. `13.0/2.0 == 7.5 as f64 == 7 as i32`. The X and Y could be smaller than expected and have overlaps with the left/top border. An alternative is to always round up the division result, but I'm afraid that it's not desired for everyone.
2. I enabled border and gaps in Hyprland [like this](https://github.com/15cm/nixfiles/blob/ec894322f82a29096794961ca1b6a9aa82bcf5c4/home/modules/programs/hyprland/hyprland.conf.jinja#L121-L123). Without an offset, the preview eats into the border a bit: 

My preference is to shift the preview away from the left and top border a bit. But I don't think there's a default value that works for everyone. So I rather expose it to allow users to perform minor adjustments.

This is a hot fix where we intercept in between ueberzugpp and wayland. It is hard to get it right. These options allow users to get a workable config at least for now.

sxyazi:
Oh I see, that makes sense. I've made some simplifications, would you like to give it a review?

As for the docs, it's placed at [yazi-rs/yazi-rs.github.io](https://github.com/yazi-rs/yazi-rs.github.io), you can click "Edit this page" at the bottom to navigate to the corresponding file. I'll add these two new options to the docs and have a CONTRIBUTING.md for it.

yazi-adaptor/src/ueberzug.rs
```diff
@@ -42,9 +44,24 @@ impl Ueberzug {
 		)
 	}
 
+	fn adjust_rect(mut rect: Rect) -> Rect {
+		let scale = PREVIEW.ueberzug_scale;
+		let (x, y, w, h) = PREVIEW.ueberzug_offset;
+
+		rect.x = 0f64.max(rect.x as f64 * scale + x) as u16;
```
15cm, line 51: I personally prefer the original division rather than the multiplication. The reason is that generally people find this config when they hit the issue with a monitor scale. A multiplication requires them to provide a `1/scale` as a float, which is kind of awkward.

yazi-config/preset/yazi.toml
```diff
@@ -9,10 +9,12 @@ show_hidden    = false
...
 [preview]
-tab_size   = 2
-max_width  = 600
-max_height = 900
-cache_dir  = ""
+tab_size        = 2
+max_width       = 600
+max_height      = 900
+cache_dir       = ""
+ueberzug_scale  = 1
```
15cm, line 16: I guess you want to keep all comments of the config in your website, which is ok for users but inconvenient for developers. Without context I have no way to tell the usage of `ueberzug_scale` and `ueberzug_offset`, as their name is not that intuitive, especially `ueberzug_offset` is an unnamed tuple. I'd suggest have comments somewhere in the code and keep them in sync with your doc.

sxyazi: The configuration for Yazi is currently centralized on a standalone website.

This is primarily done to avoid maintaining multiple copies of documentation simultaneously. Maybe in the future, we can explore generating documentation from the code, but for now, let's keep this approach.

yazi-adaptor/src/ueberzug.rs
sxyazi: I would like to keep it as a more general option for the long term, allowing users to freely zoom in or out on the image to address potential issues that may arise in the future, not just specific to this bug. At this point, using `scale<1` for zooming out and `scale>1` for zooming in seems more intuitive?

15cm: It makes sense to me then. We should mention in the doc that for now the scaling of ueberzug doesn't WAI so this option is needed to fixed that, with an example scale value to pass in.

After we get it fixed in ueberzug, I agree with you that we can live this option for the purpose you mentioned.

15cm:
LGTM. thx for the prompt responses.

sxyazi:
Thank you!

PR #314, opened 2023-10-24
fix: change the rust tool chain in `shell.nix` from stable to nightly

15cm:
The rustfmt.toml rules are only supported by rust nightly. With the current stable rust tool chain in flake.nix, `cargo fmt` gives me warnings like
```
Warning: can't set `wrap_comments = true`, unstable features are only available in nightly channel.
Warning: can't set `format_code_in_doc_comments = false`, unstable features are only available in n
```

The format result with those warnings is different from the current code base for almost all files. I guess the main contributors of the repo use nightly rust. If so, we should make it consistent in shell.nix to help other contributors onboard with the same env.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -11,7 +11,7 @@
     };
   };
 
-  outputs = { self, nixpkgs, flake-utils, rust-overlay, ... } @ inputs:
+  outputs = { self, nixpkgs, flake-utils, rust-overlay, ... }@inputs:
     let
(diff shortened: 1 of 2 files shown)


sxyazi:
Thank you for the PR!

Will these changes affect regular users? I've completely forgotten my Nix knowledge :D

sxyazi:
Thank you!

PR #316, opened 2023-10-24
feat: add an identifier before the extension to prevent file conflicts rather than after

rick-yao:
…ile and there is file with same name in folder

### DIFF
--- a/yazi-shared/src/fns.rs
+++ b/yazi-shared/src/fns.rs
@@ -44,16 +44,24 @@ pub fn expand_url(mut u: Url) -> Url {
 }
 
 pub async fn unique_path(mut p: Url) -> Url {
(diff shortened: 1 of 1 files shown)


yazi-shared/src/fns.rs
```diff
@@ -51,9 +51,9 @@ pub async fn unique_path(mut p: Url) -> Url {
 	let mut i = 0;
 	while fs::symlink_metadata(&p).await.is_ok() {
 		i += 1;
-		let mut name = name.clone();
-		name.push(format!("_{i}"));
-		p.set_file_name(name);
+		let name_str = name.as_os_str().to_str().unwrap();
```
sxyazi, line 54: Don't use `unwrap()`, that will causes panic

rick-yao:
@sxyazi I made a new commit using file_stem() and extension() , please kindly review it.

Issue #318, opened 2023-10-24
JSON preview using `jq` shows files as empty


LeXofLeviafan:
### What system are you running Yazi on?

Linux X11

### What terminal are you running Yazi in?

Konsole v23.08.1

### Yazi version

yazi 0.1.5

### Did you try the latest main branch to see if the problem has already been fixed?

Not tried, and I'll explain why below

### Describe the bug

When `tab_size` is set to `8` (which is the default tab width in a Linux terminal and therefore the preferred one when looking at file diffs for example), JSON files are displayed as empty.

This is because `--indent` option of `jq` [only allows values in range from `-1` to `7`](https://github.com/jqlang/jq/blob/jq-1.7/src/main.c#L482-L486) (`0` stands for "no whitespace", and `-1` stands for using tabs), and [yazi is neither aware of that nor does it have any fallback behaviour](https://github.com/sxyazi/yazi/blob/v0.1.5/core/src/external/jq.rs); this applies to the latest commit in `master` as well: https://github.com/sxyazi/yazi/blob/24fd587431e91ecb52344a285227497e4ec5d9d3/yazi-core/src/external/jq.rs#L10

(This arbitrary limit exists because for some inexplicable reason `jq` [uses bitmask flags to store indent width](https://github.com/jqlang/jq/blob/jq-1.7/src/jv.h#L217-L222)…)

Note that I have 2 JSON files which are getting displayed in preview with `tab_size` set to `8`; though considering they have a different colour scheme, I'm guessing `jq` is simply not applied to them (due to their huge size or perhaps the number of lines).

### Expected Behavior

Preview of a JSON file should display the file contents – at least in plain text if the parser failed to process it.

### To Reproduce

Set `tab_size` to `8` and navigate to a JSON file.

### Configuration

In `yazi.toml`:
```toml
[preview]
tab_size = 8
```

### Anything else?

When the problem is in the number of spaces, it's still possible to display the requested width by using requesting tabs and replacing each tab at the start with respective number of spaces (e.g. with regexps). Which is probably what the `tab_size` option is normally used for, anyway…

[...]

PR #320, opened 2023-10-25
fix: `jq` previews empty when the user sets `tab_size=8`

sxyazi:
Fixes https://github.com/sxyazi/yazi/issues/318

This PR also adds the functionality of fallbacking to the built-in highlighter when the user provides an invalid JSON file to `jq`.

### DIFF
--- a/yazi-core/src/external/jq.rs
+++ b/yazi-core/src/external/jq.rs
@@ -1,16 +1,16 @@
 use std::{path::Path, process::Stdio};
 
 use anyhow::Result;
-use tokio::{io::{AsyncBufReadExt, BufReader}, process::Command};
+use tokio::{io::{AsyncBufReadExt, AsyncReadExt, BufReader}, process::Command, select};
 use yazi_config::PREVIEW;
 use yazi_shared::PeekError;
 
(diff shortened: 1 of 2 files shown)


yazi-core/src/external/jq.rs
```diff
@@ -1,16 +1,16 @@
...
 use anyhow::Result;
-use tokio::{io::{AsyncBufReadExt, BufReader}, process::Command};
+use tokio::{io::{AsyncBufReadExt, AsyncReadExt, BufReader}, process::Command, select};
 use yazi_config::PREVIEW;
 use yazi_shared::PeekError;
 
 pub async fn jq(path: &Path, skip: usize, limit: usize) -> Result<String, PeekError> {
 	let mut child = Command::new("jq")
-		.args(["-C", "--indent", &PREVIEW.tab_size.to_string(), "."])
+		.args(["-C", "--indent", "-1", "."])
```
LeXofLeviafan, line 10: `--indent -1` can be replaced with a more explicit [`--tab`](https://github.com/jqlang/jq/blob/jq-1.7/src/main.c#L90)

sxyazi: Updated!

Issue #317, opened 2023-10-24
Running as sudo doesn't allow access to standard $USER folder


alvaro17f:
### What system are you running Yazi on?

Linux Wayland

### What terminal are you running Yazi in?

alacritty

### Yazi version

yazi 0.1.5

### Did you try the latest main branch to see if the problem has already been fixed?

Tried, but the problem is still present

### Describe the bug

When I run yazi as sudo, I am able to see root folders, but I'm not able to see my `/home/$USER` folder

### Expected Behavior

I would expect to be able to see every root directory that root have access to.

### To Reproduce

run: `sudo yazi` or `sudo -E yazi`

sxyazi:
When you run as root, `$USER=root`, are you sure that the `/home/root` directory exists?

alvaro17f:
No, I mean, if you run `sudo yazi`, you get accesss to `/root` but you're not able to acces anymore to `/home/$USER/`

alvaro17f:
https://github.com/sxyazi/yazi/assets/89421445/aabbf99f-26cc-4dcc-a5ce-fffda762f73b

sxyazi:
Please do a `ls -al /home/alvaro17f` and paste the output here.

sxyazi:
From the video, there may be an issue with `sudo`, as it significantly slows down the startup compared to launching without `sudo`, I'm not sure what happened behind

alvaro17f:
```sh
drwx--x--- alvaro17f alvaro17f 834 B  Wed Oct 25 11:08:32 2023  .
drwxr-xr-x root      root       18 B  Tue Oct 17 09:06:21 2023  ..
```

alvaro17f:
BTW, it's working fine with `joshuto` or `lf`

sxyazi:
Sadly I can't reproduce it and can't guess the possible reasons. Could you try to help debug it on your side?

alvaro17f:
I have found where does the problem comes from.
It only occurrs when I have a folder mounted with fuse inside of my user's folder.

sxyazi:
I made a PR https://github.com/sxyazi/yazi/issues/317, let me know if it works for you!

alvaro17f:
Reviewed 👍🏻

alvaro17f:
You're truly AWESOME.

Thank you @sxyazi

PR #321, opened 2023-10-25
fix: channel blockage caused by fuse permissions under the root user

sxyazi:
Fixes https://github.com/sxyazi/yazi/issues/317

### DIFF
--- a/yazi-core/src/files/files.rs
+++ b/yazi-core/src/files/files.rs
@@ -52,8 +52,10 @@ impl Files {
 			while let Ok(Some(item)) = it.next_entry().await {
 				select! {
 					_ = tx.closed() => break,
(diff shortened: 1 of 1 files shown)


alvaro17f (commented):
This could be a temporary fix, but now It doesn't show the fuse folder.

image

sxyazi:
No, this is not a temporary fix.

It's reasonable that it doesn't show up here because FUSE doesn't want it to be displayed. As far as I know, metadata for all files should not be denied, even if it lacks permissions. Even a file owned by the root user with permissions set to 000 should still be displayed correctly (metadata is not denied), but FUSE is denying it. I've also observed the same behavior in joshuto.

image

Issue #325, opened 2023-10-27
Please add binary releases for platform `linux/arm64` / Rust target `aarch64-unknown-linux-gnu`


nifr:
### Is your feature request related to a problem? Please describe

Currently there are no binary releases for Rust target `aarch64-unknown-linux-gnu`.

The following users would benefit
- users who want to run `yazi` in Docker containers on Apple Sillicon (M1, M2, ..)
- users who want to run `yazi` on i.e. Raspberry PIs

### Will you be willing to contribute this feature?

- [X] Yes, I will try to implement it

### Describe the solution you'd like

Target `aarch64-unknown-linux-gnu` is added to [.github/workflows/release.yml](https://github.com/sxyazi/yazi/blob/aceec545ae5c8a275e76f6fc3f1621ebb3c2a7f4/.github/workflows/release.yml#L15).

sxyazi:
Nice to have! I noticed that you have created a PR for it, let me know if it's ready for review!

sxyazi:
> Would be super nice if you could release a `v0.1.6` with the new binaries after merge.

So sorry Yazi is not yet ready to release v0.1.6 - we have not fulfilled [our promises to users for v0.1.6](https://github.com/sxyazi/yazi/milestone/5).

But I can edit the current v0.1.5 and add a "yazi-aarch64-unknown-linux-gnu.zip" attachment to it, not sure if it would be useful to you.

sxyazi:
Done! The new "yazi-aarch64-unknown-linux-gnu.zip" attachment has been updated to v0.1.5, and can be downloaded at https://github.com/sxyazi/yazi/releases/tag/v0.1.5

PR #326, opened 2023-10-27
ci: add Rust target "aarch64-unknown-linux-gnu" to release workflow

nifr:
This PR adds `aarch64-unknown-linux-gnu` as Rust target in order to have binary releases for `linux/arm64` in the future.

See #325


sxyazi:
Thank you for the PR, I'm delighted to see `linux/arm64` added and passing tests smoothly; I will merge it now!

PR #331, opened 2023-11-02
feat(yazi-plugin): retrieve HOME also on windows

sRavioli:
Right now Yazi only searches for the `$HOME` variable that is not present on windows. Either the user has to set `$env:HOME = "$env:USERPROFILE"` in his profile file or Yazi could also check for the `$USERPROFILE` variable.

My implementation of this functionality is pretty basic

### DIFF
--- a/yazi-plugin/preset/utils.lua
+++ b/yazi-plugin/preset/utils.lua
@@ -30,6 +30,7 @@ end
 
 function utils.readable_path(path)
 	local home = os.getenv("HOME")
+  home = home and home or os.getenv("USERPROFILE")
 	if home == nil then
 		return path
 	elseif string.sub(path, 1, #home) == home then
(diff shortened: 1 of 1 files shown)


yazi-plugin/preset/utils.lua
```diff
@@ -30,6 +30,7 @@ end
 
 function utils.readable_path(path)
 	local home = os.getenv("HOME")
+  home = home and home or os.getenv("USERPROFILE")
```
sxyazi, line 33: Maybe we can just `local home = os.getenv("USERPROFILE") or os.getenv("HOME")`?

sRavioli: yup, it's cleaner

sRavioli:
> BTW this was intentionally not implemented for Windows, as it seems that Windows lacks the concept of `~/` or even a root directory (`/`); it uses drive letter instead

true, windows is weird. On my shell (powershell-core) `~` gets expanded to `$env:USERPROFILE` but on cmd that does not happen. 

If i'm not mistaken windows should automatically set the userprofile variable: `%USERPROFILE%` on cmd and `$env:USERPROFILE` in powershell-core

sxyazi:
Thank you, LGTM!

Issue #340, opened 2023-11-05
yazi crashes one the limit of open files is reached


teto:
### What system are you running Yazi on?

Linux Wayland

### What terminal are you running Yazi in?

kitty 0.30.1 

### Yazi version

yazi-0.1.5pre20231027_396f60d

### Did you try the latest main branch to see if the problem has already been fixed?

Tried, but the problem is still present

### Describe the bug

when I've got a lot of stuff going on yazi fails with
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: Io(Os { code: 24, kind: Uncategorized, message: "Too many open files" }), paths: [] }', yazi-core/src/manager/watcher.rs:64:48
                    stack backtrace:
                                       0: rust_begin_unwind
                                                              1: core::panicking::panic_fmt
                                                                                              2: core::result::unwrap_failed
                                                                                                                               3: yazi_core::context::Ctx::make
                                                                                                                                                                  4: yazi::main
                                                                                                                                                                               note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
                                                                      zsh: abort (core dumped)  yazi
```
This seems expected since as soon as I run `ulimit -n 4096` (my default is 1024) yazi works again. Not sure what a proper fix is:
- dont crash and exit with the suggestion of using `limit -n`
- disable filewatching ?

### Expected Behavior

no crash

### To Reproduce

open tons of file descriptor (1024)

sxyazi:
Hi I made a PR to address it, https://github.com/sxyazi/yazi/pull/342, let me know if it works for you!

SolitudeSF:
what is the reason for opening so many fds?

teto:
> what is the reason for opening so many fds?

I build a software + I have firefox open. It (seemingly) never posed any issue apart from yazi.

SolitudeSF:
im asking why would yazi need to have so many fds open.

PR #342, opened 2023-11-06
feat: raise open file descriptors limit at startup

sxyazi:
Close https://github.com/sxyazi/yazi/issues/340


teto:
What does the PR do ? does it try to raise the number of open file or does it better address the error ? (sry not clear from my limited experience with rust)

sxyazi:
Yes, this PR tries to increase the number of open files, and this is a better approach.

Because that error is not triggered after watching the files but rather during the initialization of the watcher, i.e. it occurs before the actual watching begins, which means the same error will occur with all I/O operations.

teto:
I am dubious this is the correct fix: the max-open-file property is a system property that should be controlled by the user, not the program. 
Also this just delays the issue: depending on configuration raising the number may fail, if I run even more services, yazi will fail as it does today.
Is the file watcher necessary for yazi or can it work in a degraded state ? otherwise I would suggest as a fix to just fail more gracefully with a message "please raise number of open files with ulimit -n XXXX or close file handles" ?

sxyazi:
While I'm not sure how you encountered this issue (Are there any steps I can reproduce?), seems to me that it's not specific to the watcher but rather affects all I/O operations within Yazi.

When this error occurs, it not only impacts the watcher but even regular directory listings are affected. Currently, it just becomes apparent during watcher initialization and causes the program to exit prematurely. For apps like Yazi that utilize multi-threading to accelerate processing, a default soft limit of 256 seems quite limiting for large directories.

I'm not quite sure what you mean by "raising the number may fail.", are you saying that if a user sets the system's hard limit to a very small value, the increase is ineffective? I believe that would be an issue with the user's environment. The hard limit should typically be set to "unlimited" or a very large value, and it's unlikely to be exceeded. Please correct me if I'm wrong!

CleanShot 2023-11-06 at 19 22 14

teto:
> I'm not quite sure what you mean by "raising the number may fail.", are you saying that if a user sets the system's hard limit to a very small value, the increase is ineffective? I believe that would be an issue with the user's environment. 

I mean exactly that: it can fail and all that can fail will fail. 

> The hard limit should typically be set to "unlimited" or a very large value, and it's unlikely to be exceeded. Please correct me if I'm wrong!

well I dunno about that, and I doubt it. I think changing this value is not for yazi to decide since it has sideeffects that can affect the user system in a way the user might not realize (more fds => strain resources). IMO it would be best to not change the value but gracefully exit with a message on how to increase that number ?

Btw, I've my valuie from 1024 to 30k

sxyazi:
The value is specific to the process, and what this PR does is modify Yazi's own maximum available limit. Prompting the user to do a `ulimit -n 1024` manually is not a reasonable approach -- It requires the user to do this every shell session before using Yazi.

Even if it's done by configuring it permanently, it's still not a suitable method. Some single-threaded apps don't need such a concurrency, and it can impact the overall system performance. I believe this should be the responsibility of the concurrent app itself.

It's just making a `setrlimit` system call and passing a completely legal value, I'm not sure why it would fail. Maybe the user has disabled this system call? That would be an issue with the user's environment.

teto:
ok, seems legit in retrospect 

> It's just making a setrlimit system call and passing a completely legal value, I'm not sure why it would fail. 

if the request exceeds the hardlimit it could fail so a proper error message and exit would be neat in that case

sxyazi:
> if the request exceeds the hardlimit it could fail so a proper error message and exit would be neat in that case

No, it won't exceed the hard limit. The crate takes its maximum upper bound, so it should always succeed unless the user has explicitly disabled that system call: https://github.com/paritytech/fdlimit/blob/eee618fbe779452099f1857dd9191ec711067af6/src/lib.rs#L91-L93

Since this error only occurs during I/O operations and is not specific to the watcher, any I/O operation can potentially encounter "Too many open files."

One possible approach is to add this check for all I/O operations, but that would be a substantial amount of work, and some I/O operations occur in async contexts where it's not practically feasible. I'm not sure if there's a better solution than simply raising the app's own limit, which should work in most cases.

teto:
thanks for the explanation. Sounds good enough. Also now users can search and find this PR and ticket so a solution is easier for everyone. Go ahead !

PR #345, opened 2023-11-06
feat: Add basic emac keybindings to input

### LINKED ISSUE #343: Support for more editing commands in input box

I frequently use the jump to start or end of line, clear to start, and forward delete in the terminal. It would be great if these worked in addition to the vim keybindings. 

- [x] Yes, I will try to implement it

I would like to be able to use `ctrl+a` to jump to the start of the line, `ctrl+e` to the end, `ctrl+d` to delete a single character (to the right of the cursor), `alt+backspace` to delete a word, and `alt+d` to delete a word to the right of the cursor. 

_No response_

### Additional context

It looks to me that there is already support for handling `modifier+key` via crossterm in the input handling code. Additionally, it seems like most of the desired functionality is already implemented, and the rest of the functionality would be similar to what is already there. 

If this sounds like it would be useful to others as well and would not contribute to much to the complexity of the codebase, I can try to take a crack at it.

Tweoss:
Adds 
- [x] `ctrl+a`: go to start of line
- [x] `ctrl+e`: go to end of line
- [x] `ctrl+b`: go back one character
- [x] `ctrl+f`: go forward one character
- [x] `ctrl+h`: delete one character (normal backspace)
- [x] `ctrl+d`: delete one character (forward backspace)
- [x] `alt+b`: go one word left
- [x] `alt+f`: go one word right
- [x] `ctrl+u`: delete to start of line
- [x] `ctrl+k`: delete to end of line
- [x] `alt+backspace`, `ctrl+w`: delete one word left
- [x] `alt+d`: delete one word right

Resolves #343

### DIFF
--- a/yazi-core/src/input/input.rs
+++ b/yazi-core/src/input/input.rs
@@ -56,11 +56,17 @@ impl Input {
 		}
 
 		match key {
-			Key { code: KeyCode::Backspace, shift: false, ctrl: false, alt: false } => self.backspace(false),
+			Key { code: KeyCode::Backspace, shift: false, ctrl: false, alt: false } => self.backspace(),
 			// Handle Emacs-style keybindings.
-			Key { code: KeyCode::Char('a'), shift: false, ctrl: true, alt: false } => self.move_(isize::MIN),
-			Key { code: KeyCode::Char('e'), shift: false, ctrl: true, alt: false } => self.move_(isize::MAX),
-			Key { code: KeyCode::Char('d'), shift: false, ctrl: true, alt: false } => self.backspace(true),
+			Key { code: KeyCode::Char('a'), shift: false, ctrl: true, alt: false } => {
+				self.move_(isize::MIN)
+			}
+			Key { code: KeyCode::Char('e'), shift: false, ctrl: true, alt: false } => {
+				self.move_(isize::MAX)
+			}
+			Key { code: KeyCode::Char('d'), shift: false, ctrl: true, alt: false } => {
+				self.forward_delete()
+			}
 			Key { code: KeyCode::Char('b'), shift: false, ctrl: false, alt: true } => self.backward(),
 			Key { code: KeyCode::Char('f'), shift: false, ctrl: false, alt: true } => self.forward(false),
 			_ => false,
@@ -80,26 +86,29 @@ impl Input {
 		true
 	}
 
-	pub fn backspace(&mut self, forward_delete: bool) -> bool {
+	pub fn forward_delete(&mut self) -> bool {
 		let snap = self.snaps.current_mut();
-		if forward_delete {
-			// Return false when there is no character on the right to delete.
-			// Note that the cursor can be at index `snap.value.len()` when in
-			// edit mode.
-			if snap.cursor > snap.value.len() - 1 {
-				return false;
-			} else {
-				snap.value.remove(snap.idx(snap.cursor).unwrap());
-			}
-			self.move_(0);
+		// Return false when there is no character on the right to delete.
+		// Note that the cursor can be at index `snap.value.len()` when in
+		// edit mode.
+		if snap.cursor > snap.value.len() - 1 {
+			return false;
 		} else {
-			if snap.cursor < 1 {
-				return false;
-			} else {
-				snap.value.remove(snap.idx(snap.cursor - 1).unwrap());
-			}
-			self.move_(-1);
+			snap.value.remove(snap.idx(snap.cursor).unwrap());
 		}
+		self.move_(0);
+		self.flush_value();
+		true
+	}
+
+	pub fn backspace(&mut self) -> bool {
+		let snap = self.snaps.current_mut();
+		if snap.cursor < 1 {
+			return false;
+		} else {
+			snap.value.remove(snap.idx(snap.cursor - 1).unwrap());
+		}
+		self.move_(-1);
 
 		self.flush_value();
 		true
--- a/yazi-core/src/input/input.rs
+++ b/yazi-core/src/input/input.rs
@@ -57,7 +57,7 @@ impl Input {
 	///
 	/// Otherwise, returns how many characters to move to reach right *AFTER* the
 	/// word boundary, or the end of the iterator.
-	fn find_word_boundary(
+	pub(super) fn find_word_boundary(
 		input: impl Iterator<Item = char> + Clone,
 		stop_before_boundary: bool,
 	) -> usize {
@@ -81,18 +81,6 @@ impl Input {
 		spaces_count + character_count
 	}
 
-	pub(super) fn move_word(&mut self, forwards: bool, end: bool) -> bool {
-		let snap = self.snap();
-		let idx = snap.idx(snap.cursor).unwrap_or(snap.len());
-
-		let movement = if forwards {
-			Self::find_word_boundary(snap.value[idx..].chars(), end) as isize
-		} else {
-			-(Self::find_word_boundary(snap.value[..idx].chars().rev(), false) as isize)
-		};
-		self.move_(movement)
-	}
-
 	fn delete_range(&mut self, range: impl RangeBounds<usize>) {
 		let snap = self.snap_mut();
 		snap.cursor = match range.start_bound() {
@@ -114,11 +102,10 @@ impl Input {
 			return self.type_str(c.encode_utf8(&mut bits));
 		}
 
-		use KeyCode as K;
-		use KeyCode::Char as C;
+		use KeyCode::{Backspace, Char as C, Delete};
 
 		match key {
-			Key { code: K::Backspace, shift: false, ctrl: false, alt: false } => self.backspace(),
+			Key { code: Backspace, shift: false, ctrl: false, alt: false } => self.backspace(),
 			// Handle Emacs-style keybindings.
 			Key { code: C('a'), shift: false, ctrl: true, alt: false } => self.move_(isize::MIN),
 			Key { code: C('e'), shift: false, ctrl: true, alt: false } => self.move_(isize::MAX),
@@ -126,8 +113,8 @@ impl Input {
 			Key { code: C('f'), shift: false, ctrl: true, alt: false } => self.move_(1),
 			Key { code: C('h'), shift: false, ctrl: true, alt: false } => self.backspace(),
 			Key { code: C('d'), shift: false, ctrl: true, alt: false } => self.forward_delete(),
-			Key { code: C('b'), shift: false, ctrl: false, alt: true } => self.move_word(false, false),
-			Key { code: C('f'), shift: false, ctrl: false, alt: true } => self.move_word(true, false),
+			Key { code: C('b'), shift: false, ctrl: false, alt: true } => self.backspace(),
+			Key { code: C('f'), shift: false, ctrl: false, alt: true } => self.forward(false),
 			Key { code: C('u'), shift: false, ctrl: true, alt: false } => {
 				let snap = self.snap_mut();
 				let end = snap.idx(snap.cursor).unwrap_or(snap.len());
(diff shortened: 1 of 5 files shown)


Hanaasagi:
It would be great to add `ctrl-h` for deleting character before the cursor.

musjj:
It'd be nice if we can have `Ctrl+W` (delete word backwards), it's a really basic one that I use pretty often. `Ctrl+U` (delete to start) and `Ctrl+K` (delete to end) would be desirable too.

There's this reference for the rest of the readline keybindings:
https://github.com/chzyer/readline/blob/master/doc/shortcut.md

Tweoss:
@musjj I've implemented most of those keybinds now, except transpose and the ones related to multiple lines. The ones you've explicitly listed should be implemented

yazi-core/src/input/input.rs
```diff
@@ -221,15 +271,30 @@ impl Input {
 		true
 	}
 
+	pub fn forward_delete(&mut self) -> bool {
+		let snap = self.snaps.current_mut();
+		// Return false when there is no character on the right to delete.
+		// Note that the cursor can be at index `snap.value.len()` when in
+		// edit mode.
+		if snap.cursor > snap.value.len() - 1 {
```
sxyazi, line 279: Pressing `Ctrl-D` when the input is empty will result in a process crash

yazi-fm/src/executor.rs
```diff
@@ -282,8 +282,8 @@ impl<'a> Executor<'a> {
 				"insert" => self.cx.input.insert(exec.named.contains_key("append")),
 				"visual" => self.cx.input.visual(),
 
-				"backward" => self.cx.input.backward(),
-				"forward" => self.cx.input.forward(exec.named.contains_key("end-of-word")),
+				"backward" => self.cx.input.move_word(false, false),
```
sxyazi, line 285: We need to keep the function name matches the command name, which allows us to implement some validation through macros like:

https://github.com/sxyazi/yazi/blob/521c8b6bfefbaa70b8564812ee377f219871bbe1/yazi-fm/src/executor.rs#L223-L234

yazi-core/src/input/input.rs
```diff
@@ -143,53 +143,63 @@ impl Input {
...
+		input: impl Iterator<Item = char> + Clone,
+		stop_before_boundary: bool,
+	) -> usize {
+		// Move until we don't see any more whitespace.
+		let spaces_count = input.clone().take_while(|c| CharKind::new(*c) == CharKind::Space).count();
+		let input = input.skip(spaces_count);
+
+		// If we want the *NEXT* end of word, then we want to skip the current
+		// character.
+		let mut input = input.skip(if stop_before_boundary { 1 } else { 0 }).peekable();
```
sxyazi, line 168: Can be simplified as `stop_before_boundary as usize`

yazi-core/src/input/input.rs
Tweoss: I didn't think about that, thanks for catching!

sxyazi:
Hmm I can't build it, it's throwing a lot of errors for now

Tweoss:
Ah, I see. I didn't realize the file structure for the commands had changed in #338 and made a really bad merge. It might be a good idea to abort and restart from the new structure.

sxyazi:
Can you reset it to before the merge? and then I'll try to resolve the conflicts.

Tweoss:
Okay, I think this builds (I tried with a clean clone) and has the fixes

sxyazi:
I found an inconsistent behavior where `forward --end-of-word` should stop at the end of each word, but it stops at spaces when they exist between words for now:

https://github.com/sxyazi/yazi/assets/17523360/190bf850-2614-4573-99c9-1b4568aff53f

I tried modifying the `stop_before_boundary` parameter of `find_word_boundary`, but couldn't make it work.

Tweoss:
I see why that's happening, I think the fix would just be to move the line to that handles moving to the end of the word to be earlier. I can fix that now

sxyazi:
Okay please give it a fix!

Tweoss:
There appears to be inconsistency with spaces when skipping to start of words as well, let me resolve that

Tweoss:
Fixed! (hopefully)

https://github.com/sxyazi/yazi/assets/70821802/174d2dfc-fc54-4915-966b-4feeaa283603

sxyazi:
Thank you, let me test again!

Tweoss:
I'll also test the new keybinds, I think something may be strange with going back a word.

sxyazi:
There is another inconsistency issue with `de` (`delete` and `forward --end-of-word`), it should yield ` ce ef` instead of `b ce ef`:


https://github.com/sxyazi/yazi/assets/17523360/7bb09e61-3f63-42e7-8995-992286849816

Tweoss:
Oh nooo! I think I understand. When in vim mode, moving to the end of the word `Ab` when the cursor is at `A` entails moving one right => `aB`. However, when deleting, we want to move two to the right. When in insert mode, this is not an issue. `|ab` to `ab|` is the same whether deleting or moving. I guess there's no way to work around this but to add a special case for vim mode deletion or not.

Tweoss:
This is what it looks like now, when hitting `d+e` repeatedly.

https://github.com/sxyazi/yazi/assets/70821802/8a6dd282-546b-4a07-b6eb-a8b03d669444

sxyazi:
I generally don't want to treat it as a special case; it requires us to consider all scenarios (currently `ye` still doesn't work), which adds a burden when we later add other operations.

Can we further simplify `find_word_boundary` and only use it for the new Emacs functionality, without modifying the existing `forward` and `backward` Vi functionalities? This way we can enjoy the new features without losing the efforts we've already put into the existing ones. 😃

Tweoss:
Yes, I don't really like special casing based on operations. I was hoping to reuse the code for the forward + backward to simplify the keyboard shortcuts and thought using the same function might be quite elegant.

Now that you point out `ye` not working, I realize that no operations is the only case where we want to set `stop_before_boundary` to true. In other words, there ought to be no additional burden in adding operations. 

However, I would note that the emacs keybinds only work in insert mode, so it could make sense to separate code as well as functionality (unless in the future we want emacs in normal mode as well).

Regardless, if you think find_word_boundary should be restricted to Emacs functionality, then we should definitely do that.

sxyazi:
> However, I would note that the emacs keybinds only work in insert mode, so it could make sense to separate code as well as functionality (unless in the future we want emacs in normal mode as well).

Yes, I think that's reasonable. The Vi mode essentially serves as an extension of the basic Emacs, and for Vi, it's not possible to use Emacs key bindings in normal mode, I think we should keep this behavior consistent.

Also since these key bindings are hard-coded, makes it easier to distinguish them from user-configurable keybindings, thus avoiding potential conflicts.

Tweoss:
Sounds good! I reverted the changes to forward and backward and removed the unnecessary arguments from find_word_boundary.

yazi-core/src/input/input.rs
```diff
@@ -55,12 +102,79 @@ impl Input {
...
+				self.move_(step as isize)
+			}
+			Key { code: C('u'), shift: false, ctrl: true, alt: false } => {
+				let snap = self.snap_mut();
+				let end = snap.idx(snap.cursor).unwrap_or(snap.len());
+				self.delete_range(..end);
+				true
+			}
+			Key { code: C('k'), shift: false, ctrl: true, alt: false }
+			| Key { code: Delete, shift: false, ctrl: false, alt: false } => {
```
sxyazi, line 137: Is this `Delete` should be added for `Ctrl-u` instead of `Ctrl-k`? If so I will fix it.

Tweoss: I `believe` that Delete cuts to the end of line, based off what iTerm sends. I'll look into it real quick

musjj:
Hey, I haven't tested this PR personally, but from looking at the video it looks like that the movements feels inconsistent from readline/emacs. OBS is broken for me, but this is how it works in bash and other readline-based programs.

With `Alt-F`, from left to right:
```bash
ab bc cd ab-bc-cd
^ ^  ^  ^  ^  ^  ^
```

With `Alt-B`, from right to left:
```bash
ab bc cd ab-bc-cd
^  ^  ^  ^  ^  ^ ^
```

(the `^` represents the position of the block cursor)

You can check the behavior by simply running `bash`. Note that the block cursor is "left-oriented" (not sure if it's the right term) as in the actual cursor position is located on the left side of the block. You can run `echo -e -n "\x1b[\x36 q"` to make the behavior clearer.

yazi-core/src/input/input.rs
Tweoss: This is from nushell's listening to keybinds.

image

In `reedline`, however, the `Delete` key maps to forward delete.

I'm not sure now what the behavior should be. Perhaps it would be better to leave `Delete` out.

sxyazi: Okay I'll remove it first

musjj:
Captured some gifs for clarity:

block
beam

sxyazi:
https://github.com/sxyazi/yazi/assets/17523360/5d2e5f6c-dea0-4d5b-8410-39c8085801b5

Seems we need ignore all characters other than letters when navigating words in Emacs

Tweoss:
FWIW, this is how nushell (so reedline) and Google Docs handles word boundaries.

https://github.com/sxyazi/yazi/assets/70821802/85586e2c-db59-4741-a2ef-a0fa798acf98

https://github.com/sxyazi/yazi/assets/70821802/e487c24e-dcc4-46ce-aef2-e8db843e60dd

MacOS seems to adhere to bash rules. 



https://github.com/sxyazi/yazi/assets/70821802/7250218b-16ce-4798-8269-f54345b4a768

musjj:
I'm surprised that reedline works differently :thinking:. But if we're aiming to stay true to readline then bash's behavior is the way to go. Maybe additional options can be added later on to customize this sort of things?

sxyazi:
Ah I just realized there are different variants of Emacs implementations.

I'm not sure which one is correct, since I don't use all Emacs keybindings. @15cm, do you have any suggestions? I remember you are an Emacs user. 🤣

15cm:
I am a dedicated vim/evil mode user so I don't have the answer. But I found a [cheat sheet](https://gist.githubusercontent.com/anthonyclarka2/749d89c94c6229fabe8cd00d5fb96a3a/raw/d1fa1691adbde3e7e876b389787ed2faab823d32/emacs_navigation_cheat_sheet.svg) for you.

sxyazi:
I couldn't find the specific implementation specifications for `Alt-F` and `Alt-B`, so I've decided to keep the current behavior.

It doesn't seem to have major issues, just a finer granularity compared to bash (treating `-` as a word boundary, not just spaces). This may be more convenient when editing filenames like `foo-bar`, but it could also be considered a bug to be fixed in future PRs.

Anyway, let's go ahead and merge it for now!

Tweoss:
Just checking @sxyazi , you'll merge the PR? (I don't have write access to the repository)

Also, thank you for spending time on this PR and giving me pointers. The ability to manipulate the input in insert mode feels liberating, and I can't wait to install this from main on my local machine.

sxyazi:
Yeah I will merge it now. I went to sleep after replying yesterday 🤣

sxyazi:
Merged! Thank you very much for contributing this exciting feature! 🎉

sxyazi:
Hey @teto, your [requested feature](https://github.com/sxyazi/yazi/issues/51#:~:text=sorting%0ANo%20request-,Emacs,-readline%20keybindings%20for) has been done by this PR, give it a shot!

Brixy:
Hi guys,

Thank you very much! This makes renaming a lot easier.

With `ctrl + u` yazi crashes for some files. I am not quite sure, but long file names seem to be the problem here.

This is the error message:

```
The application panicked (crashed).
  called `Option::unwrap()` on a `None` value
in /home/markus/data/foreign/github/yazi/yazi-core/src/input/snap.rs, line 82
thread: main
```

sxyazi:
@Brixy Thanks for reporting, I think https://github.com/sxyazi/yazi/commit/027c7a79c18fe7ad0461bc42fa395543df4fb0f8 should fix it.

PR #349, opened 2023-11-07
fix: precache n-1 and n+1 pages

### LINKED ISSUE #348: Some results are not colorized when searching

Linux X11

alacritty 0.12.3 (5efb0690)

### Yazi version

yazi 0.1.5

### Did you try the latest main branch to see if the problem has already been fixed?

Tried, but the problem is still present

### Describe the bug

https://github.com/sxyazi/yazi/assets/9482395/81948f8f-e94c-4c36-884a-f21d3a58c5f8

As shown in the video, when search results cause the current page to scroll, this part is not rendered in color. It's white, not blue. When I scroll down, the color is rendered.

### Expected Behavior

All results is colorized.

### To Reproduce

Use a search that can match many results.

### Configuration

no config.

### Anything else?

_No response_

Hanaasagi:
Close #348 

### What does this PR do?

- Precache n+1/n-1 page for better experience. (n - 1 also has this problem, run `yazi` and press `G`)

### Test

https://github.com/sxyazi/yazi/assets/9482395/f4d1127e-588a-49e7-bc03-2d62c91c2767

### DIFF
--- a/yazi-core/src/tab/folder.rs
+++ b/yazi-core/src/tab/folder.rs
@@ -43,16 +43,7 @@ impl Folder {
 		self.prev(Default::default());
 
 		if self.page == old {
-			emit!(Pages(self.page)); // Force update
-			match MANAGER.layout.folder_height() {
-				limit if limit > 0 => {
-					let max_page = (self.files.len() + limit - 1) / limit;
-					if self.page + 1 < max_page {
-						emit!(Pages(self.page + 1));
-					}
-				}
-				_ => {}
-			}
+			self.set_page(true); // Force update
 		}
 
 		true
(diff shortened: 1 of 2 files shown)


sxyazi:
Thanks for your PR!

I've changed it to use a bound as the condition, which should make it more precise. Would you like to give it a review?

Hanaasagi:
Thanks for your help. And I found another case that hasn't been fixed. It seems that the n+1 is not cached after the initial load. I push a new commit for this.

https://github.com/sxyazi/yazi/assets/9482395/90b77ab8-0abf-4315-9f5f-b8e0261ca6b4

sxyazi:
So weird, I can't reproduce it, not sure what's going wrong...

sxyazi:
Ah, I reproduced it when I enlarged my terminal window. I created a new commit that should fix it.

Hanaasagi:
Tested it locally, and it works as expected. Thank you! :tada:

sxyazi:
Okay, let's merge it now. Thanks for your test!

PR #352, opened 2023-11-09
fix: prefix matching should have higher priority in completion

XOR-op:
Currently, the completion logic only use `contains` as the criteria. However, an ergonomic completion should prioritize candidates having the exact prefix with the input. This PR fixes this problem.

### DIFF
--- a/yazi-core/src/completion/commands/show.rs
+++ b/yazi-core/src/completion/commands/show.rs
@@ -36,21 +36,43 @@ impl Completion {
 			return false;
 		};
 
-		let flow = cache.iter().try_fold(Vec::with_capacity(30), |mut v, s| {
-			if s.contains(opt.word) && s != opt.word {
+		let candidate_size = 30;
+
+		// prioritize those with exact prefix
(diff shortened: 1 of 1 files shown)


yazi-core/src/completion/commands/show.rs
```diff
@@ -36,21 +36,43 @@ impl Completion {
...
-
-		self.ticket = opt.ticket;
-		self.cands = match flow {
+		let mut cand = match exact_flow {
 			ControlFlow::Continue(v) => v,
 			ControlFlow::Break(v) => v,
 		};
+
+		let fuzzy_flow =
+			cache.iter().try_fold(Vec::with_capacity(candidate_size - cand.len()), |mut v, s| {
```
sxyazi, line 57: I prefer to only go through a 1-pass process i.e. merging two `fold`s into one.

XOR-op: Updated with one pass logic

sxyazi:
Thank you, merged!

PR #353, opened 2023-11-09
feat: make width of completion relative to input

XOR-op:
Allow displaying more contents without breaking visualization


sxyazi:
Thank you!

PR #377, opened 2023-11-18
feat: Add more optimization options

Walker-00:
Well I just added additional optimizations options for file size.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,8 @@ resolver = "2"
 members  = [ "yazi-*" ]
 
 [profile.release]
+codegen-units = 1
+opt-level = 3
 lto   = true
 strip = true
 panic = "abort"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -3,6 +3,8 @@ resolver = "2"
 members  = [ "yazi-*" ]
 
 [profile.release]
+codegen-units = 1
+opt-level = 3
```
sxyazi, line 7: Seems `opt-level = 3` is the default value of releasing, https://doc.rust-lang.org/cargo/reference/profiles.html#release

Walker-00: ahh yea.

sxyazi: Okay, I'll only keep `codegen-units`.

sxyazi:
Thank you!

PR #385, opened 2023-11-21
feat: better handling of killing path separator

XOR-op:
Better handling of killing trailing path separator.

Example:
```bash
.config/yazi/ # original input
# press ctrl-w
.config/yazi  # w/o this PR
.config/      # w/ this PR
```

### DIFF
--- a/yazi-core/src/input/commands/kill.rs
+++ b/yazi-core/src/input/commands/kill.rs
@@ -4,13 +4,23 @@ use yazi_shared::{event::Exec, CharKind};
 
 use crate::input::Input;
 
+#[derive(Debug, Copy, Clone)]
+enum KillMode {
+	Default,
+	BigWord,
+}
+
 pub struct Opt<'a> {
 	kind: &'a str,
+	mode: KillMode,
 }
 
 impl<'a> From<&'a Exec> for Opt<'a> {
 	fn from(e: &'a Exec) -> Self {
-		Self { kind: e.args.first().map(|s| s.as_str()).unwrap_or_default() }
+		Self {
+			kind: e.args.first().map(|s| s.as_str()).unwrap_or_default(),
+			mode: e.named.get("big-word").map_or(KillMode::Default, |_| KillMode::BigWord),
+		}
 	}
 }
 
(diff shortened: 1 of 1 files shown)


sxyazi:
Can we add it behind a parameter like `kill forward --big-word`?

XOR-op:
Sorry for the late update since I have a busy month. Now this behavior is controlled by the argument.

yazi-core/src/input/commands/kill.rs
```diff
@@ -4,13 +4,23 @@ use yazi_shared::{event::Exec, CharKind};
...
 	kind: &'a str,
+	mode: KillMode,
 }
 
 impl<'a> From<&'a Exec> for Opt<'a> {
 	fn from(e: &'a Exec) -> Self {
-		Self { kind: e.args.first().map(|s| s.as_str()).unwrap_or_default() }
+		Self {
+			kind: e.args.first().map(|s| s.as_str()).unwrap_or_default(),
+			mode: e.named.get("big-word").map_or(KillMode::Default, |_| KillMode::BigWord),
```
sxyazi, line 22: Is better to impl the `FromStr` (also `Default`) trait for the `KillMode` to do this.

yazi-core/src/input/commands/kill.rs
```diff
@@ -62,8 +72,12 @@ impl Input {
 			input.take_while(|&c| CharKind::new(c) == first).count() + 1
 		}
 
-		let spaces = count_spaces(input.clone());
-		spaces + count_characters(input.skip(spaces))
+		let space_or_trailing_slash = match input.clone().next() {
+			Some(std::path::MAIN_SEPARATOR) if matches!(mode, KillMode::BigWord) => 1,
+			Some(c) if CharKind::new(c) == CharKind::Space => count_spaces(input.clone()),
```
sxyazi, line 77: It seems that `if CharKind::new(c) == CharKind::Space` is unnecessary.

sxyazi:
I'm going to close this because it still needs some work and hasn't seen any activity recently. @XOR-op thanks for the effort!

PR #386, opened 2023-11-21
feat: cursor and page key navigation parity with Vim bindings

flexiondotorg:
This pull request adds additional keybindings for the cursor and page keys so that luddites such as myself 😉 can navigate `yazi` in much the same way as those who use Vim-style bindings.


sxyazi:
LGTM, thank you!

PR #393, opened 2023-11-22
feat: include commit hash in yazi --version

Hanaasagi:
Just an example.

```
>>​ cargo run -- --version
    Finished dev [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/yazi --version`
yazi 0.1.5 (43fb9fe3f 2023-11-22)
```

NOTE: This PR does not include changes to Cargo.lock. I think this should be committed by the maintainer.

Ref: https://github.com/rustyhorde/vergen


sxyazi:
Awesome!

Issue #390, opened 2023-11-22
The sorting of the list after renaming is confusing


Hanaasagi:
### What system are you running Yazi on?

Linux X11

### What terminal are you running Yazi in?

alacritty 0.12.3 (5efb0690)

### Yazi version

main 5439c80a03987fe286e721e0025b5ad4c4105615

### Did you try the latest main branch to see if the problem has already been fixed?

Tried, but the problem is still present

### Describe the bug

https://github.com/sxyazi/yazi/assets/9482395/8eff3314-6b25-4a38-bbd9-6b79aa4ad070

some items move to middle, some else move to bottom.

### Expected Behavior

Not sure, but I think if the filenames are same, the position should not be changed.

### To Reproduce

Move to item and press `r` then `enter`. repeat...

### Configuration

yazi.toml
```toml
[manager]
# layout         = [ 0, 3, 7 ]
layout         = [ 1, 4, 3 ]
sort_by        = "modified"
sort_sensitive = true
sort_reverse   = true
sort_dir_first = true
linemode       = "none"
show_hidden    = false
show_symlink   = true

```

keymap.toml

```toml
[manager]

keymap = [

	{ on = [ "r" ],         exec = "rename --force",                                               desc = "Rename a file or directory" },
]

```

sxyazi:
Weird, I can't reproduce this issue on my side by using your config...


https://github.com/sxyazi/yazi/assets/17523360/c1971505-87a7-4735-8cf2-619078e8077e

Hanaasagi:
I tried the yazi 0.1.5 from the pacman source again, and the issue persists. I will recheck the configuration in the next couple of days. By the way, I'm using btrfs. Does this affect? `stat`, `lstat`, `statx` ?

sxyazi:
I reproduced it on my Linux; it seems to be a platform-specific issue. I'll look into it.

sxyazi:
I found the reason behind it. The `Cargo.toml` file renamed in that video has the same modification time as other files (maybe they were placed together by `git clone`). Renaming in Yazi performs two operations on the file list:

1. Deletes the old file from the file list;
2. Appends the renamed file as a new one at the "end" of the list.

Considering we sort by "last modified time" and they have the same time, this results in the renamed file always being the last one among all files with the "same modification time" - You can do an `ls -al` to check this.

Maybe I can combine these two operations into one "Replace" while preserving the original order. I'll try that out later.

sxyazi:
[...]

PR #398, opened 2023-11-24
fix: stepwise renaming causes uncertain sorting order

sxyazi:
Try to fix https://github.com/sxyazi/yazi/issues/390


Hanaasagi (approved):
Thanks for turning this around so quickly! LGTM



Uploading simplescreenrecorder-2023-11-25_00.34.13.mp4…

sxyazi:
Thank you for reporting this issue.

I found several related problems while debugging and fixed them together in this PR.

PR #400, opened 2023-11-25
feat: snap input: put cursor right before the extension, if any

dmerejkowsky:
Fix #345

### DIFF
--- a/yazi-core/src/input/snap.rs
+++ b/yazi-core/src/input/snap.rs
@@ -35,6 +35,13 @@ impl InputSnap {
 		self.cursor = self.cursor.min(self.value.chars().count().saturating_sub(self.mode.delta()));
 		self.offset =
 			self.offset.min(self.cursor.saturating_sub(Self::find_window(&self.rev(), 0, limit).end));
+		// If we have an extension, move cursor right before the last dot ...
+		if let Some((base, ext)) = self.value.rsplit_once('.') {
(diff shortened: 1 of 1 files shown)


yazi-core/src/input/snap.rs
```diff
@@ -35,6 +35,9 @@ impl InputSnap {
 		self.cursor = self.cursor.min(self.value.chars().count().saturating_sub(self.mode.delta()));
 		self.offset =
 			self.offset.min(self.cursor.saturating_sub(Self::find_window(&self.rev(), 0, limit).end));
+		if let Some((_base, ext)) = self.value.rsplit_once('.') {
```
sxyazi, line 38: I think it's pointless to move the cursor before the dot in filenames like `.bashrc`

dmerejkowsky: That makes sense - code updated :)

I also used `saturating_sub` instead of just `-`  because I got a crash otherwise (not sure why)

sxyazi:
Can we add it only for the [`rename` command](https://github.com/sxyazi/yazi/blob/d54983143aa1088e2aef52854b125f84cdb32036/yazi-core/src/manager/commands/rename.rs#L31)? Something like `rename --cursor ext`, this way is more flexible and won't affect other inputs, such as the shell box.

sxyazi:
Done by https://github.com/sxyazi/yazi/pull/513

PR #407, opened 2023-11-28
direnv watch flake.lock

hakan-demirli:
nix environment should update if lock file is changed. 
```watch_file``` is a direnv command which does that automatically: https://github.com/direnv/direnv/blob/9f22a3ea7e9fc835bb0569f14b35d756d02a4b7c/man/direnv-stdlib.1.md?plain=1#L368

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,1 +1,2 @@
+watch_file flake.lock
 use flake


XYenon:
It looks like `watch_file flake.lock` is already included by `use flake`.

https://github.com/direnv/direnv/blob/9f22a3ea7e9fc835bb0569f14b35d756d02a4b7c/stdlib.sh#L1276

hakan-demirli:
> It looks like `watch_file flake.lock` is already included by `use flake`.
I see. Then I have another problem since it is not updating env unless I reopen the terminal.

Issue #433, opened 2023-12-06
Past symlink as symlink


einhander:
### What system are you running Yazi on?

Linux X11

### What terminal are you running Yazi in?

kitty 0.31.0

### Yazi version

yazi 0.1.5 (1d1a512 2023-12-06)

### Did you try the latest main branch to see if the problem has already been fixed?

Tried, but the problem is still present

### Describe the bug

Copy-paste directory with symlinks inside it results in paste directory with files

### Expected Behavior

Copy and paste directory with symlinks

### To Reproduce

copy directory with symlinks inside it, paste it

### Configuration

no config

sxyazi:
Done by https://github.com/sxyazi/yazi/pull/436

PR #436, opened 2023-12-07
fix: add `--follow` option to `paste` command

eatradish:
Instead of copying the file pointed to by the symlink , create a soft link to the original path under the pasted path

See #433

### DIFF
--- a/yazi-core/src/manager/commands/paste.rs
+++ b/yazi-core/src/manager/commands/paste.rs
@@ -3,11 +3,14 @@ use yazi_shared::event::Exec;
 use crate::{manager::Manager, tasks::Tasks};
 
 pub struct Opt {
-	force: bool,
+	force:  bool,
+	follow: bool,
 }
 
 impl From<&Exec> for Opt {
-	fn from(e: &Exec) -> Self { Self { force: e.named.contains_key("force") } }
+	fn from(e: &Exec) -> Self {
+		Self { force: e.named.contains_key("force"), follow: e.named.contains_key("follow") }
+	}
 }
 
 impl Manager {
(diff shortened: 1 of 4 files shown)


sxyazi:
BTW another factor worth considering is that copying symlinks is more dangerous than copying actual files, as it can lead to the loss of user data, unless the user is truly aware of what they are doing and remains cautious.

sxyazi:
@einhander Okay thank you for showing me its real use case. I will try to retain both modes in the next few days.

eatradish:
> @eatradish I've just realized that `cp` behaves differently in macOS and Linux, and I did indeed reference macOS in my design since it's my development environment.
> 
> I did some tests on my Linux VM, where `/a/x` is a symbolic link. When I copied `/a` to `/b/c`:
> 
>     * If `x` points to an absolute path of a file, there's no issue; `/b/c/x --> /absolute/real-x`.
> 
>     * However, if it points to a relative path, it becomes a broken link; `/b/c/x --> ../real-x`, where `/b/real-x` doesn't exist.
> 
> 
> I'm curious if copying symlinks is truly necessary and if there are any practical use cases for it. I haven't actually used it in real scenarios, so let me know your thoughts.

I think the behavior of `cp` (GNU Coreutils) is to keep the behavior uniform with normal file copying in commland-line only (tty) environments, whereas under Windows when you copy a `shortcut' its target is also a shortcut, which is probably more common-sensical under a GUI

sxyazi:
@eatradish Does this mean Yazi should keep the same default behavior as GNU Coreutils? i.e. to only copy the symlink itself rather than the actual file it points to?

einhander:
> @eatradish Does this mean Yazi should keep the same default behavior as GNU Coreutils? i.e. to only copy the symlink itself rather than the actual file it points to?

@sxyazi  My opinion - Yes, Yazi should keep the same default behavior as GNU Coreutils.

sxyazi:
Okay, I'll finish the remaining work on this PR.

sxyazi:
Hey @eatradish, I've added a new `paste --follow` option and changed the default behavior to *not* follow. Please give it a try to see if it works for you!

PR #439, opened 2023-12-08
fix: do not use `LuaJIT` in riscv64 and loongarch64 architecture

eatradish:
... `LuaJIT` break some architecture (like: riscv64, loongarch64), make `mlua/luajit52` as default feature and add `mlua/lua52` feature to fix some architecture build.

### DIFF
--- a/yazi-plugin/Cargo.toml
+++ b/yazi-plugin/Cargo.toml
@@ -16,7 +16,12 @@ yazi-shared = { path = "../yazi-shared", version = "0.1.5" }
 # External dependencies
 ansi-to-tui   = "^3"
 anyhow        = "^1"
-mlua          = { version = "^0", features = [ "luajit52", "vendored", "serialize" ] }
 ratatui       = "^0"
 tracing       = { version = "^0", features = [ "max_level_debug", "release_max_level_warn" ] }
 unicode-width = "^0"
+
+[target.'cfg(any(target_arch = "riscv64", target_arch="loongarch64"))'.dependencies]
+mlua          = { version = "^0", features = [ "lua52", "vendored", "serialize" ] }
+
+[target.'cfg(not(any(target_arch = "riscv64", target_arch="loongarch64")))'.dependencies]
+mlua          = { version = "^0", features = [ "luajit52", "vendored", "serialize" ] }


eatradish:
There exists another way to solve this problem:

```
[target.'cfg(target_arch = "riscv64")'.dependencies]
mlua          = { version = "^0", features = [ "lua52", "vendored", "serialize" ] }

[target.'cfg(not(target_arch = "riscv64"))'.dependencies]
mlua          = { version = "^0", features = [ "luajit52", "vendored", "serialize" ] }
```

The disadvantage of this is that when there is a new architecture in the future, you will need to manually modify Cargo.toml to get it to compile, the advantage is that you don't need to use the `--features` flag, not sure which is the better option?

sxyazi:
> There exists another way to solve this problem

I prefer the simplicity of this approach, so that when other crates depend on `mlua` (like in this PR, https://github.com/sxyazi/yazi/pull/401/files#diff-18ec41b92ebb27bff4660e995870a426f248c04a498a4e132713dac8e01611a5), we don't need to add features for them as well.

eatradish:
> > There exists another way to solve this problem
> 
> I prefer the simplicity of this approach, so that when other crates depend on `mlua` (like in this PR, https://github.com/sxyazi/yazi/pull/401/files#diff-18ec41b92ebb27bff4660e995870a426f248c04a498a4e132713dac8e01611a5), we don't need to add features for them as well.

The main thing is that in some future unknown architectures that don't support LuaJIT, it won't be possible to turn off LuaJIT via `--no-default-features --features lua-without-jit` (you can only modify Cargo.toml manually), and I'll re-improve my PRs if you think that's ok

sxyazi:
Yea that's ok, it's just like a workaround.

I will try out `luau` and `luau-jit`, I want to know if they can be built on RISC-V?

eatradish:
> Yea that's ok, it's just like a workaround.
> 
> I will try out `luau` and `luau-jit`, I want to know if they can be built on RISC-V?

`luau-jit` on `riscv64gc` :

图片

eatradish:
> Yea that's ok, it's just like a workaround.
> 
> I will try out `luau` and `luau-jit`, I want to know if they can be built on RISC-V?

Done. look: https://github.com/sxyazi/yazi/pull/439/commits/84487d640bf6c7d8bba18331a5307fc4a166cd6a

sxyazi:
LGTM, thanks!

PR #454, opened 2023-12-24
feat: filter files in real-time

Rolv-Apneseth:


yazi-core/src/files/files.rs
```diff
@@ -322,6 +324,44 @@ impl Files {
...
 
+	// --- Filter
+	pub fn set_filter(&mut self, keyword: &str) -> bool {
+		if keyword == self.filter.as_deref().unwrap_or_default() {
+			return false;
+		}
+
+		let old_hidden: Vec<File>;
+		(self.hidden, old_hidden) =
+			self.hidden.iter().cloned().partition(|f| !self.show_hidden && f.is_hidden());
```
sxyazi, line 335: Is this should be `self.show_hidden` not `!self.show_hidden`?

Rolv-Apneseth: Perhaps this should be 2 separate lines, so it doesn't need to keep checking `self.show_hidden`, but I thought that logic was correct. Maybe I'm misunderstanding but shouldn't it go =>

1. `show_hidden = true` means all files are put in `old_hidden` so they can be re-added to `self.items`
2. `show_hidden = false` means any files that should be hidden (`f.is_hidden()`) remain in `self.hidden` but everything else goes to `old_hidden`

Rolv-Apneseth:
Those improvements were nice, didn't think of doing it like that

yazi-core/src/files/files.rs
sxyazi: Sorry, after some thinking, I still can't understand it.

Could you break it down into multiple lines? - I'll try to refactor it once I understand. 🤣

Rolv-Apneseth: Sure yeah

Rolv-Apneseth: Not sure why that formatting is weirdly unindented by 1 level but the change should be in that latest commit

Rolv-Apneseth: How are you doing formatting by the way? Just specific chunks, using that `rustfmt.toml`, rather than the whole file?

sxyazi: `rustfmt.toml` for Yazi used some nightly features, so it requires `rustfmt` nightly.

You can do a `rustfmt +nightly **/*.rs` to format all files (might need to install rustfmt nightly first via `rustup`).

Rolv-Apneseth: Ah ok thanks

Rolv-Apneseth:
If you need me to do anything else just let me know and I can get to it later / tomorrow

sxyazi:
Okay, currently switch between hiding and showing files causes the file list to be lost:

1. Display all files
2. Type a keyword to filtering
3. Hide all files
4. Cancel filtering - can't cancel, files lost

This requires https://github.com/sxyazi/yazi/pull/401 to be merged in order to proceed. I've completely refactored Yazi's file system in that PR. Maybe I'll be able to merge it tomorrow if there are no issues during the code review.

Rolv-Apneseth:
Ok, that's not great.

That's cool you're almost done with that PR though, seems like a big one.

sxyazi:
I think I've addressed all the edge cases. Would you like to give it a review?

Rolv-Apneseth:
Sure yeah I'll have a look

yazi-core/src/folder/files.rs
```diff
@@ -231,8 +219,15 @@ impl Files {
 			};
 		}
 
-		let (hidden, items) =
-			if self.show_hidden { (vec![], urls) } else { urls.into_iter().partition(|u| u.is_hidden()) };
+		let (hidden, items) = if let Some(filter) = &self.filter {
+			urls.into_iter().partition(|u| {
+				(u.is_hidden() && !self.show_hidden) || !u.file_name().is_some_and(|s| filter.matches(s))
```
Rolv-Apneseth, line 224: Major nitpick but if this is getting checked for every file is it a little better to check `self.show_hidden` first every time? This is the same further down too.

```suggestion
				(!self.show_hidden && u.is_hidden()) || !u.file_name().is_some_and(|s| filter.matches(s))
```

Rolv-Apneseth (commented):
All looks good to me. You've done a lot more than I originally did for the PR haha.

Seems good when using as well. Only have 2 minor things:

1. It may be a little confusing when filtering from search mode, as the search mode shows the full path but of course filter is only going by the file names. Probably a non-issue as ideally you would just search again.

2. The preview doesn't update after filtering so whatever file was hovered before filtering remains in the preview window.

yazi-core/src/folder/files.rs
sxyazi: Nice catch! I forgot to change this check for URLs - it needs to parse the string path, so it's a slow operation. I'll fix it!

But the `f.is_hidden()` for files is intentional - it's a cheap bitwise operation, and because most files aren't hidden, it can short-circuit the `!self.show_hidden`, checking only once.

Rolv-Apneseth: Ah ok cool

sxyazi:
> The preview doesn't update after filtering so whatever file was hovered before filtering remains in the preview window.

Should be fixed now!

Rolv-Apneseth:
> > The preview doesn't update after filtering so whatever file was hovered before filtering remains in the preview window.
> 
> Should be fixed now!

Yep that did it!

sxyazi:
Okay, let me merge it. Thank you very much for contributing this awesome feature! ❤️️

Issue #295, opened 2023-10-21
Syntax highlighting on macOS's Terminal.app is strange


mdesmedt:
Trying yazi for the first time, it builds and runs perfectly on MacOS Sonoma. However, inside the Terminal app with the default "Basic" theme syntax highlighting looks weird. There's white-on-teal highlighting which shows up in nearly every file. Not sure if it's a code bug or possible config improvement for Mac.

<img width="1256" alt="Screenshot 2023-10-21 at 7 41 25 AM" src="https://github.com/sxyazi/yazi/assets/73319561/7ff723e5-0639-4124-a5a0-d7af193bea77">

sxyazi:
Thanks for the report.

Currently, Yazi's code highlighting only supports terminals that have true color support. Unfortunately, macOS's Terminal.app is one of the few terminal emulators that still lacks true color support.

This is definitely a point for improvement, and it's already in the plans for v0.1.6. For now, you can use terminals like iTerm2, Kitty, etc., which not only offer more powerful functionality but also support image previews.

sxyazi:
Done by https://github.com/sxyazi/yazi/pull/460

PR #460, opened 2023-12-28
feat: support ANSI themes

smores56:
The current default theme is [base16-ocean.dark](https://github.com/sxyazi/yazi/blob/2d3512e965f4bd46d7fd8310424138b0c6cbe103/yazi-plugin/src/external/highlighter.rs#L29), which as you can see below doesn't match the terminal colors by default:

Screenshot from 2023-12-28 00-28-14

If I load an 8-bit theme like [this one](https://github.com/chtenb/neo-ansi/blob/main/terminal-neo-ansi.tmTheme) from this [neo-ansi repo](https://github.com/chtenb/neo-ansi), I currently see all black theming:

Screenshot from 2023-12-28 00-30-21

This is because ANSI is configured in `.tmTheme` files as #RR000000, a.k.a. the 8-bit color is in the red component and all other components, especially the Alpha, are `0x00`. Assuming colors are 8-bit if the alpha is zero is suggested by `syntect`'s author [in this comment](https://github.com/trishume/syntect/issues/309#issuecomment-695336160), and is implemented in bat already [here](https://github.com/sharkdp/bat/blob/b89dc15be1d86869bf73e5b9f96af7b930753a55/src/terminal.rs#L8). Given these precedents, I went ahead and implemented it, as seen below:

Screenshot from 2023-12-28 00-28-26

This implementation uses `crossterm` to write the escape sequences, which has a few benefits:
- We are already importing `crossterm`, so there are no new dependencies
- We have fewer magic constants, as `crossterm` provides them for us
- We make fewer string allocations, as we write directly to our output string this way instead of writing each line to a new string and then re-allocatiing

I think that ANSI themes are a better default for these kind of color-specific tools (since they always match the colors of the terminal), but there are no ANSI themes packed by default into `syntect`'s [ThemeSet::load_defaults](https://docs.rs/syntect/latest/syntect/highlighting/struct.ThemeSet.html#method.load_defaults) that we use. Maybe a future PR could change to [this ANSI theme](https://github.com/sharkdp/bat/blob/master/assets/themes/ansi.tmTheme) from [bat](https://github.com/sharkdp/bat/tree/master).


sxyazi:
Awesome, this is the feature I've been wanting to add all along. Thank you for implementing it! I'll review it ASAP.

sxyazi:
Looks good to me! Would you like to change the default to an ANSI theme in this PR as well?

I think it's quite reasonable as it not only makes Yazi more compatible with users' terminals but also potentially resolves https://github.com/sxyazi/yazi/issues/295.

smores56:
Yeah, I totally agree. Just didn't want to do it without the author's permission. I'm not exactly sure where the `ansi.tmTheme` should go, my current location is `yazi-config/preset/ansi.tmTheme`, but `yazi-plugin/preset/ansi.tmTheme` also makes sense. Let me know if you have a preference.

smores56:
I used the [neo-ansi theme](https://github.com/chtenb/neo-ansi/blob/main/terminal-neo-ansi.tmTheme) and not [bat's ANSI theme](https://github.com/sharkdp/bat/blob/master/assets/themes/ansi.tmTheme) since the `bat` theme was mostly black.

Screenshot from 2023-12-28 12-35-33

smores56:
I'm happy if you're happy, feel free to squash and merge whenever! Thanks for the awesome file manager!

sxyazi:
Thank you! Let me make some little changes.

sxyazi:
Hi, I've made some changes:

- Add support for bold, italic, and underline
- Remove unnecessary ANSI sequence conversions - highlighting results are converted to ANSI sequences and then re-parsed into TUI components. Now it's just one step - from highlighted results to TUI components. This can save a lot of memory overhead and ANSI sequence parsing costs.
- Switch to the default theme of `bat` and port it to another separate crate `yazi-prebuild`. I've copied' bat's code for parsing colors and the issue with black text disappears.

Would you like to give it a review?

PR #469, opened 2024-01-02
feat: add user and group-related plugin APIs

Urie96:


sxyazi:
Thank you for contributing these APIs. I will add the Lua code snippets to the tips of the Yazi doc so that users can copy them from there when they need.

PR #477, opened 2024-01-04
Add wiki link to README.md

RioNight:


README.md
```diff
@@ -22,8 +22,9 @@ https://github.com/sxyazi/yazi/assets/17523360/92ff23fa-0cd5-4f04-b387-894c12265
 
 ## Documentation
 
-- Usage: https://yazi-rs.github.io/docs/installation
+- Installation: https://yazi-rs.github.io/docs/installation
```
sxyazi, line 25: This is indeed a "usage" document, not an "installation" document. It just so happens that the first section is an installation guide, but I believe it doesn't conflict; installation is simply the initial step to start using it.

README.md
```diff
@@ -22,8 +22,9 @@ https://github.com/sxyazi/yazi/assets/17523360/92ff23fa-0cd5-4f04-b387-894c12265
...
 - Features: https://yazi-rs.github.io/features
+- Wiki: https://github.com/sxyazi/yazi/wiki
```
sxyazi, line 27: I tend to avoid putting the Wiki link here because there isn't much information on the Wiki itself:

- It only contains the Windows installation guide, which is already linked in the installation section of the usage documentation at https://yazi-rs.github.io/docs/installation/#windows
- As for the tmux setup, it's already linked in the image preview of the README, https://github.com/sxyazi/yazi?tab=readme-ov-file#image-preview

RioNight:
Understood.

Issue #112, opened 2023-09-05
Print the cwd to stdout


linsui:
Currently yazi provide `--cwd-file` to get the cwd from a file. I thought it would be more intuitive to print the path to stdout directly, similiar to fzf. `yazi --cwd-file=/dev/stdout` doesn't works so I guess the implementation is different.

sxyazi:
The impl of Yazi and fzf fundamentally differs. fzf sends its TUI to `stderr`; you can verify this by running `fzf 2> /dev/null`, which will not display any content.

Yazi uses `stdout` as its TUI endpoint, which I believe is a more suitable approach for TUI apps. Therefore it's no longer an option for using `stdout` as `cwd-file`.

For `yazi --cwd-file=/dev/stdout`, this is an incorrect case. You should specify a temporary file for `cwd-file`, just like it's mentioned in the README:

https://github.com/sxyazi/yazi/blob/eee68a99b4db9de0cba4b4e0b2a5b63bff618071/README.md?plain=1#L137-L146

linsui:
> The impl of Yazi and fzf fundamentally differs. fzf sends its TUI to stderr; you can verify this by running fzf 2> /dev/null, which will not display any content.

Interesting! And just for fun, this also works:

`cd $(yazi --cwd-file /dev/stderr 3>&1 1>&2 2>&3 | cat)`

sxyazi:
> `cd $(yazi --cwd-file /dev/stderr 3>&1 1>&2 2>&3 | cat)`

Oh that's really cool! Could you explain its principle?

linsui:
It's stolen from https://serverfault.com/questions/63705/how-to-pipe-stderr-without-piping-stdout. ;)

PR #484, opened 2024-01-06
Added flag to write opened file to stdout

aserowy:
Hi,

i added a flag to be able to write the selected file for opening to stdout. To enable multiple flags at the same time i refactored these actions into an enum for quit actions which gets handled on app.quit. Thus, all actions on quit are now treated at the same stage of the app lifecycle.

My motivation was to ensure my workflow stays intact when switching from lf to yazi and i thought maybe others find the option usefull as well.

Kind regards
Alexander

### DIFF
--- a/nix/shell.nix
+++ b/nix/shell.nix
@@ -5,7 +5,6 @@ pkgs.mkShell {
     rustToolchain
     rust-analyzer
 
-    nodejs_20
     nodePackages.cspell
 
     file
--- a/nix/shell.nix
+++ b/nix/shell.nix
@@ -5,6 +5,7 @@ pkgs.mkShell {
     rustToolchain
     rust-analyzer
 
+    nodejs_20
     nodePackages.cspell
 
     file
--- a/flake.lock
+++ b/flake.lock
@@ -5,11 +5,11 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1694529238,
-        "narHash": "sha256-zsNZZGTGnMOf9YpHKJqMSsa0dXbfmxeoJ7xHlrt+xmY=",
+        "lastModified": 1701680307,
+        "narHash": "sha256-kAuep2h5ajznlPMD9rnQyffWG8EM/C73lejGofXvdM8=",
         "owner": "numtide",
         "repo": "flake-utils",
-        "rev": "ff7b65b44d01cf9ba6a71320833626af21126384",
+        "rev": "4022d587cbbfd70fe950c1e2083a02621806a725",
         "type": "github"
       },
       "original": {
@@ -20,11 +20,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1701174899,
-        "narHash": "sha256-1W+FMe8mWsJKXoBc+QgKmEeRj33kTFnPq7XCjU+bfnA=",
+        "lastModified": 1704161960,
+        "narHash": "sha256-QGua89Pmq+FBAro8NriTuoO/wNaUtugt29/qqA8zeeM=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "010c7296f3b19a58b206fdf7d68d75a5b0a09e9e",
+        "rev": "63143ac2c9186be6d9da6035fa22620018c85932",
         "type": "github"
       },
       "original": {
@@ -51,11 +51,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1701224160,
-        "narHash": "sha256-qnMmxNMKmd6Soel0cfauyMJ+LzuZbvmiDQPSIuTbQ+M=",
+        "lastModified": 1704507282,
+        "narHash": "sha256-PDfS8fj40mm2QWpbd/aiocgwcI/WHzqLKERRJkoEvXU=",
         "owner": "oxalica",
         "repo": "rust-overlay",
-        "rev": "4a080e26d55eaedb95ab1bf8eeaeb84149c10f12",
+        "rev": "a127cccf7943beae944953963ba118d643299c3b",
         "type": "github"
       },
       "original": {
--- a/flake.lock
+++ b/flake.lock
@@ -5,11 +5,11 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1701680307,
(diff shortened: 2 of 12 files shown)


sxyazi:
Awesome! I'll review it ASAP.

sxyazi:
It seems meaningless to write the file list to `stdout`. Currently, Yazi writes the TUI to `stdout`, so using `yazi --chooser-stdout > path.txt` will prevent the TUI from displaying. 

AFAIK, fzf writes the TUI to `stderr` and outputs the selected file path to `stdout`.

aserowy:
it enables using stdout after the tui closes to enable further editor support without using temp files e.g.

aserowy:
> AFAIK, fzf writes the TUI to `stderr` and outputs the selected file path to `stdout`.

How big of a change would this be in yazi? Seems overall like a good idea.

sxyazi:
> it enables using stdout after the tui closes to enable further editor support without using temp files e.g.

Sorry I didn't understand this, how should I use `--chooser-stdout` for now?

sxyazi:
> > AFAIK, fzf writes the TUI to `stderr` and outputs the selected file path to `stdout`.
> 
> How big of a change would this be in yazi? Seems overall like a good idea. This would enable pipes and a more linux like behavior.

I quickly checked the ratatui and crossterm libraries used by Yazi, but couldn't find any relevant information. I'm not sure how to change it and how much effort it would involve.

Also I don't think writing TUI to `stderr` is appropriate for TUI apps because it's not like `fzf`, which purely outputs the user-selected path to `stdout`, and the UI is just its accessory.

sxyazi:
> In nvim i am using the following function to read directly from buffer without a tmpfile to open a selected file:
> 
> ```lua
> function M.close()
>     local lines = vim.api.nvim_buf_get_lines(0, 0, -1, false)
> 
>     if #lines == 0 then
>         return 1
>     end
> 
>     local path = lines[1]
>     if not file_exists(path) then
>         return 1
>     end
> 
>     vim.cmd("e " .. path)
>     vim.cmd("bd#")
>     vim.cmd("filetype detect")
> 
>     return 0
> end
> ```
> 
> Like i wrote initially, the change enables my current workflows. This is one example.

How do you select files in Yazi without the UI? Does nvim separate escape sequences from the selected file list in `stdout`?

aserowy:
FYI: https://ratatui.rs/faq/#should-i-use-stdout-or-stderr

aserowy:
> How do you select files in Yazi without the UI? Does nvim separate escape sequences from the selected file list in stdout?

Seems like it. Here the current output when im leaving the term open:

image

sxyazi:
I found this https://github.com/ratatui-org/website/issues/274; it seems that `stderr` might be slower than `stdout`. Yazi needs to transcode a large number of images to the terminal, and this could become quite noticeable.

aserowy:
Ok, how should we proceed? Adding niche functionallity, and a refactoring to handle all quit actions at the same stage?

sxyazi:
Due to the performance issues mentioned above, I've decided not to switch Yazi from `stdout` to `stderr`.

This would require more efforts in terms of performance optimization, and I still question whether it's suitable for a TUI application - afaik, it's an unconventional approach, a bit tricky, and I've only seen it in `fzf`.

Therefore, I would like to remove `chooser_stdout` and only keep the refactored results.

aserowy:
feel free. Thus, no switch for me from lf sadly.

sxyazi:
I'm sorry that Yazi couldn't match your workflow.

In fact, I never really understood what your X issue was, why only `stdout` would fit into your workflow, and why files wouldn't work, or what your workflow even looked like, and whether there's a workaround here, like https://github.com/sxyazi/yazi/issues/112#issuecomment-1706655289.

Anyway, feel free to close this PR if it's no longer valuable to you. Sorry again for any inconvenience it may have caused you.

aserowy:
Everything is cool! Please take the refactoring. I hope it helps! Was a cool finger practice :D

The link is interesting, maybe i can switch regardless (didnt even knew this is an option)

Valentin271:
For what it's worth, you might want to take a look at https://github.com/orhun/personal-blog/pull/24 to know more about stdout/stderr performance.

You can make stderr as fast as stdout by buffering it (code in the article) by wrapping stderr in a [BufWriter](https://doc.rust-lang.org/std/io/struct.BufWriter.html).

Valentin271:
> Funny. Thus, a buffered stderr should lead to better perf than the normal stdout in crossterm :D Something like this (untested!) would do the trick?

**In theory** yes, I won't explain what the article did better than me but the tldr is `stdout` is line buffered while `stderr` in unbuffered _by default_.

`LineWriter::new(stderr())` is the exact same as `stdout()`, `BufWriter` should **in theory** be a little bit faster for a TUI.

sxyazi:
The article is cool! I think it will clear up all my doubts about `stderr`. I'll read it and reconsider supporting `stderr`.

aserowy:
let me know if i should convert the removal of the flag commit. Otherwise the pr should only contain the refactoring.

sxyazi:
Let me make some changes and merge it first.

I'll try to add `stderr` support in the subsequent PR, which will happen in the next target v0.2.1, not this target (v0.2.0).

nix/shell.nix
```diff
@@ -5,6 +5,7 @@ pkgs.mkShell {
     rustToolchain
     rust-analyzer
 
+    nodejs_20
```
sxyazi, line 8: Is this necessary?

aserowy: ups, this shouldnt got pushed. Enabled gh copilot on my side.

sxyazi:
Thank you for the PR, let me merge it now.

sxyazi:
> I'll try to add `stderr` support in the subsequent PR, which will happen in the next target v0.2.1, not this target (v0.2.0).

Hi, long time no see! 

Yazi [switched to `stderr`](https://github.com/sxyazi/yazi/commit/411ba2fe85a896428288692d1851b83c2514fdbc) yesterday, so `stdout` can now be used for outputting the CWD. With the latest main branch, you can use

```bash
yazi --cwd-file /dev/stdout | cat
```

to get the CWD when exiting.


https://github.com/sxyazi/yazi/assets/17523360/7381010d-fcd4-4f4c-9b12-84359da61b69

PR #510, opened 2024-01-12
feat: add support for highlighting by file type

### LINKED ISSUE #262: Support color coding executables, symbolic links, and empty files
Currently yazi supports color coding by `mime` and `name`.

[As discussed](https://github.com/sxyazi/yazi/pull/161#issuecomment-1758092119) I would like to suggest to support color coding

- executable files,
- symbolic links, and
- empty files (or maybe different colors for different file sizes, including 0 byte)

Color codes for these cases should (probably) overwrite `mime` and `name` color settings.

_Personally_ I just use the following setting (light theme) and would consider color coding the cases mentioned above to be more important.

```
[filetype]
rules = [
	{ name = "*", fg = "#000000" },
	{ name = "*/", fg = "#0000ee" }
]
```

Thank you very much for considering this—and congratulation for your massive pr [#161](https://github.com/sxyazi/yazi/pull/161).

Akmadan23:
Closes #262

This adds support for highlighting executable files and symlinks.

### DIFF
--- a/yazi-config/src/theme/filetype.rs
+++ b/yazi-config/src/theme/filetype.rs
@@ -50,11 +50,7 @@ impl Filetype {
 					return false;
 				};
 
-				let mode_bin = format!("{:b}", metadata.permissions().mode());
-				return mode_bin.len() == 16 && match *mode_bin.as_bytes() {
-					[.., _, _, a, _, _, b, _, _, c] => a == b'1' || b == b'1' || c == b'1',
-					_ => false,
-				}
+				metadata.permissions().mode() & 0o100111 > 0o100000
 			},
 		});
 
--- a/yazi-config/src/theme/filetype.rs
+++ b/yazi-config/src/theme/filetype.rs
@@ -1,7 +1,7 @@
 use std::{path::Path, str::FromStr};
 
 use serde::{Deserialize, Deserializer};
-use yazi_shared::MIME_DIR;
+use yazi_shared::{MIME_DIR, fs::File};
 
 use super::{Color, Style, StyleShadow};
 use crate::Pattern;
@@ -39,23 +39,16 @@ pub struct Filetype {
 }
 
 impl Filetype {
-	pub fn matches(&self, path: &Path, mime: Option<&str>) -> bool {
+	pub fn matches(&self, file: &File, mime: Option<&str>) -> bool {
 		let is_dir = mime == Some(MIME_DIR);
 		let kind_check = self.kind.as_ref().is_some_and(|t| match t {
-			FileKind::Symlink => path.is_symlink(),
 			#[cfg(unix)]
-			FileKind::Executable => {
-				use std::os::unix::fs::PermissionsExt;
-				let Ok(metadata) = path.metadata() else {
-					return false;
-				};
-
-				metadata.permissions().mode() & 0o100111 > 0o100000
-			},
+			FileKind::Executable => !file.cha.is_dir() && file.cha.permissions & 0o111 != 0,
+			FileKind::Symlink => file.cha.is_link(),
 		});
 
 		kind_check
-			|| self.name.as_ref().is_some_and(|n| n.match_path(path, is_dir))
+			|| self.name.as_ref().is_some_and(|n| n.match_path(&file.url, is_dir))
 			|| self.mime.as_ref().zip(mime).map_or(false, |(m, s)| m.matches(s))
 	}
 }
--- a/yazi-config/src/theme/filetype.rs
+++ b/yazi-config/src/theme/filetype.rs
@@ -1,4 +1,4 @@
-use std::path::Path;
+use std::{path::Path, str::FromStr};
 
 use serde::{Deserialize, Deserializer};
 use yazi_shared::MIME_DIR;
@@ -6,20 +6,45 @@ use yazi_shared::MIME_DIR;
 use super::{Color, Style, StyleShadow};
 use crate::Pattern;
 
+#[derive(Deserialize)]
+#[serde(try_from = "String")]
+pub enum FileKind {
+	Executable,
+	Symlink,
+}
+
+impl FromStr for FileKind {
+	type Err = anyhow::Error;
+
+	fn from_str(s: &str) -> Result<Self, Self::Err> {
+		match s {
+			"executable" => Ok(Self::Executable),
+			"symlink" => Ok(Self::Symlink),
+			_ => Err(anyhow::anyhow!("invalid file kind"))
+		}
+	}
+}
+
+impl TryFrom<String> for FileKind {
+	type Error = anyhow::Error;
+
+	fn try_from(s: String) -> Result<Self, Self::Error> { Self::from_str(&s) }
+}
+
 pub struct Filetype {
 	pub name:  Option<Pattern>,
 	pub mime:  Option<Pattern>,
-	pub typ:   Option<String>,
+	pub kind:  Option<FileKind>,
 	pub style: Style,
 }
 
 impl Filetype {
 	pub fn matches(&self, path: &Path, mime: Option<&str>) -> bool {
 		let is_dir = mime == Some(MIME_DIR);
-		let special_types = self.typ.as_ref().is_some_and(|t| match t.as_str() {
-			"symlink" => path.is_symlink(),
+		let kind_check = self.kind.as_ref().is_some_and(|t| match t {
+			FileKind::Symlink => path.is_symlink(),
 			#[cfg(unix)]
-			"executable" => {
+			FileKind::Executable => {
 				use std::os::unix::fs::PermissionsExt;
 				let Ok(metadata) = path.metadata() else {
 					return false;
@@ -31,10 +56,9 @@ impl Filetype {
 					_ => false,
 				}
 			},
-			_ => false
 		});
 
-		special_types
+		kind_check
 			|| self.name.as_ref().is_some_and(|n| n.match_path(path, is_dir))
 			|| self.mime.as_ref().zip(mime).map_or(false, |(m, s)| m.matches(s))
 	}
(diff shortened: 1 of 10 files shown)


yazi-config/src/theme/filetype.rs
```diff
@@ -9,13 +9,33 @@ use crate::Pattern;
...
+			#[cfg(unix)]
+			"executable" => {
+				use std::os::unix::fs::PermissionsExt;
+				let Ok(metadata) = path.metadata() else {
+					return false;
+				};
+
+				let mode_bin = format!("{:b}", metadata.permissions().mode());
+				return mode_bin.len() == 16 && match *mode_bin.as_bytes() {
+					[.., _, _, a, _, _, b, _, _, c] => a == b'1' || b == b'1' || c == b'1',
```
Akmadan23, line 30: This is the most intuitive way I got it to work...

yazi-config/src/theme/filetype.rs
```diff
@@ -33,6 +53,8 @@ impl Filetype {
 		struct FiletypeRule {
 			name: Option<Pattern>,
 			mime: Option<Pattern>,
+			#[serde(rename = "type")]
```
Akmadan23, line 56: The `type` field name is a proposal, it makes sense to me but we could use something else if it's better

yazi-config/src/theme/filetype.rs
bvr-yr: i think you can use smth like
``` rust
metadata.permissions().mode() & 0o111 != 0
```
or `0o100` just for user-executable bit

sxyazi:
Thank you for the PR! I have some change requests:

- The `path.is_symlink()` is a time-consuming IO operation, and the current implementation does additional IO operations for each file on every match. We can modify the `path` parameter of `matches` to `file` and [pass the entire `File` struct](https://github.com/sxyazi/yazi/blob/c335b0272b48dd76a82fec5db4c47f6d89cd95fb/yazi-fm/src/lives/folder.rs#L87) containing preloaded metadata. You can get the necessary data through [`file.cha`](https://github.com/sxyazi/yazi/blob/c335b0272b48dd76a82fec5db4c47f6d89cd95fb/yazi-shared/src/fs/file.rs#L11).

- Please change `typ` to `kind` and create an enum to represent it, like `enum FiletypeKind { Symlink, Executable }`.

- It seems that `mode_bin.len()` is not always 16. Perhaps we can use constants from `libc` to better implement it, similar to what [`permissions()`](https://github.com/sxyazi/yazi/blob/c335b0272b48dd76a82fec5db4c47f6d89cd95fb/yazi-shared/src/fs/fns.rs#L97) does.

yazi-config/src/theme/filetype.rs
Akmadan23: Right, I didn't think about that... Thank you for the advice!

Akmadan23:
Should we move the `FileKind` enum in a separate file or is it fine to keep it there?

> It seems that `mode_bin.len()` is not always 16. Perhaps we can use constants from `libc` to better implement it, similar to what [`permissions()`](https://github.com/sxyazi/yazi/blob/c335b0272b48dd76a82fec5db4c47f6d89cd95fb/yazi-shared/src/fs/fns.rs#L97) does.

Checking if it's not a directory should be enough right?

sxyazi:
> Should we move the `FileKind` enum in a separate file or is it fine to keep it there?
> 
> > It seems that `mode_bin.len()` is not always 16. Perhaps we can use constants from `libc` to better implement it, similar to what [`permissions()`](https://github.com/sxyazi/yazi/blob/c335b0272b48dd76a82fec5db4c47f6d89cd95fb/yazi-shared/src/fs/fns.rs#L97) does.
> 
> Checking if it's not a directory should be enough right?

Nope, I mean you can use something like `libc::S_IXUSR` constants here instead of hardcoding it - `libc` will help us to define its value automatically, it may have different values on different platforms.

Also we don't need to check `is_dir()`, this will limit it to only available for files but not available for directories, if users need only apply it for files/directories, they can add additional constraints like `name = "*"` or `name = "*/"`.

sxyazi:
I can't wait to release v0.2.0, let me finish the next work of this PR! 😆

Akmadan23:
> Nope, I mean you can use something like [`libc::S_IXUSR`](https://www.gnu.org/software/libc/manual/html_node/Permission-Bits.html#index-S_005fIXUSR) constants here instead of hardcoding it - `libc` will help us to define its value automatically, it may have different values on different platforms.

Alright, no problem.

> Also we don't need to check `is_dir()`, this will limit it to only available for files but not available for directories, if users need only apply it for files/directories, they can add additional constraints like `name = "*"` or `name = "*/"`.

Well, if we don't check `is_dir()` all directories will be highlighted as executable files, and that's not the intended behavior...

sxyazi:
I added some other filetypes support, let's merge it then.

yazi-config/src/theme/filetype.rs
```diff
@@ -1,21 +1,36 @@
...
+			FiletypeIs::Link => file.cha.is_link(),
+			FiletypeIs::Orphan => file.cha.is_orphan(),
+			FiletypeIs::Sock => file.cha.is_socket(),
+			FiletypeIs::Sticky => file.cha.is_sticky(),
+		};
+		if !b {
+			return false;
+		}
+
+		self.name.as_ref().is_some_and(|n| n.match_path(&file.url, file.is_dir()))
```
Akmadan23, line 33: Doesn't seem to actually work... `b` is ignored if true

sxyazi:
Docs added here https://yazi-rs.github.io/docs/configuration/theme#filetype

yazi-config/src/theme/filetype.rs
sxyazi: Hmm have you set `name` or `mime` for it?

Akmadan23: Oh ok, now I get it. I was used to my old implementation which didn't require that

sxyazi: Yeah I want to use it as an extension to the existing `name` and `mime` - further restricting files that match either `name` or `mime`, rather than adding a new rule. This is more conducive to later expanding other attributes. 😉

og900aero:
It would also be worth solving that, the libraries.based are also colored, not just the file-based symlinks.

Brixy:
Thank you very much, guys, for this excellent addition and for closing [my original issue](/sxyazi/yazi/issues/262) with it.

Just noticed the following:

`is = "exec"` seems to have a higher priority than `is = "link"`, even for links to non-exec files.

With

```
{ name = "*/", is = "exec", fg = "green" }
```

(which does not make sense for folders) all linked folders a colored as executables.

If you set the color codes for exec files below linked files and folders, everything works correctly:

```
[filetype]

rules = [
	{ name = "*", is = "orphan", fg = "red" },
	{ name = "*/", is = "orphan", fg = "red" },
	{ name = "*", is = "link", fg = "cyan" },
	{ name = "*/", is = "link", fg = "cyan" },
	{ name = "*", is = "exec", fg = "green" } # <- has to be last rule
]
```

Not quite sure whether this is a tiny issue.

Thank a lot!

sxyazi:
Hey @Brixy, I'm not quite sure if I understand your meaning. Which files do you want to highlight, and which files do you not want to highlight at the same time?

Brixy:
> Hey @Brixy, I'm not quite sure if I understand your meaning. Which files do you want to highlight, and which files do you not want to highlight at the same time?

With the complete setting posted above *everything works as expected.* So, there is no real problem.

I just noticed that `exec` needs to be placed below `link`.

If you use

```
[filetype]

rules = [
	{ name = "*", is = "exec", fg = "green" },
	{ name = "*/", is = "exec", fg = "green" }, # makes no sense ;-)
	{ name = "*", is = "orphan", fg = "red" },
	{ name = "*/", is = "orphan", fg = "red" },
	{ name = "*", is = "link", fg = "cyan" },
	{ name = "*/", is = "link", fg = "cyan" }
]
```

all symbolic links are green, i.e. marked as executables -- although the links lead to non-executable files and folders.

Maybe yazi qualifies links as exec files in general?!

Please ask back if I did not make this clear. I am not sure whether this is an issue anyway.

sxyazi:
`{ name = "*/", is = "orphan", fg = "red" }` is meaningless; when a link is an orphan, it can only be a (pointing to an invalid location) file.

sxyazi:
I think I understand what you're saying, but it's not that links take precedence over executable files; in fact, they coexist.

Let's say:
- `a` points to `b`, where `b` is present and an executable file, then `a` has both "link" and "executable" properties
- `a` points to `c`, but `c` doesn't exist, then `a` only has the ~~"link"~~ "orphan" property (fixed in https://github.com/sxyazi/yazi/commit/85334457f30521c1f79687a514fc99a7cd5644cc).

I think this makes sense; when a file is a "link" pointing to an "executable", users can choose to present it either as a "link" or as an "executable" file.

Brixy:
Thanks a lot for your commit and your answers!

> I think this makes sense; when a file is a "link" pointing to an "executable", users can choose to present it either as a "link" or as an "executable" file.

Yes, totally 😀

---

Please allow one last question -- which is a shortform of my clumsy initial question.

If you have `a` pointing to `b` (but `b` does not exist) and this setting

```
[filetype]
rules = [
	{ name = "*", is = "exec", fg = "green" },
	{ name = "*", is = "orphan", fg = "red" },
]
```

I would expect the orphan `a` to be red, but it’s green. Is this intended? (It’s no problem to swap the rules, of course. Just asking in case it’s important for you.)

Brixy:
> I think we should follow this convention, separating the `x` property from it looks a bit strange. 🤔

👍

Oh sorry. I did’t `ls -la`, but just compared this case with some other terminal file managers.

Thanks for your patience!

PR #514, opened 2024-01-15
feat: add `YAZI_LEVEL` env variable for shells

Isti115:
It can be useful for several purposes, such as detecting nested instances.
For example, I like to have an indicator on my shell prompt line when using a shell that has been launched from my terminal file manager.

For reference, other terminal file managers implementing the same feature:
- https://github.com/ranger/ranger/commit/23b7f16961a7b679dec16f6ee91401a6b8f6ca82
- https://github.com/gokcehan/lf/commit/e82cbb820f5c50e551ca8ac54f3ec87683859dfd

Sorry for anything that's unusual about the way I implemented it, I haven't yet contributed much to Rust projects, so I'm not that familiar with the conventions around the language. I'm happy to receive any reviews and make adjustments as required!

---

ps. As a bonus, here's an example configuration for `starship`, it might be worth adding this to the "Tips" section of the docs, if you agree:
```
[env_var.YAZI]
variable = 'YAZI_LEVEL'
format = 'yz [ $env_value]($style) '
style = 'bold green'
```

### DIFF
--- a/yazi-plugin/src/external/shell.rs
+++ b/yazi-plugin/src/external/shell.rs
@@ -29,18 +29,12 @@ impl ShellOpt {
 }
 
 pub fn shell(opt: ShellOpt) -> Result<Child> {
+	let level = env::var("YAZI_LEVEL").ok().and_then(|s| s.parse::<u32>().ok()).unwrap_or(0);
+
 	#[cfg(unix)]
 	return Ok(unsafe {
 		Command::new("sh")
-			.env(
-				"YAZI_LEVEL",
-				(env::var("YAZI_LEVEL")
-					.ok()
-					.and_then(|lvl_str| lvl_str.parse::<u32>().ok())
-					.unwrap_or(0)
-					+ 1)
-				.to_string(),
-			)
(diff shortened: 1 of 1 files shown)


yazi-plugin/src/external/shell.rs
```diff
@@ -32,6 +32,15 @@ pub fn shell(opt: ShellOpt) -> Result<Child> {
 	#[cfg(unix)]
 	return Ok(unsafe {
 		Command::new("sh")
+			.env(
+				"YAZI_LEVEL",
+				(env::var("YAZI_LEVEL")
+					.ok()
+					.and_then(|lvl_str| lvl_str.parse::<i32>().ok())
```
sxyazi, line 39: Can it be a negative number? If not we should use `u32` here.

Isti115: Technically it can, if the user manually sets it to be so, but I cannot think of any use case where that would actually make sense, so it's probably ok to go with `u32`. I pushed a `fixup` commit.

sxyazi:
Thank you, let me merge it!

PR #523, opened 2024-01-16
feat: add option to enter directory on `open` command

Akmadan23:
This is another nice feature to have in my opinion, and it's something that almost any TUI file manager has.

### DIFF
--- a/yazi-core/src/manager/commands/open.rs
+++ b/yazi-core/src/manager/commands/open.rs
@@ -8,22 +8,31 @@ use yazi_shared::{emit, event::{EventQuit, Exec}, fs::{File, Url}, Layer, MIME_D
 use crate::{manager::Manager, select::Select, tasks::Tasks};
 
 pub struct Opt {
-	targets:     Option<Vec<(Url, Option<String>)>>,
-	interactive: bool,
+	targets:      Option<Vec<(Url, Option<String>)>>,
+	interactive:  bool,
+	or_enter_dir: bool,
 }
 
 impl From<&Exec> for Opt {
 	fn from(e: &Exec) -> Self {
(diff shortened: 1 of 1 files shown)


sxyazi:
That's intentional. See https://yazi-rs.github.io/docs/faq#why-cant-open-and-enter-be-a-single-command

You can use this [Smart enter](https://yazi-rs.github.io/docs/tips/#smart-enter-enter-for-directory-open-for-file) tip to impl it.

Akmadan23:
> hat's intentional. See https://yazi-rs.github.io/docs/faq#why-cant-open-and-enter-be-a-single-command
> 
> You can use this [Smart enter](https://yazi-rs.github.io/docs/tips/#smart-enter-enter-for-directory-open-for-file) tip to impl it.

Oh, I see, I did not know that. But why can't we just add this simple flag to achieve it?

sxyazi:
I want to keep it concise - one command should only do one thing.

Opening and entering are not essentially the same, and if we add it, we must also consider whether "entering should support opening" or "opening should support entering". In the end, we might need to add them for both commands.

Akmadan23:
> Opening and entering are not essentially the same, and if we add it, we must also consider whether "entering should support opening" (many people want l to support opening, but l is by default bound to the enter command) or "opening should support entering". In the end, we might need to add them for both commands.

Well, I actually tried to implement `enter --or-open` before, but since `enter` is a member of `Tab` it was really unpractical to access `open`, which is a `Manager` member, so I did it the other way around.
As for the `l` binding, we could add a note in `enter`'s documentation to use `open --or-enter-dir` for that use case...

Anyways, I understand your point here. Feel free to close PR if you don't see any value in this addition.

sxyazi:
Okay let me close it, sorry.

PR #531, opened 2024-01-17
feat: add support for snap packages

gapplef:
The `snapcraft.yaml` is quite straightforward. I'll explain a little bit, in case you may want to change it in the future.

On the top is various meta information, only things to be care are `grade` and `confinement`: 
 - grade can be `devel` or `stable`, defines the quality grade of the snap. 
 - confinement can be `strict` or `classic`, the latter means the snap have full system access just like traditionally applications.

The packing process is defined in `parts` section. 
- Building is take care by `rust` plugin, and carry out with `craftctl default`. You may replace it with any custom build commands. 
- `craftctl set version=...` is used to set the version info. 
- The rest is all for dependencies.
  current only `fd-find`, `ripgrep`, `zoxide`, `fzf` are boundle with yazi, these may need to be change with requirements.
- - - -
As for the release workflow,  I did not add new matrix, current build is take place on `x86_64-unknown-linux-gnu` with  `snapcore/action-build@v1`. The release file would be `yazi*.snap`. 
The snap package can be install with `sudo snap install <package.snap> --dangerous --classic`

Finally, to make it easier for users to get future updates, the snap can be publish to snap store. Snapcraft also provide its own [build service](https://snapcraft.io/build) to automatic build and release snaps.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,7 +51,7 @@ jobs:
           CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER: /usr/bin/aarch64-linux-gnu-gcc
         run: cargo build --release --locked --target ${{ matrix.target }}
 
-      - name: Build-snap
+      - name: Build snap
         if: matrix.target == 'x86_64-unknown-linux-gnu'
         uses: snapcore/action-build@v1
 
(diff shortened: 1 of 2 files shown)


sxyazi:
Thank you for your work and detailed description, let me make some minor changes and then merge it!

Issue #596, opened 2024-01-29
Yazi hang when enter a very large directory


trunglebka:
### What system are you running Yazi on?

Linux Wayland

### What terminal are you running Yazi in?

wezterm

### Yazi version

yazi 0.2.2 (68ffd82 2024-01-29)

### Did you try the latest main branch to see if the problem has already been fixed?

Tried, but the problem is still present

### Describe the bug

When entering a very large directory (contains 838684 .mp3 files) yazi hang and I can't do anything with it. 

### Expected Behavior

Working normally (for that large directory I really don't care much about detail of every file and just want to peek some of it since I can't read all of it)

### To Reproduce

using ya:
- Enter it directly by: `ya <very large dir>` - It will fully loaded (list files) after about 30s (not measured exactly). From this I can leave and re-enter the dir smoothly
- Enter it indirectly by: `ya <it's parent>` and then enter it by arrow keys, sometime it can list the files, sometime it freeze like the below image - about 13 minutes with 100% CPU
![image](https://github.com/sxyazi/yazi/assets/16150250/8b4ae70f-4174-4ea7-bb83-62046c426e30)

### Anything else?
The bigger problem is that it's very difficult to navigate to other files or directories when they are located beneath such a large directory.

trunglebka:
Well, it was able to enter that directory after 14 minutes (approximated by CPU time since I'm not using it and it is almost 0% CPU now)
![image](https://github.com/sxyazi/yazi/assets/16150250/a12456cb-f12f-4540-96a3-ff605d42ed38)

sxyazi:
I noticed that you are using the old version v0.2.1, please try if it exists in the latest version of v0.2.2, https://github.com/sxyazi/yazi/releases/tag/v0.2.2

sxyazi:
Hi, I made a PR to fix this, let me know if it works for you, https://github.com/sxyazi/yazi/pull/599.

trunglebka:
It's still, `yazi 0.2.2 (6735516 2024-01-29)`
![image](https://github.com/sxyazi/yazi/assets/16150250/b443012c-d5ce-4de7-bbc7-5bee1cbdbed8)

sxyazi:
Weird, it works fine on my side...

Could you record a GIF so I can see what was happening?

trunglebka:
Please see this video: https://1drv.ms/u/s!AjwpLQ8KafRujq4jOWTLlAJE-wSkdw?e=CRSFVM
[...]

PR #599, opened 2024-01-29
perf: read directory in bulk in the background at startup, make Yazi ~2.5 times faster than before

sxyazi:
Fixes https://github.com/sxyazi/yazi/issues/596

This PR optimizes the speed of reading large directories, leveraging the advantages of concurrency, it is ~2.5 times faster than before.

Here is a benchmark with a directory containing 500,000 files:

## `eza -l`, 19.03s

https://github.com/sxyazi/yazi/assets/17523360/21c2c70c-85dd-41f7-a5df-7add442d6997

## `ls -l`, 10.99s

https://github.com/sxyazi/yazi/assets/17523360/f05d7484-5dd5-4d2f-8b78-303d29423af0

## `yazi`, 4.79s

https://github.com/sxyazi/yazi/assets/17523360/2a5f65be-c801-40ee-a52f-1d8ecd5f98fa

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -230,9 +230,9 @@ checksum = "7f30e7476521f6f8af1a1c4c0b8cc94f0bee37d91763d0ca2665f299b6cd8aec"
 
 [[package]]
 name = "bytemuck"
-version = "1.14.0"
+version = "1.14.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "374d28ec25809ee0e23827c2ab573d729e293f281dfe393500e7ad618baa61c6"
+checksum = "ed2490600f404f2b94c167e31d3ed1d5f3c225a0f3b80230053b3e0b7b962bd9"
 
 [[package]]
 name = "byteorder"
@@ -269,14 +269,14 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "chrono"
-version = "0.4.31"
+version = "0.4.33"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f2c685bad3eb3d45a01354cedb7d5faa66194d1d58ba6e267a8de788f79db38"
+checksum = "9f13690e35a5e4ace198e7beea2895d29f3a9cc55015fcebe6336bd2010af9eb"
 dependencies = [
  "android-tzdata",
  "iana-time-zone",
  "num-traits",
- "windows-targets 0.48.5",
+ "windows-targets 0.52.0",
 ]
 
 [[package]]
@@ -303,9 +303,9 @@ dependencies = [
 
 [[package]]
 name = "clap_complete"
-version = "4.4.8"
+version = "4.4.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "eaf7dcb7c21d8ca1a2482ee0f1d341f437c9a7af6ca6da359dc5e1b164e98215"
+checksum = "df631ae429f6613fcd3a7c1adbdb65f637271e561b03680adaa6573015dfb106"
 dependencies = [
  "clap",
 ]
@@ -918,9 +918,9 @@ checksum = "029d73f573d8e8d63e6d5020011d3255b28c3ba85d6cf870a07184ed23de9284"
 
 [[package]]
 name = "indexmap"
-version = "2.1.0"
+version = "2.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d530e1a18b1cb4c484e6e34556a0d948706958449fca0cab753d649f2bce3d1f"
+checksum = "433de089bd45971eecf4668ee0ee8f4cec17db4f8bd8f7bc3197a6ce37aa7d9b"
 dependencies = [
  "equivalent",
  "hashbrown",
@@ -950,15 +950,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e05c02b5e89bff3b946cedeca278abc628fe811e604f027c45a8aa3cf793d0eb"
 dependencies = [
  "libc",
-]
-
-[[package]]
-name = "itertools"
-version = "0.11.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b1c173a5686ce8bfa551b3563d0c2170bf24ca44da99c7ca4bfdab5418c3fe57"
-dependencies = [
- "either",
 ]
 
 [[package]]
@@ -1085,9 +1076,9 @@ checksum = "b5e6163cb8c49088c2c36f57875e58ccd8c87c7427f7fbd50ea6710b2f3f2e8f"
 
 [[package]]
 name = "lru"
-version = "0.12.1"
+version = "0.12.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2994eeba8ed550fd9b47a0b38f0242bc3344e496483c6180b69139cc2fa5d1d7"
+checksum = "db2c024b41519440580066ba82aab04092b333e09066a5eb86c7c4890df31f22"
 dependencies = [
  "hashbrown",
 ]
@@ -1103,9 +1094,9 @@ dependencies = [
 
 [[package]]
 name = "luajit-src"
-version = "210.5.4+c525bcb"
+version = "210.5.5+f2336c4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2a10ab4ed12d22cb50ef43ece4f6c5ca594b2d2480019e87facfd422225a9908"
+checksum = "d8bcba9790f4e3b1c1467d75cdd011a63bbe6bc75da95af5d2cb4e3631f939c4"
 dependencies = [
  "cc",
  "which",
@@ -1166,9 +1157,9 @@ dependencies = [
 
 [[package]]
 name = "mlua"
-version = "0.9.4"
+version = "0.9.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "069264935e816c85884b99e88c8b408d6d92e40ae8760f726c983526a53546b5"
+checksum = "1d3561f79659ff3afad7b25e2bf2ec21507fe601ebecb7f81088669ec4bfd51e"
 dependencies = [
  "bstr",
  "erased-serde",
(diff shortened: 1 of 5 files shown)


uncenter:
I wonder if a loading indicator for huge directories would be nice to help the user understand that something is happening in the background?

sxyazi:
> I wonder if a loading indicator for huge directories would be nice to help the user understand that something is happening in the background?

Yeah definitely. It's been on my todo list for a long time ;)

mgumz:
@sxyazi: is yazi taking a lazy-rendering approach to display the names of the files in a directory? (so, 100 lines of visible items, only draw 100 items)?

sxyazi:
@mgumz 

Yeah, Yazi always lazily renders, which is the baseline for optimization.

However, this is unrelated to this PR - note the number of files in the bottom right corner of the Yazi video, which is the time required for **all files** to load, making it meaningful to compare with `ls` and `eza`.

mgumz:
@sxyazi Yah, but it is kind of relevant: exa/eza/ls etc produce output … just to get it dumped to /dev/null. 

$> for i in `seq 1 100000`; do touch $i.txt; done

$> time ( find files | wc -l; )  
0.01s user 0.14s system 43% cpu 0.351 total 

```golang
package main

import (
        "fmt"
        "os"
)

func main() {

        dn := os.Args[1]
        fi, _ := os.ReadDir(dn)
        total := int64(0)
        n := 0
        for i := range fi {
                f, _ := fi[i].Info()
                total += f.Size()
                n = i
        }

        fmt.Println(dn, n, total)
}                    
```

$> time go run main.go files
./files 100000 0
go run main.go ./files 
0.26s user 0.50s system 64% cpu 1.166 total

(yes, just for the lulz i added the compile time here as well)

change main.go to print out each name + size():

$> time go run main.go ./files > /dev/null
go run main.go ./files > /dev/null  
0.32s user 0.57s system 59% cpu 1.494 total

$>  time ( ls -l files  > /dev/null ) 
( ls -l files > /dev/null; )  
0.52s user 0.58s system 63% cpu 1.754 total


did you execute `eza`, `ls` and `yazi` in that order? did you drop OS caches in between the runs? did you run multiple runs (see https://github.com/sharkdp/hyperfine)? 

point is: I believe your 2.5x speed improvement within yazi, but the comparisons with those other tools … feels like apples and oranges and is kind of unneeded to show your improvement.

mgumz:
My main point is that you opened a discussion line which is not needed. yazi is now 2.5 times faster to open folders, done.

> This is based on the premise that I don't think /dev/null

/dev/null is not the time consuming thing - producing the 500k lines of text is (piping each file-entry through a strftime() equivalent, creating permission lines etc). 

And yazi by default is looking at all the 500k files … and showing the output only for the … 40 / 100 lines. Which in itself is a strange comparison.

Your new method makes it much more … comparable: now one can see all the dir-entries are really "rendered" (to make then /dev/null very happy).

https://github.com/eza-community/eza/blob/main/src/output/time.rs#L82 <- and this is what eza/exa calls on each entry. 
My snippet from above takes roughly 1s (including the time it takes to compile it) - and there I also avoid piping everything through some sort of strftime() like call. Your lazy loaded version takes for 500k files also ~ 1s per 100k files. 
When I leave out the compile overhead, I am now looking at 0.3s (max) to iterate over 100k entries. And I do not do anything fancy like bulk-loading-in-async-fashion (dont get me wrong: its the right thing to do for the kind of app yazi is). As soon as I print out the mod-time, the whole thing becomes twice as "slow".

```
; time ( ls -1 files > /dev/null )
( ls -1 files > /dev/null; )  0.33s user 0.09s system 93% cpu 0.439 total
; time ( ls -l files > /dev/null )
( ls -l files > /dev/null; )  0.52s user 0.59s system 67% cpu 1.643 total
```

```
; time gfind files -printf '%TY-%Tm-%TdT%TH:%TM:%TS %p\n' > /dev/null
gfind files -printf '%TY-%Tm-%TdT%TH:%TM:%TS %p\n' > /dev/null  0.43s user 0.20s system 85% cpu 0.741 total
; time gfind files > /dev/null 
gfind files > /dev/null  0.04s user 0.06s system 63% cpu 0.151 total
```

I just made `ls` faster by a factor of 4. I made `gfind` faster by factor of 4. =)


Since I was curious about what happened with the same 500k files you hammered your /tmp filesystem on MacOS:

```
; time ( find files > /dev/null )
( find files > /dev/null; )  0.04s user 6.28s system 15% cpu 40.937 total
; time ( ls -1 files > /dev/null )
( ls -1 files > /dev/null; )  2.14s user 6.07s system 19% cpu 41.191 total
; time ( ls -l files > /dev/null )
( ls -l files > /dev/null; )  3.13s user 20.09s system 21% cpu 1:48.07 total
```

yazi (on an M1) for the 500k files, hand stopped: 27s with display of mtime, 24s with default settings.

So, anyway. Kudos for 2.5 x more speed on load time on such crazy high numer of files.

sxyazi:
> /dev/null is not the time consuming thing - producing the 500k lines of text is (piping each file-entry through a strftime() equivalent, creating permission lines etc).

I'm confused, are you saying that ls is slow due to formatting time, rather than the IO itself? I find it hard to believe that Yazi would be slower than ls with single-threaded file reading, even with multi-threading to speed up file reading.

> I just made ls faster by a factor of 4. I made gfind faster by factor of 4. =)
> yazi (on an M1) for the 500k files, hand stopped: 27s with display of mtime, 24s with default settings. Now gnu-ls is 8 times faster than yazi, omg omg omg :)

Please provide benchmark code that can reproduce this result.

mgumz:
> Please provide benchmark code that can reproduce this result.

I did: `ls -1` (not producing time-formatted output) is faster than `ls -l`. `gls` (`ls` from `coreutils`) is faster than standard MacOS `ls`.

> I'm confused, are you saying that ls is slow due to formatting time, rather than the IO itself?

The output generation plays a role. See my "benchmarks" of `ls`. See my time measuring of yazi on 500k files, one time with mtime-display, one time not: (24s vs 27s for the same thing, 10% "improvement").
See http://blog.kazuhooku.com/2014/11/the-internals-h2o-or-how-to-write-fast.html, slide 41+ about stringification.

Obviously yazi benefits from being able to utilize more threads (and thus puts more pressure on the system in that moment, ) =). yazi might speed up the loading time even more by just collecting the filenames, and getting fs::metadata() only for those 100 items actually rendered (yes, that depends on the sort mode).

500k files, only counting the number of files in a folder, single threaded.

```
#include <dirent.h>
#include <stdio.h>
#include <stdlib.h>
#include <sys/types.h>

int main(int argc, char *argv[]) {

  char *dn = argv[1];
  DIR *d = opendir(dn);
  struct dirent *e;
  int n = 0;

  for (;;) {
    e = readdir(d);
    if (e == NULL) {
      break;
    }
    n++;
  }

  printf("%s: %d entries\n", dn, n);

  return 0;
}
```

```
; time ./a.out files
files: 500001 entries
./a.out files  0.01s user 0.31s system 20% cpu 1.509 total
```

I wonder about a bunch of things now:

* Why is `gls` so much faster than `ls` for comparable the same thing.
* What is the reason that scanning 500k files (41s) on apfs is much worse than 100k files (1.5s).
* Why did MacOS seems to have a limit on HFS+ on the amount of files in a singe folder around 17k.

None of those are relevant for this PR here.

My point is: by introducing the other tools into the presentation of the (great) achievement of 2.5x faster loading times of yazi, you opened up discussions like mine here. So, next time just leave out those comparisons and noone can start arguing.

sxyazi:
> I did: ls -1 (not producing time-formatted output) is faster than ls -l. gls (ls from coreutils) is faster than standard MacOS ls.

I guess you must have misunderstood something. As far as I know, `ls -1` only loads the file names and simply prints them, without loading the "metadata" information of the files, and only `ls -l` will load metadata, which is why it was chosen as the comparison object:


```c
// https://github.com/coreutils/coreutils/blob/a357cfb021cc0ecaff1de81014271977ce1502e7/src/ls.c#L1729-L1732

  format_needs_stat = sort_type == sort_time || sort_type == sort_size
    || format == long_format  // <---------------- Note this, i.e. `ls -l`
    || print_scontext
    || print_block_size;
```

And after a bunch of condition checks, the file status is actually loaded through `do_stat` or `do_lstat`:

```c
https://github.com/coreutils/coreutils/blob/a357cfb021cc0ecaff1de81014271977ce1502e7/src/ls.c#L3404-L3442

  if (command_line_arg
      || print_hyperlink
      || format_needs_stat   // <---------------- Note this, i.e. `ls -l`
      /* When coloring a directory (we may know the type from
         direct.d_type), we have to stat it in order to indicate
         sticky and/or other-writable attributes.  */
      || (type == directory && print_with_color
          && (is_colored (C_OTHER_WRITABLE)
              || is_colored (C_STICKY)
              || is_colored (C_STICKY_OTHER_WRITABLE)))
      /* When dereferencing symlinks, the inode and type must come from
         stat, but readdir provides the inode and type of lstat.  */
      || ((print_inode || format_needs_type)
          && (type == symbolic_link || type == unknown)
          && (dereference == DEREF_ALWAYS
              || color_symlink_as_referent || check_symlink_mode))
      /* Command line dereferences are already taken care of by the above
         assertion that the inode number is not yet known.  */
      || (print_inode && inode == NOT_AN_INODE_NUMBER)
      || (format_needs_type
          && (type == unknown || command_line_arg
              /* --indicator-style=classify (aka -F)
                 requires that we stat each regular file
                 to see if it's executable.  */
              || (type == normal && (indicator_style == classify
                                     /* This is so that --color ends up
                                        highlighting files with these mode
                                        bits set even when options like -F are
                                        not specified.  Note we do a redundant
                                        stat in the very unlikely case where
                                        C_CAP is set but not the others. */
                                     || (print_with_color
                                         && (is_colored (C_EXEC)
                                             || is_colored (C_SETUID)
                                             || is_colored (C_SETGID)
                                             || is_colored (C_CAP)))
                                     )))))

    {
```

On the other hand, Yazi loads ***all*** the "metadata" information of 500,000 files at once, and the real performance bottleneck here is the "metadata" itself (not some mysterious string formatting):

https://github.com/sxyazi/yazi/blob/e51e8ad789914b2ab4a9485da7aa7fbc7b3bb450/yazi-core/src/folder/files.rs#L79-L83


So, this is comparing ***two completely different things***, and I'm not sure why you are doing it.

> My point is: by introducing the other tools into the presentation of the (great) achievement of 2.5x faster loading times of yazi, you opened up discussions like mine here. So, next time just leave out those comparisons and noone can start arguing.

You [initiated this discussion](https://github.com/sxyazi/yazi/pull/599#issuecomment-1958795998), and all optimizations in this PR are related to I/O (if you have looked at the code, if not, you really should take a look), but you asked if it was related to lazy rendering, which confused me. 

When [I explained that lazy rendering is not related to this PR](https://github.com/sxyazi/yazi/pull/599#issuecomment-1959179980), you started [some confusing responses about it being related to output](https://github.com/sxyazi/yazi/pull/599#issuecomment-1960108695), and you raised a series of questions:

> did you execute eza, ls and yazi in that order? did you drop OS caches in between the runs? did you run multiple runs (see https://github.com/sharkdp/hyperfine)?

Which I really don't understand.

mgumz:
@sxyazi … apparently I am unable to bring my point across.

I read the PR description "This PR optimizes the speed of reading large directories, leveraging the advantages of concurrency, it is ~2.5 times faster than before." 

My point was: there the description should have ended.

(Because with having that benchmarky-comparison in place, you allowed me to initiate the discussion).

Nevermind.

PR #628, opened 2024-02-04
feat: support char device in filetype

HE7086:
Was messing around with themes and realized that character special device is not supported in filetype despite already being implemented. This adds it back.


sxyazi:
Nice catch, thank you!

Issue #551, opened 2024-01-21
Exit shell popups and `/` search with `Backspace` like in vim


RioNight:
### Is your feature request related to a problem? Please describe

It would be nice to exit the shell and search popups and the `/` search by `Backspace`-ing when there is no input in the input field.

### Will you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

When there is no input in the input field, `Backspace` should close the popup / prompt.

### Additional context

This is similar to the behavior in vim and would make it easier to use. If you don't have `Escape` mapped to `Caps Lock`, it is a bit of stretch to press `Escape`.

sxyazi:
Does this mean that all characters must be deleted, and then press `Enter` again to exit?

RioNight:
I was thinking of the vim behavior.
When you press `/` in vim to enter search mode:
1. pressing `Backspace` right after will exit out of search mode back into normal mode, and
2. entering a few characters, say the characters `hello`, then `Backspace`-ing them all to get an empty prompt, and then pressing `Backspace` once more on the empty prompt, will exit out of search mode into normal mode.

Pressing `Enter` on an empty prompt in yazi actually makes `rg` and `fd` search for `nil`. You might want to take care of that separately.
![image](https://github.com/sxyazi/yazi/assets/58403689/8e759ff6-e014-49f9-89f2-c2cc903437ac)

PR #630, opened 2024-02-05
feat: if prompt is empty, close it on `backspace` command

Rolv-Apneseth:
In relation to #551 

Also fixed issue of search` being run with `nil` mentioned in [this comment](https://github.com/sxyazi/yazi/issues/551#issuecomment-1902705562).


sxyazi:
Hi, thanks for your PR!

I've made some simplifications - closing input via `self.close()`.

Allowing empty keyword searches is expected behavior, as some users use it to achieve similar functionality to a "flat view". However, displaying `nil` is indeed a bug, which I've addressed through another method.

Rolv-Apneseth:
Ah I see, thanks for that

PR #646, opened 2024-02-08
feat: add counter component to the header for displaying currently selected/yanked items

Rolv-Apneseth:


sxyazi:
Thank you for the PR!

I just want to add an API to get the selected file, but I realized an issue - currently, the selection state is specific to the directory, meaning it can only get the selection state of the current directory.

But it would make more sense to get all selected files for display. We will add support for selecting across directories in the future.

So I need to do some refactoring to modify it to a tab-specific global selection state. I'll continue reviewing this PR after the refactoring is done :)

Rolv-Apneseth:
Ah I didn't catch that actually. Best of luck then, we can come back to this later

og900aero:
Would it be possible to write not only the number of selected/yanked items, but also the size of all selected items?

Rolv-Apneseth:
> Would it be possible to write not only the number of selected/yanked items, but also the size of all selected items?

That may be better as a plugin or at least a separate feature, @sxyazi what do you think?

sxyazi:
Yep it should be work as a UI plugin and not included in this PR.

Rolv-Apneseth:
Should be ready for you to review when you have the time @sxyazi , pretty small PR

yazi-config/preset/theme.toml
```diff
@@ -28,6 +28,11 @@ tab_active   = { fg = "black", bg = "lightblue" }
 tab_inactive = { fg = "white", bg = "darkgray" }
 tab_width    = 1
 
+# Selected counter
+count_selected = { fg = "white", bg = "green" }
+count_copied   = { fg = "white", bg = "yellow" }
+count_cut      = { fg = "white", bg = "red" }
```
sxyazi, line 34: Its better to keep this color as same as the marker_*

yazi-plugin/preset/components/header.lua
```diff
@@ -14,6 +14,34 @@ function Header:cwd()
 	return span:style(THEME.manager.cwd)
 end
 
+function Header:selected_count()
```
sxyazi, line 17:
```suggestion
function Header:counter()
```

yazi-plugin/preset/components/header.lua
```diff
@@ -14,6 +14,34 @@ function Header:cwd()
...
+	local selected = #cx.active.selected
+	local yanked = #cx.yanked
+	local is_cut = cx.yanked.is_cut
+
+	local count
+	local style
+	if is_cut then
```
sxyazi, line 24:
```suggestion
	if cx.yanked.is_cut then
```

yazi-plugin/preset/components/header.lua
```diff
@@ -42,8 +70,8 @@ end
 function Header:render(area)
 	local chunks = self:layout(area)
 
-	local left = ui.Line { self:cwd() }
-	local right = ui.Line { self:tabs() }
+	local left = ui.Line({ self:cwd() })
+	local right = ui.Line({ self:selected_count(), self:tabs() })
```
sxyazi, line 74:
```suggestion
	local left = ui.Line { self:cwd() }
	local right = ui.Line { self:counter(), self:tabs() }
```

yazi-plugin/preset/components/header.lua
Rolv-Apneseth: Done, thanks

yazi-plugin/preset/components/header.lua
```diff
@@ -14,6 +14,33 @@ function Header:cwd()
...
+	local count
+	local style
+	if cx.yanked.is_cut then
+		count = yanked
+		style = THEME.manager.count_cut
+	elseif yanked > 0 then
+		count = yanked
+		style = THEME.manager.count_copied
+	else
+		count = selected
```
sxyazi, line 30:
```suggestion
		count = cx.active.selected
```

yazi-config/preset/theme.toml
Rolv-Apneseth: `lightred` works fine but I think light green and yellow look a bit illegible with the white text, at least with my colour scheme. Do you think it looks alright?

Rolv-Apneseth: Depends entirely on what those values are I suppose but you would assume the lighter colours would lean more towards white than the darker ones

yazi-plugin/preset/components/header.lua
Rolv-Apneseth: If you want the least amount of variable declarations, should I also do `local count = #cx.yanked` and then update it to `selected` within the `if` block?

yazi-config/preset/theme.toml
sxyazi: Would it be better if it were `yellow` instead of `lightyellow`? Maybe we can use a dark foreground color, like `{ fg="black", bg="yellow" }`?

yazi-plugin/preset/components/header.lua
sxyazi: Yeah, because accessing these variables also has a cost (albeit a small one), but it would be better if possible to put them in an if-else block.

Let me try to change it

yazi-config/preset/theme.toml
Rolv-Apneseth: Actually yeah that might make sense as the default since the default `fg` for the active tab is `black` too, I just have that different for my config. So use light colours like the marker styles and then a black `fg`?

yazi-plugin/preset/components/header.lua
Rolv-Apneseth: Ah so just moving `selected` and `yanked` into the `if` block should be what you want then? Just need to repeat the `count = #cx.yanked` line but that's fine

sxyazi: yep i pushed a new commit should resolved this

Rolv-Apneseth:
Yep those changes seem good. I do also like hiding the tab component if there's only one.

sxyazi:
Alright, I think there are no more problems here, let us merge it!

sxyazi:
@Brixy Thanks for the reporting, should also be fixed in https://github.com/sxyazi/yazi/pull/690

PR #654, opened 2024-02-10
feat: add a way to use zoxide as an input field

diniamo:
I'd appreciate it if we discussed this, before you make any changes.

Missing:
- [ ] New default keybind for `jump fzf`
- [ ] Fix the hang on line 55
- [ ] Replace `Err(anyhow!(""))` with the correct way to create an empty error

Breaking changes:
- `jump zoxide` now uses an input field, and the old functionality is moved to `jump zoxide_interactive`
- Default `jump fzf` keybind is changed

### DIFF
--- a/yazi-core/src/tab/commands/jump.rs
+++ b/yazi-core/src/tab/commands/jump.rs
@@ -1,7 +1,10 @@
+use anyhow::anyhow;
+use yazi_config::popup::InputCfg;
 use yazi_plugin::external::{self, FzfOpt, ZoxideOpt};
 use yazi_scheduler::{Scheduler, BLOCKER};
 use yazi_shared::{event::Cmd, fs::ends_with_slash, Defer};
 
+use crate::input::Input;
 use crate::tab::Tab;
 
 pub struct Opt {
@@ -13,6 +16,7 @@ pub enum OptType {
 	None,
 	Fzf,
 	Zoxide,
+	ZoxideInteractive,
 }
 
 impl From<Cmd> for Opt {
@@ -21,6 +25,7 @@ impl From<Cmd> for Opt {
 			type_: match c.args.first().map(|s| s.as_str()) {
 				Some("fzf") => OptType::Fzf,
 				Some("zoxide") => OptType::Zoxide,
+				Some("zoxide_interactive") => OptType::ZoxideInteractive,
 				_ => OptType::None,
 			},
 		}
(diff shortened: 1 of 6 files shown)


yazi-core/src/tab/commands/jump.rs
```diff
@@ -33,17 +38,28 @@ impl Tab {
...
 			Scheduler::app_stop().await;
 
-			let result = if opt.type_ == OptType::Fzf {
-				external::fzf(FzfOpt { cwd }).await
-			} else {
-				external::zoxide(ZoxideOpt { cwd }).await
+			let result = match opt.type_ {
+				OptType::Fzf => external::fzf(FzfOpt { cwd }).await,
+				OptType::ZoxideInteractive => external::zoxide(ZoxideOpt { cwd, query: None }).await,
+				OptType::Zoxide => {
```
Rolv-Apneseth, line 51: I don't think the `anyhow` errors are necessary.

@sxyazi might have a cleaner way but I can get this feature to work by copying how the input is used elsewhere:

```rust
				OptType::Zoxide => {
					tokio::spawn(async move {
						let rx = Input::_show(InputCfg::zoxide());
						let rx = Debounce::new(UnboundedReceiverStream::new(rx), Duration::from_millis(50));
						pin!(rx);

						while let Some(Ok(s)) | Some(Err(InputError::Typed(s))) = rx.next().await {
							if let Ok(url) =
								external::zoxide(ZoxideOpt { cwd: cwd.clone(), query: Some(s) }).await
							{
								Tab::_cd(&url)
							}
						}
					});
					return;
				}
				_ => return,
			};
```

Also minor but `cargo +nightly fmt` or `rustfmt +nightly **/*.rs` to format the Rust files.

sxyazi:
Thanks for your PR!

I plan to reimplement zoxide as a plugin to provide more flexibility, which will eventually deprecate the `jump` command, conflicting a bit with the content of this PR. I just haven't had the time to do it yet, but it should be soon ;)

sxyazi:
Hi, I've now reimplemented zoxide as a plugin in https://github.com/sxyazi/yazi/issues/865, which will allow users to copy it as a template and easily modify the commands inside.

That is, you can copy https://github.com/sxyazi/yazi/blob/main/yazi-plugin/preset/plugins/zoxide.lua as your `~/.config/yazi/plugins/zoxide.yazi/init.lua`, and then call [`ya.input()`](https://yazi-rs.github.io/docs/plugins/utils#ya.input) to receive the user's input, and passes it to the `zoxide` command - Yazi will load your own `zoxide.yazi` plugin automatically

PR #659, opened 2024-02-11
feat: Vim-like notification

sxyazi:
https://github.com/sxyazi/yazi/assets/17523360/397eec4b-b4e8-413f-9b62-9f2b0765f30d

### DIFF
--- a/yazi-fm/src/notify/notify.rs
+++ b/yazi-fm/src/notify/notify.rs
@@ -1,0 +1,54 @@
+use std::rc::Rc;
+
+use ratatui::{buffer::Buffer, layout::{Constraint, Layout, Offset, Rect}, style::{Style, Stylize}, widgets::{Block, BorderType, Paragraph, Widget}};
+use yazi_core::notify::{Level, Message};
+
+use crate::{widgets::Clear, Ctx};
+
+pub(crate) struct Notify<'a> {
+	cx: &'a Ctx,
+}
+
+impl<'a> Notify<'a> {
+	pub(crate) fn new(cx: &'a Ctx) -> Self { Self { cx } }
+
+	fn chunks(area: Rect, messages: &[Message]) -> Rc<[Rect]> {
+		let chunks = Layout::horizontal([Constraint::Percentage(100), Constraint::Min(40)]).split(area);
+
+		Layout::vertical(messages.iter().map(|m| Constraint::Length(m.height() as u16)))
+			.spacing(1)
+			.split(chunks[1])
+	}
+}
+
+impl<'a> Widget for Notify<'a> {
+	fn render(self, area: Rect, buf: &mut Buffer) {
+		let notify = &self.cx.notify;
+
+		let limit = notify.limit();
+		let chunks = Self::chunks(area, &notify.messages[..limit]);
+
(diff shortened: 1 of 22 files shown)


yazi-fm/src/notify/notify.rs
```diff
@@ -0,0 +1,54 @@
...
+
+pub(crate) struct Notify<'a> {
+	cx: &'a Ctx,
+}
+
+impl<'a> Notify<'a> {
+	pub(crate) fn new(cx: &'a Ctx) -> Self { Self { cx } }
+
+	fn chunks(area: Rect, messages: &[Message]) -> Rc<[Rect]> {
+		let chunks = Layout::horizontal([Constraint::Percentage(100), Constraint::Min(40)]).split(area);
```
kdheepak, line 16: You may also write this as:

```
let chunks = Layout::horizontal([Constraint::Fill(1), Constraint::Length(40)]).split(area);
```

sxyazi: Awesome, will apply this suggestion in the next commit, thanks!

DreamMaoMao:
so cool! Can I call this notification component in the plug-in?

sxyazi:
> so cool! Can I call this notification component in the plug-in?

Yeah sure, it's designed for the plugin system, I will add related APIs to call it.

sxyazi:
> Yeah sure, it's designed for the plugin system, I will add related APIs to call it.

Done in https://github.com/sxyazi/yazi/pull/780

cbr9:
What is the purpose of having an built-in notification system? Wouldn't system-wide notifications make more sense? With this system, if you close yazi, you lose the notification. 

I don't exactly know how the client-server architecture will work. However, jobs could be delegated to the server, thus removing the need to keep yazi open for I/O operations. If that is the case, then only system notifications make sense here. And even if that is not the case, I would still like to have a list of notifications that is more persistent than my terminal, which I open and close all the time

sxyazi:
> What is the purpose of having an built-in notification system? Wouldn't system-wide notifications make more sense? With this system, if you close yazi, you lose the notification.

Currently, it is used for:

- Notifications when a user cannot open a file, such as when the text opener in Yazi is set to `$EDITOR` by default, which is often because the user has not configured it.
- Notifications when a synchronous shell command run by the user with `:` fails.
- Notifications when a user cannot select a particular file, such as when trying to select `/foo/bar` after already selecting `/foo`. Yazi's nested conflict detection will avoid this and provide a notification to the user.

These are all highly real-time and associated with user behavior. It seems that persisting them doesn't make much sense.

> I don't exactly know how the client-server architecture will work. However, jobs could be delegated to the server, thus removing the need to keep yazi open for I/O operations. If that is the case, then only system notifications make sense here. And even if that is not the case, I would still like to have a list of notifications that is more persistent than my terminal, which I open and close all the time

As for server architecture, there are currently no plans for that. There is a PR introducing a client-server architecture similar to `lf`, https://github.com/sxyazi/yazi/pull/826, but they are different, as Yazi does not start an additional server process in the background.

cbr9:
@sxyazi that makes quite a bit of sense. Thanks for the explanation!

Issue #652, opened 2024-02-09
Enable the ability to sort the which key


taylo2allen:
### Will you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

In the yazi.toml add [which] that has the methods sort_by & sort_sensitive that allows you to sort the which menu.

### Describe alternatives you've considered

One way for me to sort the which menu is to physically go to my keymap.toml and sort my keybinding definitions.

sxyazi:
Sounds good to me. Does anyone want to give it a try?

abhaysp95:
@taylo2allen I completed the rust basics and I have just started using Yazi. I would like to help. I want to understand few points for this issue:
* What is meant by sorting the which key ?
* Currently, sort_by & sort_sensitive option are under [manager]. What is wanted is to created a new table named [which] ?
* Can you kindly show the steps of what you are doing for 'alternatives' you mentioned

CC: @sxyazi

sxyazi:
Thank you very much for the help! @abhaysp95

I'll wait for OP's reply and provide more detailed suggestions on how to implement it :)

taylo2allen:
I have a lot of keykindings to move to different folders on my system for example:
`
	{ on = [ "g", "C" ], exec = "cd ~/.config", desc = "GoTo: ~/.config" },
	{ on = [ "g", "c", "b" ], exec = "cd ~/.config/xmobar", desc = "GoTo: ~/C/xmobar" },
	{ on = [ "g", "c", "f" ], exec = "cd ~/.config/fish", desc = "GoTo: ~/C/fish" },
	{ on = [ "g", "c", "q" ], exec = "cd ~/.config/qutebrowser", desc = "GoTo: ~/C/qutebrowser" },
	{ on = [ "g", "c", "r" ], exec = "cd ~/.config/ranger", desc = "GoTo: ~/C/ranger" },
	{ on = [ "g", "c", "x" ], exec = "cd ~/.config/xmonad", desc = "GoTo: ~/C/xmonad" },

`
in the code above I use "g", "c" as a leader key to go to all of the folders I have bookmarked in my .config folder. When I press the "g" key the whichkey menu pops up with the next possible keys you can press, I would like to sort the options for keybindings like "c", "b" to be shown first before "c", "x". The only way I found to sort the list of keybindings is to shift the definitions around in my keymap.toml file.
[...]

PR #662, opened 2024-02-11
feat: enable the ability to sort the which key

abhaysp95:
Closes https://github.com/sxyazi/yazi/issues/652 - @taylo2allen

* sort "which" window in following way:
    - sort_by: "key|desc|none"
    - sort_sensitive: true|false
    - sort_reverse: true|false


Rolv-Apneseth:
You should do a quick format on this - `cargo +nightly fmt`

abhaysp95:
Done the formatting changes

sxyazi:
Thank you for the PR! @abhaysp95 

I've made some little changes, would you like to give it a review?

yazi-core/src/which/sorter.rs
```diff
@@ -9,30 +11,31 @@ pub struct WhichSorter {
...
-				let a = a.on.iter().map(|c| c.to_string()).collect::<String>();
-				let b = b.on.iter().map(|c| c.to_string()).collect::<String>();
-				by_alphabetical(&a, &b)
-			}),
-			SortBy::Desc => items.sort_unstable_by(|a, b| {
-				// what if description isn't present (need to check if it's mandatory or not)
-				// in case if it is not present, should I just panic ?
-				by_alphabetical(a.desc.as_ref().unwrap(), b.desc.as_ref().unwrap())
-			}),
+		let mut new = Vec::with_capacity(indices.len());
```
Rolv-Apneseth, line 35: What do you think about this:

```rust
*items = indices.into_iter().map(|i| mem::take(&mut items[i])).collect();
```

Rolv-Apneseth:
Does the sorting work for you? I think commit b44452d breaks it for me.

yazi-core/src/which/sorter.rs
sxyazi: good idea, will apply this suggestion

sxyazi:
> Does the sorting work for you? I think commit [b44452d](https://github.com/sxyazi/yazi/commit/b44452dbd914142d1ef5b5edfd37716e23e2d2d2) breaks it for me.

Hi, I pushed a new commit should fix this.

Rolv-Apneseth:
Nice, that fixed it

abhaysp95:
> Thank you for the PR! @abhaysp95
> 
> I've made some little changes, would you like to give it a review?

It looks like I wrote the code in a verbose way. I tried to follow the example of how it was implemented for FileSorter. Maybe I thought that this will help in being more open to future changes. But I don't have any such problems with the change. All fine by me.

sxyazi:
Thanks for the review, alright let me merge it!

PR #663, opened 2024-02-11
ci: add clippy/rustfmt workflow

uncenter:
Adds a "check" workflow to run Clippy and rustfmt, also makes the workflows a little more consistent across files.


sxyazi:
Looks good to me, thank you alot!

PR #665, opened 2024-02-11
feat: add <Delete> to [input] keymap

grig-iv:


sxyazi:
Nice catch, thank you!

PR #674, opened 2024-02-14
feat: expose selected/yanked files as API

sxyazi:
Prepare for https://github.com/sxyazi/yazi/pull/646 - @Rolv-Apneseth

### DIFF
--- a/yazi-core/src/manager/commands/open.rs
+++ b/yazi-core/src/manager/commands/open.rs
@@ -34,7 +34,7 @@ impl Manager {
 		for u in selected {
 			if self.mimetype.get(u).is_some() {
 				done.push((u.clone(), String::new()));
-			} else if self.guess_its_folder(u) {
+			} else if self.guess_folder(u) {
 				done.push((u.clone(), MIME_DIR.to_owned()));
 			} else {
 				todo.push(u.clone());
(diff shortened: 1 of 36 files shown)


sxyazi:
Yeah it's already, I added a new `cx.yanked` and `cx.active.selected` property, it should simplify the code at https://github.com/sxyazi/yazi/pull/646.

Let me know if I missed something needed/better to have!

yazi-core/src/manager/commands/open.rs
```diff
@@ -98,13 +107,30 @@ impl Manager {
 		});
 	}
 
-	fn quit_with_selected(selected: &[&File]) -> bool {
+	fn guess_its_folder(&self, url: &Url) -> bool {
```
Rolv-Apneseth, line 110: Maybe just `guess_folder`?

Rolv-Apneseth:
With these changes, when in visual mode, yanking/cutting only does the file under the cursor.

sxyazi:
> With these changes, when in visual mode, yanking/cutting only does the file under the cursor

Yeah this is a known issue, I will fix it in the next PR. This PR focuses on the plugin interface

yazi-core/src/manager/commands/open.rs
sxyazi: Applied this suggestion, thanks!

Rolv-Apneseth:
Ah ok. All seems good to me then

Rolv-Apneseth:
I'm gonna be away from this computer for a day or two so I'm gonna just push the changes to #646 and rebase after this is merged

PR #679, opened 2024-02-14
feat: add `scrolloff` in `[manager]`

### LINKED ISSUE #675: Add a new `scrolloff` option to `[manager]`

_No response_

Currently its value is fixed to 5, that is, when moving the cursor in the file list, if there are fewer than 5 files below the cursor position, the list will automatically scroll. It would be great if it could be customized.

It is the same concept as `scrolloff` in Neovim, https://neovim.io/doc/user/options.html#'scrolloff'

_No response_

### Additional context

_No response_

dedukun:
This fixes #675 

I've made like in neovim where if the `scrolloff` is a large numbe, like e.g. `200`, the cursor will stay in the middle of the window.


sxyazi:
Thank you for the PR!

yazi-core/src/folder/folder.rs
```diff
@@ -108,8 +108,15 @@ impl Folder {
 		let len = self.files.len();
 
 		let limit = LAYOUT.load().current.height as usize;
+		let half_screen = limit / 2;
+		let scrolloff = if MANAGER.scrolloff as usize > half_screen {
+			half_screen
+		} else {
+			MANAGER.scrolloff as usize
+		};
```
sxyazi, line 116: Just tested in my neovim, seems this check is unnecessary - neovim scrolls even when `scrolloff` > `half_screen`.

dedukun: It does, however the cursor will be in the middle of the screen, without that check the cursor would be in the current line when you switch the direction of the movement.

With that check

https://github.com/sxyazi/yazi/assets/25795432/ea1c4dea-349f-4798-8819-0313a1966f67

Withouth that check

https://github.com/sxyazi/yazi/assets/25795432/ff420b67-8235-406f-b8b5-d1a38bec2fab

dedukun: From https://neovim.io/doc/user/options.html#'scrolloff'

> If you set it to a very large value (999) the cursor line will always be in the middle of the window (except at the start or end of the file or when long lines wrap).

dedukun: However if you don't want that behaviour I can remove the check

sxyazi: Oh thank you for the explanation! I understand now.

This is useful for users who want the list to always be centered; they just need to specify a very large value.

dedukun: Exactly, however it can't be too large because I used a `u8`, do you think it might be better to use a `u16`?

The `u8` should be enough, I don't think there will be users with more than `200` lines in the window, however some might want to just put a large value like `900`  and report an issue if there's an error in that situation.

sxyazi: Let's keep it as `u8` for now.

We can always increase it once we find it too small. However, if we use `u16` and later find it too large, attempting to shrink it could lead to breaking changes - user configurations being unparseable due to numeric overflow.

sxyazi:
Hi, I made some simplifications, would you like to give it a review?

dedukun (commented):
Yes it look much better that way, it shows my lack of experience in rust :sweat_smile:

sxyazi:

Let's merge it then

sxyazi:
> If you want I can also make a PR in the docs

That's great, please do it!

PR #683, opened 2024-02-15
feat: add color to icons

### LINKED ISSUE #678: Add color to icon rules for the file list
I am posting some simple tasks here as "good first issues" that don't require too much of my attention.

I need to allocate my limited time slice to tasks that require my direct involvement.

---

_No response_

Yazi supports setting foreground color (text color) for [specific icons through `fg`](https://yazi-rs.github.io/docs/configuration/theme#icon).

It would be really cool to add these colors in [the default theme](https://github.com/sxyazi/yazi/blob/main/yazi-config/preset/theme.toml), maybe we can refer to the colors in `nvim-web-devicons`, https://github.com/nvim-tree/nvim-web-devicons/blob/master/lua/nvim-web-devicons/icons-default.lua

_No response_

### Additional context

_No response_

Andy-W-Developer:
Resolves #678 

Uses the same colors in `nvim-web-devicons` if the same files exists there.

If it does not exist, use the same color in same group e.g. powerpoint, audio and movie files.

before
after

Sorry about the issue reference spam, I ran into a problem while amending the commit message.


sxyazi:
Thank you for the PR! I've tested it, it works great.

colors

Let's merge it then

PR #685, opened 2024-02-15
fix: update version output when commit SHA is not present

orhun:
Hey!

Arch maintainer of yazi here :wave:

I realized version flag (`-V`) shows the following for me:

```sh
yazi 0.2.3 (VERGEN_IDEMPOTENT_OUTPUT 2024-02-11)
```

This is due to git commits not being available at build time.

This PR updates the version output like following when commit SHA is not available:

```sh
yazi 0.2.3 (2024-02-11)
```

### DIFF
--- a/yazi-config/src/boot/boot.rs
+++ b/yazi-config/src/boot/boot.rs
@@ -58,10 +58,12 @@ impl Default for Args {
 		let args = Self::parse();
 
 		if args.version {
+			let commit_sha =
+				(env!("VERGEN_GIT_SHA") != "VERGEN_IDEMPOTENT_OUTPUT").then_some(env!("VERGEN_GIT_SHA"));
 			println!(
-				"yazi {} ({} {})",
+				"yazi {} ({}{})",
 				env!("CARGO_PKG_VERSION"),
-				env!("VERGEN_GIT_SHA"),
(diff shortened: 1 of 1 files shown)


sxyazi:
Hi, thank you for the PR!

Would it be possible to add the name of the distribution/package manager during building? This way, when users submit bug reports, I can know their installation method or the distribution they're using.

For example, Homebrew adds:

```ruby
ENV["VERGEN_GIT_SHA"] = tap.user
```

And it looks like:

```bash
$ yazi -V
yazi 0.2.3 (Homebrew 2024-02-07)
```

orhun:
Hello,

> Would it be possible to add the name of the distribution/package manager during building?

Good idea! Does `vergen` provide this information though? Or do you mean getting this info from the OS itself? (with a 3rd party crate?)

> ENV["VERGEN_GIT_SHA"] = tap.user

Where is `tap.user` coming from in this case?

orhun:
Ah I see! That's a nice hack.

I applied it in https://gitlab.archlinux.org/archlinux/packaging/packages/yazi/-/commit/e005d6a9806568e4dfc6337a1d432cf4805329e8

Then I guess there is no need for this PR 🐻

sxyazi:
Thanks for the quick fix! Let me close this PR then

Issue #688, opened 2024-02-16
Nested selection conflict detection


sxyazi:
If anyone is interested in this issue, please reply below!

---

### Will you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

I'm trying to add support for cross-directory selection, which will allow users to select files from different directories at the same time and perform actions on them (copy, paste, etc.).

However, we must avoid situations of nested selection. For example, a user can select both `/a/b`, and `/a/c`, but cannot select `/a`, `/a/b`, nor `/a/b/c`, `/a/b`, as this would cause abnormalities in functions like copy, cut.

To achieve this, we need to modify the existing structure:

https://github.com/sxyazi/yazi/blob/53525b511efe165b0a5227263e99e81697213033/yazi-core/src/tab/tab.rs#L18

to:

```rust
pub selected: Selected
```

and `Selected` providing two `add()` and `remove()` APIs:

```rust
pub struct Selected {
    inner: BTreeSet<Url>,
    // Other fields
}

impl Selected {
    pub fn add(&mut self, url: &Url) -> bool {}
    pub fn remove(&mut self, url: &Url) {}
}
```

`add()` and `remove()` will add the passed URL to the `BTreeSet<Url>` representing the selected items. If a conflict occurs, `add()` returns `false`; otherwise, it returns `true`. The pseudocode be like:

```rust
s1.add("/a") // true
s1.add("/a/b") // false (conflict)

s2.add("/a/b/c") // true
s2.add("/a/b") // false (conflict)
s2.add("/a/b/d") // true
```

evpeople:
Hi !
I am very interested in this issue and would be happy to handle it.

sxyazi:
Thanks! @evpeople

og900aero:
I'm very interested too :)

PR #689, opened 2024-02-17
feat: nested selection conflict detection

evpeople:
I implemented conflict path detection using HashMap. Using a prefix tree might also be a worthwhile option to consider.

Closes https://github.com/sxyazi/yazi/issues/688

### DIFF
--- a/yazi-core/src/tab/selected.rs
+++ b/yazi-core/src/tab/selected.rs
@@ -91,9 +91,9 @@ mod tests {
 		let url1 = Url::from(Path::new("/a/b"));
 		let url2 = Url::from(Path::new("/c/d"));
 
-		assert!(selected.insert(url1), "Should successfully insert url1");
-		assert!(selected.insert(url2), "Should successfully insert url2");
-		assert_eq!(selected.inner.len(), 2, "There should be two URLs");
+		assert!(selected.insert(url1));
+		assert!(selected.insert(url2));
+		assert_eq!(selected.inner.len(), 2);
 	}
 
 	#[test]
@@ -102,8 +102,8 @@ mod tests {
 		let parent_url = Url::from(Path::new("/a"));
 		let child_url = Url::from(Path::new("/a/b"));
 
-		assert!(selected.insert(parent_url), "Should successfully insert parent_url");
-		assert!(!selected.insert(child_url), "Should fail to insert child_url due to conflict");
+		assert!(selected.insert(parent_url));
+		assert!(!selected.insert(child_url));
 	}
 
 	#[test]
@@ -113,9 +113,9 @@ mod tests {
 		let parent_url = Url::from(Path::new("/a/b"));
 		let sibling_url = Url::from(Path::new("/a/b/d"));
 
-		assert!(selected.insert(child_url), "Should successfully insert child_url");
-		assert!(!selected.insert(parent_url), "Should fail to insert parent_url due to conflict");
-		assert!(selected.insert(sibling_url), "Should successfully insert sibling_url");
+		assert!(selected.insert(child_url));
+		assert!(!selected.insert(parent_url));
+		assert!(selected.insert(sibling_url));
 	}
 
 	#[test]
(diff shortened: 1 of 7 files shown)


sxyazi:
Thank you for the PR!

Could you also add an `insert_many(&mut self, urls: &[&Url], parent: Url)` method? This is very useful for bulk selecting (selecting all) files - all of which have the same parent path, so it only needs to be detected once to improve performance.

Then `insert()` can simply be a wrapper for `insert_many()` like this:

```rust
	pub fn insert(&mut self, url: &Url) -> bool {
            self.insert_many(&[url], url.parent_url())
        }
```

evpeople:
Ok, I will try it

evpeople:
Considering that the paths corresponding to insert_many all have the same parent directory, 
I implemented insert_many by modifying the logic in the original insert method that checks the object being inserted to instead check the first object in insert_many.

I didn't use the parent variable. I tried some methods that involved the parent variable, but they were not successful in passing the tests.

evpeople:
To make the compilation pass, I added contains and get_inner methods. I am unsure if using the BTree inside inner within yazi-fm might lead to inconsistencies between HashMap and BTree.

yazi-core/src/tab/selected.rs
```diff
@@ -0,0 +1,251 @@
...
+	}
+	pub fn new() -> Self { Selected { inner: BTreeSet::new(), parents: HashMap::new() } }
+
+	pub fn insert(&mut self, url: Url) -> bool { self.insert_many(&[&url]) }
+
+	pub fn insert_many(&mut self, urls: &[&Url]) -> bool {
+		if urls.is_empty() {
+			return false;
+		}
+		let url_buf = urls[0].to_path_buf();
```
evpeople, line 23: I simply changed the `url` in this line to `urls[0]`

yazi-core/src/tab/selected.rs
```diff
@@ -0,0 +1,251 @@
...
+
+		let mut current_path = url_buf.clone();
+		let len_of_urls = urls.len();
+		loop {
+			current_path = match current_path.parent() {
+				Some(parent) => parent.to_path_buf(),
+				None => break,
+			};
+			let counter = self.parents.entry(current_path.clone()).or_insert(0);
+			*counter += len_of_urls;
```
evpeople, line 45: Additionally, in this line, I changed from adding 1 to adding the length of `urls`.

yazi-core/src/tab/selected.rs
sxyazi: Good idea!

yazi-core/src/tab/selected.rs
```diff
@@ -0,0 +1,223 @@
...
+		if !self.inner.remove(url) {
+			return false;
+		}
+
+		let mut current_path = url.clone();
+		while let Some(parent) = current_path.parent_url() {
+			current_path = parent;
+
+			let counter = self.parents.entry(current_path.clone()).or_insert(0);
+			*counter -= 1;
```
sxyazi, line 53: When `current_path` does not exist, inserting `0` and then `0 -= 1;` will cause an overflow.

evpeople:
The usage of Deref here has taught me a lot.

evpeople:
I have considered the issue of data overflow. Later, I realized that if the add operation is correct, then the count in the parent can safely be decremented to 0. If overflow occurs, it indicates the possibility of some concurrency-related modifications. At this point, the program should panic.

yazi-core/src/tab/selected.rs
```diff
@@ -36,7 +40,7 @@ impl Selected {
 			*self.parents.entry(u).or_insert(0) += urls.len();
 		}
 
-		self.inner.extend(urls.iter().cloned().cloned());
+		self.inner.extend(urls.iter().map(|&u| u.clone()));
```
evpeople, line 43: I've tried many ways to get this to compile here, thank you very much for your approach, it's been very enlightening for me.

sxyazi:
Hi, I made some little changes, and I believe it's all good now.

Thank you very much for contributing to this feature, let me merge it!

PR #690, opened 2024-02-17
refactor: using definite patterns to simplify keybindings

sxyazi:


Brixy:
Thanks, @sxyazi, for all your work and excellent additions to yazi.

I noticed the following issue with this pull request:

When you do a `yp`, files vanish. I don’t know what happens technically, but the files are not moved into the trash folder.

yazi-core/src/manager/commands/paste.rs
```diff
@@ -18,10 +18,11 @@ impl Manager {
 		let opt = opt.into() as Opt;
 
 		let dest = self.cwd();
-		if self.yanked.cut {
-			tasks.file_cut(&self.yanked, dest, opt.force);
-		} else {
+		if !self.yanked.cut {
 			tasks.file_copy(&self.yanked, dest, opt.force, opt.follow);
```
Rolv-Apneseth, line 22: Is the bug possibly cause of a missing `return` here?

sxyazi: oops I missed it in reviewing, let me fix it

sxyazi: fixed in https://github.com/sxyazi/yazi/commit/2f784e7ae71c07e02f318f9a6d3c45c586253e59

Rolv-Apneseth: Nice

Rolv-Apneseth:
Thanks for reporting @Brixy

Issue #319, opened 2023-10-24
[RFC] Switch to move/copy concept instead of yank/cut/paste


Brixy:
### Is your feature request related to a problem? Please describe

Hi guys!

This may be a far-reaching idea. If it does not fit, please just close this issue ;-)

Currently you can `yank` or `yank --cut` single files or a selection of files. In the target directory you `paste` the files. This is probably how most TUI file managers work.

### Will you be willing to contribute this feature?

Sorry – I am not a Rust expert.

### Describe the solution you'd like

nnn uses a different concept: You just select one or more file, and in the target directory you either move (`v`) or copy (`p`) these files.

This seems more efficient, and I could think of **these advantages:**

- You need to decide only in the target directory what to do with selected files.
- It is arguably easier to copy or move selected files from multiple directories into a single target folder because there is no need to differentiate between yanking and cutting in advance.
- You just need to color code selected files because there is no need to differentiate between selected, yanked, and cut files.
- It’s actually easier to understand. Currently, if you select and then yank two files, then “re-select” one of those files, there is no visible indication. If you then hit `d` for example, it’s a bit hard to predict what yazi — and other file managers — will do. The proposed concept just knows two states: selected or unselected. Moving, deleting, and copying only refers to selected files.
- It would be sufficient to just support counting selected files. Again no need to support counting selected, cut, and yanked files, and no need to think about e.g. cutting and then re-selecting files.
- You can simply unselect files and move or copy the rest of the selection.
- No need to support “un-yanking” (see [#313](https://github.com/sxyazi/yazi/pull/313)) or „un-cutting“ files.
- The current concept for deleting multiple files (first select, then delete) can be transferred to moving and copying.

**Possible disadvantages:**

- People are used to the current vim-like concept coming from other TUI file managers.
- The proposed concept would not support moving _and_ copying files with one single paste command (which is currently not possible in yazi).

Currently if you want to cut and paste a single file you just need to press `x` and `p`.With the proposed concept it would be `<Space>` and `p`. There wouldn’t be much difference.

15cm:
In the proposal, how do you tell when your selection is finalized or not, under these scenarios:
- Yank multiple batches of files in the same directory with multiple select operations and then paste then to another directory with one hit.
- Yank multiple batches of files in the different directories with multiple select operations and then paste then to another directory with one hit.

dan-da:
@15cm what does it mean for selection(s) to be finalized?

[...]

PR #693, opened 2024-02-18
feat: cross-directory selection

sxyazi:
An implementation of the feature request: https://github.com/sxyazi/yazi/issues/319 - @Brixy

Closes https://github.com/sxyazi/yazi/issues/319

If you want to enable this feature, please use the latest default configuration file. For users who are using the old configuration file, the original behavior will remain unchanged to avoid confusion.


musjj:
> If you want to enable this feature, please use the latest default configuration file.

Can you elaborate on this? Which option should you change to enable this new feature?

sxyazi:
@musjj 

In the previous default configuration, `h` and `l` include an `escape` command, which means the selection is cleared every time the `CWD` is switched.

If you want to use the selection across directories and have copied the default configuration from a previous version, please remove `escape`, otherwise it will continue to behave in the original mode.

https://github.com/sxyazi/yazi/blob/0d7e977076faaa8803ce6179c57e9bd868d38a1c/yazi-config/preset/keymap.toml#L34-L35

musjj:
I see, thanks for the clarification. So what this PR adds is persistent selections when navigating across directories. If you want to mimic the nnn feature requested in the linked issue, you'd need to make your own keybinding as suggested by @og900aero above.

PR #702, opened 2024-02-20
fix: allow running `ya.sync` in a sync plugin call

dedukun:
This fixes a issue (yazi would panic) I encountered when calling an async plugin that uses `ya.sync` synchronously, which happened because I was creating a `setup` function for a plugin.

### DIFF
--- a/yazi-plugin/preset/ya.lua
+++ b/yazi-plugin/preset/ya.lua
@@ -29,6 +29,10 @@ function ya.flat(t)
 end
 
 function ya.sync(f)
+	if ya.SYNC_ON then
+		return function(...) f(...) end
+	end
+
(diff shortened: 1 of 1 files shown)


sxyazi:
Thanks! Looks good to me.

dedukun:
@sxyazi I think it's best to revert this change, I though that I had tested this but it seems that this solution is not working as expected.

When trying to add a bookmark with https://github.com/dedukun/bookmarks.yazi the plugin panics with the error
```
  2024-02-20T22:50:43.706643Z ERROR yazi::lives::lives: error converting Lua nil to function
    at yazi-fm/src/lives/lives.rs:66
```

For some reason the if condition with `ya.SYNC_ON` is met when using the `ya.sync` method in `entry` asynchronously (even though it works correctly if the method exists but is not called, which makes it seem that it is loaded correctly).

Issue #703, opened 2024-02-20
Acessing `state` in plugin's `setup` function


dedukun:
### What system are you running Yazi on?

Linux Wayland

### What terminal are you running Yazi in?

wezterm 20240123-102946-02bdd7ed

### Yazi version

yazi 0.2.3 (8efafd6 2024-02-19)

### Did you try the latest main branch to see if the problem got fixed?

Tried, but the problem still

### Describe the bug

Yazi panics when trying to access the `state` variable inside a `setup` function as demonstrated in https://yazi-rs.github.io/docs/plugins/overview#sync-vs-async.

### Expected Behavior

It should be possible to set variables in the `state` so that the plugins can have custom configurations.

### To Reproduce

I've created a [new branch in bookmarks.yazi](https://github.com/dedukun/bookmarks.yazi/tree/setup) which has a simple `setup` function that tries to set a table in the `state` variable.

To recreate the issue, simply call the `setup` method in `init.lua`

### Anything else?

I've looked a bit into the lua API and I think that the issue is because the `require` just runs the lua code without the context of `YAZI_PLUGIN_NAME`.

A possible solution might be to have a custom `require` that does something similar to [entry](https://github.com/sxyazi/yazi/blob/8efafd6c9104684920c819aa3fc762d9d61689e1/yazi-plugin/src/isolate/entry.rs#L7).

dedukun:
This is a bit off topic, but in the example that I gave I'm using `print` in the `setup` function instead of `ya.err` because [`init.lua`](https://github.com/sxyazi/yazi/blob/2e9b6b7b7b7bd2d0b70232f4cf776b2a4cc63110/yazi-fm/src/main.rs#L47) is executed before the [logs are initialized](https://github.com/sxyazi/yazi/blob/2e9b6b7b7b7bd2d0b70232f4cf776b2a4cc63110/yazi-fm/src/app/app.rs#L20).

Do you want me to make a PR to fix this?

dedukun:
> Thank you for the report, this is a known issue, and I am still considering how to address it.
> 
> You can use `state("your-plugin-name")` as a workaround for now:
> 
> ```lua
> return {
>   setup = function(self, opts) 
>     local st = state("your-plugin-name")
>     st.foo = "bar"
>   end
> }
> ```
> 
> Needs [0cb572a](https://github.com/sxyazi/yazi/commit/0cb572a83787a88c8e6ef30aec148d509902a659)

[...]

PR #708, opened 2024-02-20
fix: initialize logs right after configurations

dedukun:
As we talked in https://github.com/sxyazi/yazi/issues/703#issuecomment-1954797994 this PR makes it so that the logs are initialized in the beginning of the application.

I made it run right after the `yazi_config::init();` because I'm not sure if there's any log that might occur if any error ever happens when initializing `yazy_core` or the `yazi_scheduler`.


sxyazi (approved):

sxyazi:
Thanks!

PR #710, opened 2024-02-21
perf: cache loaded plugins

sxyazi:
Resolves https://github.com/sxyazi/yazi/issues/703.

This PR removes the global variable `state` and passes it as a parameter, sync plugin:

```diff
return {
-  entry = function()
+  entry = function(state)
    state.foo = 1;
  end
}
```

Also sync block:

```diff
- local get_hovered_url = ya.sync(function()
+ local get_hovered_url = ya.sync(function(state)
  state.foo = 1;
end)
```

This allows the plugin to access the `state` during initialization:

```lua
-- ~/.config/yazi/plugins/test.yazi/init.lua
return {
  setup = function(state, opts)
    state.my_opt1 = opts.opt1
  end
}
```

```lua
-- ~/.config/yazi/init.lua
require("test"):setup {
  opt1 = "value1"
}
```


dedukun:
Hey, I tested a bit and the issue that was introduced in #702 is fixed, and I'm also able to get access to the `state` variable as you described, however something seems wrong with `state` inside the `entry` function, I can't access variables defined from `setup` in `entry`, however if I call a `ya.sync` method in `entry` I can access that variable.

The issue can be seen in the following test plugin, the print inside `entry` returns `nil`, but all other prints have the correct value for `state.opt`.


```lua
local sync = ya.sync(function(state)
	print("SYNC: " .. tostring(state.opt))
end)

local sync2 = ya.sync(function(state)
	ya.err("SYNC2: " .. tostring(state.opt))
end)

return {
	entry = function(state)
		ya.err("ENTRY: " .. tostring(state.opt))

		sync2()
	end,
	setup = function(state)
		state.opt = "test"

		sync()
	end,
}
```

sxyazi:
Thanks for testing! Let's merge then

PR #720, opened 2024-02-24
feat: add FiletypeIs to icons

### LINKED ISSUE #656: Allow { is = "" } configuration for icon rules in theme.toml

_No response_

Currently, I am able to set the style of a line item based on its properties (exec, orphan, link, et. c.) by adding `is = "{prop}"` within the filetype rules section of theme.toml.  It would be nice to have this same option in the icon rules sections as well, so that, for example, executable can be given their own icon, regardless of their file name.

_No response_

### Additional context

_No response_

Andy-W-Developer:
Resolves #656 

**Environment**
OS: Windows 11
Terminal: wezterm

**Tests**
- [x] Symbolic Links
- [ ] Executables

is = "exec" did not recognize .exe files, for both Filetype and Icon rules, is = "link" worked for both.

**Conditions**
Under Icon rules:
{ name = "*", is = "link", text = "󰿃", fg = "magenta" },

Before
before
After
after


sxyazi:

The issue of `exec` not working on Windows is a known problem, I'm not quite sure which files on Windows are executable. It seems that besides `.exe`, files like `.bat` are also executable. What do you think?

Andy-W-Developer:
The PathExt environmental variable has a list of executable file extensions:
`.com` `.exe` `.bat` `.cmd` `.vbs` `.vbe` `.js` `.jse` `.wsf` `.wsh` `.msc`

Maybe `.msi` files should be added because they are found next to `.exe` installers.

**Examples**
7zip and Firefox has `.exe` and `.msi` installers.
Adoptiums build of OpenJDK only has `.msi` installers.

PR #732, opened 2024-02-26
Separate Style of Inner Section of the Status bar from Outer

asim215:
Inner section of status line now is independent from outer section and current mode:

- easier to read:
  - mode doesn't influence inner section -> less distraction while switching modes.
  - inner section can now have bold style for size and percent.
- style independent of other sections of status bar

Need to remove comments on theme.toml:51, where I question need of separator_style field. It looks like it used in cand.rs:32.

inner_outer

20240226_17h50m02s_grim

Preview:

https://github.com/sxyazi/yazi/assets/7127165/ded7afff-2a9b-4134-9009-fe0322eee6ac


sxyazi:
Thank you for the PR and detailed explanation.

It seems that now the text color in the inner part does not change with the mode switch, it looked like this before:


https://github.com/sxyazi/yazi/assets/17523360/cbbc9cc2-01d8-4ceb-8c5e-3837e890d290

asim215:
Yes. Here is my case comparison before and after change:

https://github.com/sxyazi/yazi/assets/7127165/d263e97a-6746-4d7b-83e3-603bf2782c68

Here is why I have problem with current implementation:

1. With light colors in light terminal it became hard to read status bar info.
    - I use two colors as background colors in gruvbox light theme in normal mode: bg2=`#d5c4a1`, yellow=`#d79921`.
      Because inner section is depends of colors of outer section I can't make font color inside section all dark color: fg=`#3c3836`, fg0=`#282828`. It is always inversed based on outer section inner makes `fg=bg_outer, bg=fg_outer`. So combination of all dark font on light colors are impossible.
    - I also colorblind (red-green) so I need distinct colors to see text on screen.
    - I need text of inner section in bold, same as bold in outer section to read easily. I cannot set that in current settings of theme.toml.
2. Outer section in code are combined with separator symbol. So as color of outer section change, separator also changes. Basically setting `separator_style` doesn't have any affect on outer separator symbol. It is only useful for inner separator symbol which isn't depends on inner section, therefore can be any color. I can't find here any logic why? Maybe only as some sort of indicator? So after changes, now `separator_style` can me omitted because it takes his color from adjacent section. I have find it usage in one place cand.rs:32, don't know how it will affect after deletion that from theme.toml.
    - Before: (separator+outer_section)[outer_color, outer_style] | (inner_section)[inverse_outer_section_color] | (separator)[separator_color]
    - After: (separator+outer_section)[outer_color, outer_style] | (inner_section+separator)[inner_color, inner_style] 
4. I think it is sufficient to mark mode by color only in outer section. I don't think that there is any need to change inner section after changed mode. For me it brings: 1) distraction; 2) need to carefully choose colors for each of mode so it will look readable on light `bg2` color (complexity).
5. It is hard to read values of `size`, `percent` without bold font. Bold style is previously impossible to set for these sections of text in status bar. 

Commit require:

- [ ] Decide for need of `separator_style` after change? theme.rs, theme.toml
- [ ] Remove extra comments
- [ ] Format code to style of project

I can send you my `theme.toml`, so you will see for yourself.

asim215:
Here updated gruvbox light theme with current behavior:

https://github.com/sxyazi/yazi/assets/7127165/8931c000-9f1b-4df8-bf8e-6928872e0cb3

It is basically unreadable for me with light colors. aqua = `#8ec07c`, blue = `#83a598`.
```
mode_select = { fg = "#3c3836", bg = "#8ec07c", bold = true }
mode_unset  = { fg = "#3c3836", bg = "#83a598", bold = true } 
```

sxyazi:
> Yes. Here is my case comparison before and after change:
> 
>  Kooha-2024-02-27-12-52-30.mp4 
> Here is why I have problem with current implementation:
> 
> 1. With light colors in light terminal it became hard to read status bar info.
>    
>    * I use two colors as background colors in gruvbox light theme in normal mode: bg2=`#d5c4a1`, yellow=`#d79921`.
>      Because inner section is depends of colors of outer section I can't make font color inside section all dark color: fg=`#3c3836`, fg0=`#282828`. It is always inversed based on outer section inner makes `fg=bg_outer, bg=fg_outer`. So combination of all dark font on light colors are impossible.
>    * I also colorblind (red-green) so I need distinct colors to see text on screen.
>    * I need text of inner section in bold, same as bold in outer section to read easily. I cannot set that in current settings of theme.toml.
> 2. Outer section in code are combined with separator symbol. So as color of outer section change, separator also changes. Basically setting `separator_style` doesn't have any affect on outer separator symbol. It is only useful for inner separator symbol which isn't depends on inner section, therefore can be any color. I can't find here any logic why? Maybe only as some sort of indicator? So after changes, now `separator_style` can me omitted because it takes his color from adjacent section. I have find it usage in one place cand.rs:32, don't know how it will affect after deletion that from theme.toml.
>    
>    * Before: (separator+outer_section)[outer_color, outer_style] | (inner_section)[inverse_outer_section_color] | (separator)[separator_color]
>    * After: (separator+outer_section)[outer_color, outer_style] | (inner_section+separator)[inner_color, inner_style]
> 3. I think it is sufficient to mark mode by color only in outer section. I don't think that there is any need to change inner section after changed mode. For me it brings: 1) distraction; 2) need to carefully choose colors for each of mode so it will look readable on light `bg2` color (complexity).
> 4. It is hard to read values of `size`, `percent` without bold font. Bold style is previously impossible to set for these sections of text in status bar.
> 
> Commit require:
> 
> * [ ]  Decide for need of `separator_style` after change? theme.rs, theme.toml
> * [ ]  Remove extra comments
> * [ ]  Format code to style of project
> 
> I can send you my `theme.toml`, so you will see for yourself.

Ah I see your point, it took me some time to understand.

From the video, it seems that only the combination of `#d5c4a1` and `#d79921` causes a decrease in distinguishability, I think this is because they are very similar colors. Is it possible to work around this issue by changing the color combination?

I do not quite want to change the current theme scheme for the following reasons:

1. It will reduce the distinguishability when switching modes, when I use the code of this PR, mode switching only occurs in the bottom left (or bottom right) cell. Compared to the previous version, the recognizable area is smaller.
2. The newly introduced "inner" is somewhat confusing, such as file names, permissions, and components added by users themselves (such as user groups), they are not controlled by "inner", but they are within the "inner" area.
3. The status bar becomes ugly. 🤣

sxyazi:
I'm going to close this because it still needs some work and hasn't seen any activity recently. @asim215 thanks for the effort!

PR #738, opened 2024-02-27
feat: add hovered as `$0` for shell and opener

### LINKED ISSUE #716: Add `$0` for the currently selected file to the `shell` command

In the `shell` command, we can use `$1`, `$2`, and `$n` to reference the first selected file, the second selected file, up to the nth selected file.

However, when the first selected file is not necessarily the file currently hovered over, `$0` is used to represent the file currently hovered over, disregarding the selection status.

`$0` has been reserved, and here is the place that needs to be modified:

https://github.com/sxyazi/yazi/blob/e51e8ad789914b2ab4a9485da7aa7fbc7b3bb450/yazi-plugin/src/external/shell.rs#L43

Just need to get the hovered file before executing the shell and pass it as `$0`.

### Additional context

_No response_

rrveex:
Closes #716 

New function `hovered_and_selected`.

### DIFF
--- a/yazi-config/preset/yazi.toml
+++ b/yazi-config/preset/yazi.toml
@@ -26,27 +26,27 @@ ueberzug_offset = [ 0, 0, 0, 0 ]
 
 [opener]
 edit = [
-	{ exec = '${EDITOR:=vi} "$@"', desc = "$EDITOR", block = true, for = "unix" },
+	{ exec = '${EDITOR:=vi} "$0"', desc = "$EDITOR", block = true, for = "unix" },
 	{ exec = 'code "%*"',          orphan = true, for = "windows" },
 ]
 open = [
-	{ exec = 'xdg-open "$@"',                desc = "Open", for = "linux" },
-	{ exec = 'open "$@"',                    desc = "Open", for = "macos" },
-	{ exec = 'start "" "%1"', orphan = true, desc = "Open", for = "windows" }
+	{ exec = 'xdg-open "$0"',                desc = "Open", for = "linux" },
+	{ exec = 'open "$0"',                    desc = "Open", for = "macos" },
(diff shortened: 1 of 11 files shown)


rrveex:
`yazi.toml` - to keep preivious behaior, it should have been:

`exec $@` -> `exec $0 $@`

I think people more often do `mpv f.avi` than `mpv f1.avi f2.avi ...`

yazi-config/preset/yazi.toml
```diff
@@ -26,27 +26,27 @@ ueberzug_offset = [ 0, 0, 0, 0 ]
 
 [opener]
 edit = [
-	{ exec = '${EDITOR:=vi} "$@"', desc = "$EDITOR", block = true, for = "unix" },
+	{ exec = '${EDITOR:=vi} "$0"', desc = "$EDITOR", block = true, for = "unix" },
```
sxyazi, line 29: Using `$@` instead of `$0` can better respect the user's behavior. When the user selects a single file or uses `open --hovered`, it will only open that file. When the user selects multiple files, it will open all of them.

sxyazi: Yeah sure.

`$0` => Always the file the user hovered over.
`$@` => Always the files the user selected.

When the user hasn't selected any files, only hovered over a file, the hovered file is treated as the selected file.

sxyazi:
Hi, I've also added support for `$0` for the opener, I think it should be fine now, would you like to give it a review?

rrveex:
> Hi, I've also added support for `$0` for the opener, I think it should be fine now, would you like to give it a review?

Yep, I guess that's how it's supposed to look like. Thanks for not throwing right away my very first pseudo-rust-gibberish :)

sxyazi:
Thanks for the review, let's merge it now!

PR #742, opened 2024-02-27
Default shell setting for exec

rrveex:
#718 - sorry, didn't wait for comments...

Perhaps `MANAGER` isn't the best place for this setting?

### DIFF
--- a/yazi-config/preset/yazi.toml
+++ b/yazi-config/preset/yazi.toml
@@ -12,6 +12,7 @@ linemode       = "none"
 show_hidden    = false
 show_symlink   = true
 scrolloff      = 5
+shell          = "sh"
 
 [preview]
 tab_size        = 2
--- a/yazi-config/src/manager/manager.rs
+++ b/yazi-config/src/manager/manager.rs
@@ -20,6 +20,7 @@ pub struct Manager {
 	pub show_hidden:  bool,
 	pub show_symlink: bool,
 	pub scrolloff:    u8,
+	pub shell:        String,
 }
 
 impl Default for Manager {
--- a/yazi-plugin/src/external/shell.rs
+++ b/yazi-plugin/src/external/shell.rs
@@ -2,6 +2,7 @@ use std::{env, ffi::OsString, process::Stdio};
 
 use anyhow::Result;
 use tokio::process::{Child, Command};
+use yazi_config::MANAGER;
 
 pub struct ShellOpt {
 	pub cmd:    OsString,
@@ -33,7 +34,7 @@ pub fn shell(opt: ShellOpt) -> Result<Child> {
 
 	#[cfg(unix)]
 	return Ok(unsafe {
-		Command::new("sh")
+		Command::new(&MANAGER.shell)
 			.env("YAZI_LEVEL", (level + 1).to_string())
 			.arg("-c")
 			.stdin(opt.stdio())


sxyazi:
I don't think this option makes much sense, I tested several shells I have installed, and only `zsh` can be working:

```
$ zsh -c 'echo $1' "" "1"
1

$ fish -c 'echo $1' "" "1"

$ nu -c 'echo $1' "" "1"
Error: nu::parser::variable_not_found

  × Variable not found.
   ╭─[source:1:1]
 1 │ echo $1
   ·      ─┬
   ·       ╰── variable not found.
   ╰────
```

Therefore, its value can only be `sh` or `zsh` (where `sh` is generally `bash` or `dash`).

And it is Unix-specific - Windows users cannot change their default shell using `shell=pwsh` or `shell=nu`, which would confuse them, I also want to avoid introducing platform-specific features.

sxyazi:
As for the `read: arg count` error in that discussion, I think it can be resolved by simply changing `read` to `read _`.

rrveex:
```
# fish -c 'echo $argv' "0" "1" 
0 1
```
Don't know what a `nu` is :)

Of course the `read` thing is trivial. Even though, it's also present in built-ins, like archive/unpack, dash users would have to overwrite those to experience same behavior as in bash. Perhaps change `read` to `read _` in there?

The idea was for more complex scripting. But I guess that's for `lua` plugins, even though I find that much more difficult to write then shell (not because of `lua` vs `sh`, it's the integration of "what I want to happen" in the yazi-plugin-system that's cumbersome).

Windows users would just... not change the default :)

I understand you wanting to avoid platform specific code though.

sxyazi:
> ```
> # fish -c 'echo $argv' "0" "1" 
> 0 1
> ```

This will require us to parse their format based on the different shell configurations of the users, such as `$argv` for `fish`.

> Don't know what a nu is :)

It's [Nushell](https://www.nushell.sh/), and it's cross-platform, so it can be used on both Unix and Windows.


> Of course the `read` thing is trivial. Even though, it's also present in built-ins, like archive/unpack, dash users would have to overwrite those to experience same behavior as in bash. Perhaps change `read` to `read _` in the built-ins?

Yes, [`exiftool`](https://github.com/sxyazi/yazi/blob/fa2632eda97b3bc64370648efa4c5084da4de1ee/yazi-config/preset/yazi.toml#L40) and [`mediainfo`](https://github.com/sxyazi/yazi/blob/fa2632eda97b3bc64370648efa4c5084da4de1ee/yazi-config/preset/yazi.toml#L49) use the `read` command, and we need to fix it in the default configuration file.

> The idea was for more complex scripting. But I guess that's for `lua` plugins,

Yes, rather than writing complex shell scripts, I would recommend creating it as a plugin. Shells have their limitations, such as:

1. They cannot access data from other tabs, so operations like diffing files selected in two different tabs are not possible.
2. They cannot access the plugin interfaces, such as `ya.which`, and soon I will add `ya.input` (for requesting user input for file names when creating archives) / [`ya.notify`](https://github.com/sxyazi/yazi/pull/659) (for displaying a foreground notification after a background task completes).
3. Plugins are more portable than shell scripts. They can usually be used on both Unix and Windows, while Unix and Windows have different shells (`sh` and `cmd.exe`), which are difficult to reconcile, and even within Unix, there are different shells.

> even though I find that much more difficult to write then shell (not because of `lua` vs `sh`, it's the integration of "what I want to happen" in the yazi-plugin-system that's cumbersome)

What tasks are difficult to accomplish using plugins/`sh`?

> Windows users would just... not change the default :)

Many users expect to use the same Yazi configuration on both Windows and Unix (which also Yazi strives to achieve). They may be confused when settings like `shell=nu` only work on Unix, or when they try to change `cmd.exe` to PowerShell using `shell=pwsh` and it doesn't work.

PR #744, opened 2024-02-27
ci: fix erroneous usage of runs-on in the cachix workflow

musjj:
Whoops, looks like I messed up the syntax. The job is currently stuck on `Waiting for a runner to pick up this job...`, because I'm not using the `runs-on` property correctly.

You need to use a matrix if you want the job to be run by multiple runners.


sxyazi:
Ah yep, thanks for the quick fix!

PR #745, opened 2024-02-27
fix: correct the path to the completion files

musjj:
Looks like the current build is failing due to the wrong path to the completion files.


sxyazi:
Thanks for the fix!

PR #750, opened 2024-02-28
Allow both '\' and '/' character in folder creation

ndtoan96:
To create folder on Windows, currently I have to use `\`. But this character is hard to type. So, for convenient sake, let's allow both `/` and `\` just in folder creation operation.


ndtoan96:
Sorry, it seems some old code got mixed in. I'll close this PR and create a new one later.

PR #751, opened 2024-02-28
feat: allow both `/` and `\` in folder creation

ndtoan96:
On Windows, I have to use `\` for folder creation but this character is hard to type. So, for convenient sake, let's allow both `/` and `\` just in folder creation.


sxyazi:
LGTM, thanks!

PR #752, opened 2024-02-28
feat: support `YAZI_FILE_ONE` environment variable for `file(1)` path

ndtoan96:
The `mime` plugin shall use `YAZI_FILE_CMD_PATH` to check mime type, if that environment variable does not exist, it call `file` directly.

This PR benefits mostly Windows users. Currently, Windows users have to add `Git\usr\bin` to PATH. But aside from the `file` command, this also add a lot more commands (`ls`, `less`, `cp`...). Those commands have conflicting names with Powershell or Nushell commands and can cause undesirable behavior.

The solution is to give yazi the path of `file` command via some form of configuration. Considering that `mime` is a lua plugin, an environment variable config is easier to implement.

**Note**: If this PR is accepted, we need to update the doc, but I can't find where it is.

### DIFF
--- a/yazi-plugin/preset/plugins/mime.lua
+++ b/yazi-plugin/preset/plugins/mime.lua
@@ -15,10 +15,10 @@ function M:preload()
 		urls[#urls + 1] = tostring(file.url)
 	end
 
-	local file_cmd_path = os.getenv("YAZI_FILE_CMD_PATH") or "file"
-	local child, code = Command(file_cmd_path):args({ "-bL", "--mime-type" }):args(urls):stdout(Command.PIPED):spawn()
+	local cmd = os.getenv("YAZI_FILE_ONE") or "file"
+	local child, code = Command(cmd):args({ "-bL", "--mime-type" }):args(urls):stdout(Command.PIPED):spawn()
 	if not child then
-		ya.err(string.format("spawn `%s` command returns %s", file_cmd_path, tostring(code)))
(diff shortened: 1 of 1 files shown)


yazi-plugin/preset/plugins/mime.lua
```diff
@@ -15,7 +15,8 @@ function M:preload()
 		urls[#urls + 1] = tostring(file.url)
 	end
 
-	local child, code = Command("file"):args({ "-bL", "--mime-type" }):args(urls):stdout(Command.PIPED):spawn()
+	local file_cmd_path = os.getenv("YAZI_FILE_CMD_PATH") or "file"
+	local child, code = Command(file_cmd_path):args({ "-bL", "--mime-type" }):args(urls):stdout(Command.PIPED):spawn()
 	if not child then
 		ya.err("spawn `file` command returns " .. tostring(code))
```
sxyazi, line 21:
```suggestion
		ya.err("spawn `" .. file_cmd_path .. "` command returns " .. tostring(code))
```

sxyazi: Or using `string.format()` instead of string concatenation?

sxyazi:
Thank you for contributing to this feature!

The docs here https://github.com/yazi-rs/yazi-rs.github.io/blob/main/docs/installation.md, would you like to raise a PR to add it as well?

ndtoan96:
Yes, I'll make a PR there too.

PR #753, opened 2024-02-29
feat: flavor

sxyazi:
Docs here https://yazi-rs.github.io/docs/flavors/overview

### DIFF
--- a/yazi-boot/src/boot.rs
+++ b/yazi-boot/src/boot.rs
@@ -1,4 +1,4 @@
-use std::{ffi::OsString, fs, path::{Path, PathBuf}, process};
+use std::{ffi::OsString, path::{Path, PathBuf}, process};
 
 use clap::Parser;
 use serde::Serialize;
@@ -16,7 +16,6 @@ pub struct Boot {
 	pub config_dir: PathBuf,
 	pub flavor_dir: PathBuf,
 	pub plugin_dir: PathBuf,
-	pub state_dir:  PathBuf,
 }
 
 impl Boot {
@@ -37,7 +36,7 @@ impl Boot {
 
 impl Default for Boot {
 	fn default() -> Self {
-		let config_dir = Xdg::config_dir().unwrap();
+		let config_dir = Xdg::config_dir();
 		let (cwd, file) = Self::parse_entry(ARGS.entry.as_deref());
 
 		let boot = Self {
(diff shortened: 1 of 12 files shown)


ningw42:
The name 'flavor' is kind confusing to me. If we are going to make a 'flavor', would it be better if keymapping is also a part of a flavor? This isn't a feature request :P, just my random thoughts.

Maybe something off the topic, I encountered an unexpected error due to the new flavor feature. I am on [NixOS](https://nixos.org/) and yazi is configured via [home-manager](https://github.com/nix-community/home-manager). To match yazi's active development, I use `xdg.configFile` instead of the options provided by home-manager. I missed `xdg.configFile..recursive` which made `$XDG_CONFIG_DIRS/yazi` a read-only directory, which led to yazi's crash on boot due to [the failure of flavor directory creation](https://github.com/sxyazi/yazi/blob/c240602046e372959312380963d451904f512766/yazi-boot/src/boot.rs#L160C1-L160C89). I corrected my mistake, so yazi works as expected now. I switched to yazi right on the introduction of the plugins directory, I didn't encountered any similar issue (crash when missing the plugins directory) if I recall correctly. If anyone happens to have a read-only `$XDG_CONFIG_DIRS/yazi`, this might be confusing to him/her.

```bash
❯ RUST_BACKTRACE=full yazi
Backtrace (most recent call first):
  File "<unknown>", line 0, in better_panic::print_backtrace_info
  File "<unknown>", line 0, in better_panic::Settings::create_panic_handler::{{closure}}
  File "<unknown>", line 0, in yazi_shared::term::term::Term::goodbye
  File "<unknown>", line 0, in yazi::panic::Panic::install::{{closure}}

The application panicked (crashed).
  Failed to create flavor directory: Os { code: 30, kind: ReadOnlyFilesystem, message: "Read-only file system" }
in yazi-boot/src/boot.rs, line 108
thread: main
```

sxyazi:
Hi, `flavor` is a pre-made theme used to define a color scheme, not including key bindings, see https://yazi-rs.github.io/docs/flavors/overview for more details.

This issue arises from the fact that in newer versions of Yazi, it defaults to creating two directories, `~/.config/yazi/plugins` and `~/.config/yazi/flavors`, to allow users to conveniently install plugins and flavors with just one line of `git clone`, without needing to `mkdir` first and then `git clone`.

Unfortunately, this affects users with a read-only file system, and there is a fix for this, https://github.com/NixOS/nixpkgs/pull/295846, but I'm not familiar with Nix, I cannot review it. Perhaps later I can create an issue to see if there are any Nix-savvy Yazi users who can help review it to prevent more people from encountering this issue.

PR #763, opened 2024-03-02
feat: add `Ctrl-[` as an escape key

hankertrix:
In Vim, Ctrl + [ is an alternative to using the <Esc> key and behaves exactly the same way. This pull request adds that functionality to Yazi.

Also added ^ as a key to move to the beginning of the line, like in Vim. This is a little inaccurate as in Vim, the ^ key moves to the first non-whitespace character in a line, not to the beginning of the line, but I think for the use cases in Yazi, the ^ key being the same as the 0 key shouldn't be much of an issue.

### DIFF
--- a/yazi-config/preset/keymap.toml
+++ b/yazi-config/preset/keymap.toml
@@ -6,6 +6,7 @@
 
 keymap = [
 	{ on = [ "<Esc>" ], run = "escape",             desc = "Exit visual mode, clear selected, or cancel search" },
+	{ on = [ "<C-[>" ], run = "escape",             desc = "Exit visual mode, clear selected, or cancel search" },
 	{ on = [ "q" ],     run = "quit",               desc = "Exit the process" },
 	{ on = [ "Q" ],     run = "quit --no-cwd-file", desc = "Exit the process without writing cwd-file" },
 	{ on = [ "<C-q>" ], run = "close",              desc = "Close the current tab, or quit if it is last tab" },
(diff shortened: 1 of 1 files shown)


sxyazi:
Thanks for your PR!

`Ctrl + [` is not an alternative for the `Esc` in Vim. The reason it can be used as an `Esc` is that in the keyboard protocol, `Ctrl+[` and `Esc` are considered the same key, which is a bug/coincidence in the keyboard protocol. Yazi [uses an extended keyboard protocol](https://sw.kovidgoyal.net/kitty/keyboard-protocol/), so it is capable of recognizing them as two separate keys in some new terminals. Also see https://github.com/helix-editor/helix/issues/6551

However, if this could enhance user experience, I can accept adding it, but I'm sorry I can't accept `^` as it would be a bug since it is inconsistent with Vim's behavior.

hankertrix:
I see. Quite a few Vim users' would find it easier to use Ctrl + [ instead of <Esc>, including myself, so it'll improve the user experience to have it implemented by default.

Got it, I'll make sure to remove the `^` key binding.

I'm assuming that Ctrl + c as an alternative for <Esc> wouldn't be wanted as well since it behaves slightly differently from the <Esc> key in Vim?

sxyazi:
I will add `Ctrl-c` as a breaking change in Yazi v0.3, to replace `Ctrl-q`, as it is more conventional and closer to Ctrl.

Thank you for contributing to this feature, let's merge it now!

hankertrix:
I see, that’s great! Looking forward to that change.

Thanks for merging my pull request!

Issue #774, opened 2024-03-04
Display a warning notification when opening the file fails.


hqwuzhaoyi:
### Please describe the problem you're trying to solve

I found that in the previous version, if mac is not configured with EDITOR, there will be a flash prompt, but it cannot be opened. The new version has default to using vi to open.

But just in case, notify message has been added.

As a newcomer to Rust, I really like yazi and want to contribute to it, but I have encountered some problems.

### Would you be willing to contribute this feature?

- [X] Yes, I'll give it a shot

### Describe the solution you'd like

I will add validation to the `open` method of the `process.rs` module to see if it is not success.

```
			match external::shell(opt) {
				Ok(mut child) => {
					let status = child.wait().await?;
					if !status.success() {
						let message = match status.code() {
							Some(code) => format!("Exited with status code: {code}"),
							None => "Process terminated by signal".to_string(),
						};
						AppProxy::warn("Process failed", message.as_str());
					}
					self.succ(task.id)?;
				}
				Err(e) => {
					self.prog.send(TaskProg::New(task.id, 0))?;
					self.fail(task.id, format!("Failed to spawn process: {e}"))?;
				}
			}
```

However, unable to obtain stdout information now. I am wondering if changing piped to true can successfully get error messages.

```
impl From<&mut ProcessOpOpen> for ShellOpt {
	fn from(op: &mut ProcessOpOpen) -> Self {
		Self {
			cmd:    mem::take(&mut op.cmd),
			args:   mem::take(&mut op.args),
			piped:  false,
			orphan: op.orphan,
		}
	}
}
```

hqwuzhaoyi:
![CleanShot 2024-03-05 at 00 00 42@2x](https://github.com/sxyazi/yazi/assets/44605072/66b45849-e091-4e5d-b72b-94b44699d70b)
![CleanShot 2024-03-05 at 00 01 05@2x](https://github.com/sxyazi/yazi/assets/44605072/01f8dd2c-6df0-4c82-bd20-87f9b9a00240)

sxyazi:
Done in https://github.com/sxyazi/yazi/pull/775, thanks!

PR #775, opened 2024-03-04
feat: send a foreground notification to the user when the process fails to run

hqwuzhaoyi:
#774


sxyazi:

PR #779, opened 2024-03-05
feat: add `<C-p>` and `<C-n>` to the select component for moving the cursor up/down

### LINKED ISSUE #778: Add a few more sensible key bindings

The completion menu key binds should include <kbd>Ctrl</kbd> + <kbd>n</kbd> and <kbd>Ctrl</kbd> + <kbd>p</kbd> to move down and up the completion items respectively. This is the default key bind given in most completion plugins for Vim and Neovim, so I think these key binds will result in a better user experience for those using Vim.

The selection menu should be able to be closed by using the <kbd>q</kbd> key as <kbd>q</kbd> also closes the file manager, so it seems natural to have <kbd>q</kbd> to close any menus in the file manager that doesn't require text input, like the selection menu.

Pull request #779 implements the above key binds.

### Additional context

_No response_

hankertrix:
Particularly, <kbd>Ctrl</kbd> + <kbd>n</kbd> and <kbd>Ctrl</kbd> + <kbd>p</kbd> to move down and up the completion menu respectively.

Closes #778

### DIFF
--- a/yazi-config/preset/keymap.toml
+++ b/yazi-config/preset/keymap.toml
@@ -276,6 +276,9 @@ keymap = [
 	{ on = [ "<Up>" ],   run = "arrow -1", desc = "Move cursor up" },
 	{ on = [ "<Down>" ], run = "arrow 1",  desc = "Move cursor down" },
 
+	{ on = [ "<C-p>" ], run = "arrow -1", desc = "Move cursor up" },
+	{ on = [ "<C-n>" ], run = "arrow 1",  desc = "Move cursor down" },
+
 	{ on = [ "~" ], run = "help", desc = "Open help" }
 ]
 


sxyazi:
Thank you for the PR!

`C-n` and `C-p` look good to me, but I'm not keen on adding `q` for the following reasons:

- We already have `Esc` and `Ctrl-q` (which will be changed to `Ctrl-c` in the next major version), so it doesn't seem necessary to introduce another key.
- It's not always available, which adds to the user's cognitive load. For example, as you mentioned, it can't be used in input insert mode. That's one reason I introduced `C-q` besides `Esc`. I wanted a key that could provide a unified and consistent operating experience, while `Esc` is used to enter normal mode and cannot exit input in insert mode.

hankertrix:
Ah I see, that makes a lot of sense.

I'll remove the commit for having <kbd>q</kbd> close the selection menu.

sxyazi:
All looks good to me now, let me merge it!

PR #785, opened 2024-03-06
feat: add new linemode "mixin"

ListenV:
Better representation of different file types in Unix：

eg：symlink、socket、fifo、dev

Normal files are displayed by `size`


sxyazi:
I'm not too keen on adding it as a builtin linemode, as its complex enough that it limits its usability to specific workflows, and "mixin" is a bit vague to me. I'd prefer something simple and specific, but maybe I can add it to the [Tips page](https://yazi-rs.github.io/docs/tips). Sorry!

Issue #796, opened 2024-03-10
yazi does not respond to `SIGTERM` when running a subprocess


musjj:
### What system are you running Yazi on?

Linux Wayland

### What terminal are you running Yazi in?

kitty v0.32.2

### Did you try the latest main branch to see if the problem got fixed?

Tried, but the problem still happens.

<details>
<summary><h3><code>yazi --debug</code> output</h3></summary>

```sh
Yazi
    yazi 0.2.4 (VERGEN_IDEMPOTENT_OUTPUT 1980-01-01)

Environment
    OS: linux-x86_64 (unix)
    Debug: false

Emulator
    Emulator.via_env: ("xterm-kitty", "")
    Emulator.via_csi: Ok(Unknown([]))
    Emulator.detect: Kitty

Adaptor
    Adaptor.matches: Wayland

tmux
    TMUX: false

Zellij
    ZELLIJ_SESSION_NAME: Some("triangular-yak")

Desktop
    XDG_SESSION_TYPE: Some("wayland")
    WAYLAND_DISPLAY: Some("wayland-1")
    DISPLAY: Some(":0")

Ueberzug
    Version: Err(Os { code: 2, kind: NotFound, message: "No such file or directory" })

WSL
    /proc/sys/fs/binfmt_misc/WSLInterop: false
```
</details>

### Describe the bug

While running a blocking subprocess, yazi will stop responding to any `SIGTERM`. One of the side-effects from this is **really** long shutdown times on systemd-based OSes, if you happen to have yazi running in the background while shutting down.

When shutting down, systemd will try to send `SIGTERM` to politely terminate any running processes. It will wait for a certain time-out length (1m30s for me), before finally giving up with a `SIGKILL` instead.

<details>
<summary><b>From the <code>journalctl</code> logs</b></summary>

```
Mar 03 04:36:57 $HOST systemd[1]: Stopping User Login Management...
Mar 03 04:36:57 $HOST systemd[1]: Stopping Permit User Sessions...
Mar 03 04:36:57 $HOST systemd[1]: Stopping User Manager for UID 1000...
...
Mar 03 04:36:56 $HOST systemd[1]: session-3.scope: Stopping timed out. Killing.
Mar 03 04:36:56 $HOST systemd[1]: session-3.scope: Killing process 1090825 (.yazi-wrapped) with signal SIGKILL.
[...]

Issue #726, opened 2024-02-25
Failed to stop searching after `yazi` is quitted


NeumoNeumo:
### What system are you running Yazi on?

Linux X11

### What terminal are you running Yazi in?

kitty 0.32.2

### Yazi version

yazi 0.2.3 (42cbedb 2024-02-25)

### Did you try the latest main branch to see if the problem got fixed?

Tried, but the problem still

### Describe the bug

When searching using`rg` or `fd` in a large directory, it is OK to take a long time. However, when I want to cancel the process by just quitting `yazi`. The processes of `rg` or `fd` are still running in the background.

Open `yazi`: 
![image](https://github.com/sxyazi/yazi/assets/107997665/750222d6-b8e6-4297-a754-71c245fb9464)

Searching for some file in `yazi`:
![image](https://github.com/sxyazi/yazi/assets/107997665/3a0db7d8-90ce-4869-bc7f-1366cfba1cac)

`fd` is still running without the presence of `yazi`:
![image](https://github.com/sxyazi/yazi/assets/107997665/128f4b8b-c18c-4b18-9d9c-f9290da3e5b1)
![image](https://github.com/sxyazi/yazi/assets/107997665/f9b63e7e-6a42-4df1-98e4-bbab57ee2ce9)
(The server is using NFS so the REAL disk IO is actually displayed in the `Network` instead of `Disk IO` in the above screenshot of `htop`)

### Expected Behavior

After quitting `yazi`, all processes related to it should be killed.

### To Reproduce

1. Find a large directory, for example, the root.
2. Open `yazi` and search for some file, like "gagaga".
3. Before the `yazi` completes the search task, quit it.
4. You would find the searching process is still running in the background.

### Configuration

no config

sxyazi:
How did you exit Yazi, by using `q` to quit it normally, or by forcefully ending it with something like `kill yazi`?

NeumoNeumo:
> Hi, I made a PR trying to fix this issue #812, let me know if it works for you

Thank you. It works! :smile:

PR #797, opened 2024-03-10
fix: respond to the `SIGTERM` signal even when Yazi is in the background and has passed control of the terminal to the spawned process

sxyazi:
Fixes https://github.com/sxyazi/yazi/issues/796

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -37,10 +37,14 @@ body:
       label: "`yazi --debug` output"
       description: Please do a `yazi --debug` and paste the output here.
       value: |
+        <details>
+
         <!-- Paste the output between the backticks below: -->
         ```sh
 
         ```
+
+        </details>
     validations:
       required: true
   - type: textarea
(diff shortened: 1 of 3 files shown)


musjj:
Thanks for the quick PR! But it looks like that this causes the spawned process to linger after yazi terminates. Some other issues:

After terminating yazi, it kind of messes up my shell in a weird way. In the case of helix, the CPU usage of the lingering process spikes up really badly.

It also causes `^C` to not function correctly in the shell, even after the child process is terminated (it normally should create a new empty prompt line).

Maybe a video would be helpful:

https://github.com/sxyazi/yazi/assets/72612857/e4ed8bbb-e951-4a5c-887d-156fd24f2382

But on the other hand, this *DOES* fix the shutdown problem. For me personally, the only time yazi would be receiving `SIGTERM` is when my system is shutting down, so I don't really mind it tbh.

sxyazi:
Thank you for the test. I will create a new PR later to address this issue, which is also related to https://github.com/sxyazi/yazi/issues/726. Let's merge this first.

PR #817, opened 2024-03-14
added docker version

sinawic:
As in the installation page I could not find any packages for ubuntu
I tried to create a basic docker image using archlinux as the base image, so now I can run it in ubuntu and anywhere else that has docker installed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM archlinux
+
+RUN pacman --noconfirm -Sy yazi ffmpegthumbnailer unarchiver jq poppler fd ripgrep fzf zoxide
+
+RUN pacman --noconfirm -Sy vi
+
+ENTRYPOINT yazi
+
+CMD /
--- a/README.md
+++ b/README.md
@@ -29,6 +29,21 @@ https://github.com/sxyazi/yazi/assets/17523360/92ff23fa-0cd5-4f04-b387-894c12265
 - Usage: https://yazi-rs.github.io/docs/installation
 - Features: https://yazi-rs.github.io/features
 
+### Usage via docker image
+
+`docker run -it --rm -v /filesystem:/filesystem sinawic/yazi /filesystem`
+
+by default path will be `/`
+
+#### build yourself
+
+after cloning the repo:
+
+```bash
+cd yazi
+docker build -t yazi .
+```
+
 ## Discussion
 
 - Discord Server (English mainly): https://discord.gg/qfADduSdJu


sxyazi:
Hi, Yazi provides a [snap package](https://github.com/sxyazi/yazi/releases) for Ubuntu users, does it work for you?

sinawic:
hey
thanks for the reply
on this url https://yazi-rs.github.io/docs/installation/ I couldn't find anything for ubuntu
still I tried to dockerize it and it's working just fine
thanks for a really handy tool, great job

sxyazi:
> on this url https://yazi-rs.github.io/docs/installation/ I couldn't find anything for ubuntu

I think Ubuntu users will use [Official binaries](https://yazi-rs.github.io/docs/installation/#official-binaries) or [build Yazi from source code](https://yazi-rs.github.io/docs/installation/#build-from-source), or [install via Cargo](https://yazi-rs.github.io/docs/installation/#cargo).

> still I tried to dockerize it and it's working just fine

Thank you for adding it, but I have some concerns about Dockerizing it:

I want to know if the system clipboard (copy file path by `cc`) and image preview (especially when used with ueberzugpp) work after Dockerizing Yazi? - do they both need to call commands installed on the host machine. If so, I will accept this PR ;)

sinawic:
hey
can u name the exact packages needed to be installed on host machine?
apparently I couldn't also edit files because `vi` was not installed
I could get this solved by adding vi in `Dockerfile`
if you can name the system commands used in `yazi` I can also add it to the `Dockerfile`

sxyazi:
- [wl-clipboard](https://github.com/bugaevc/wl-clipboard) for Wayland system clipboard support
- [xclip](https://github.com/astrand/xclip) / [xsel](https://github.com/kfish/xsel) for X11 system clipboard support
- [pbcopy](https://ss64.com/mac/pbcopy.html) / [pbpaste](https://ss64.com/mac/pbpaste.html) for macOS system clipboard support, comes with macOS by default
- [ueberzugpp](https://github.com/jstkdng/ueberzugpp) for image preview under X11/Wayland

sinawic:
hey
it was new year in here and i couldn't spend much time on this
but [wl-clipboard](https://github.com/bugaevc/wl-clipboard) and [xclip](https://github.com/astrand/xclip) / [xsel](https://github.com/kfish/xsel) are fine to be installed in the arch linux as the base image but the other ones are somehow problematic and needs some r&d, and to be honest im not sure about the end result that it would work or not

sxyazi:
Hi

No, what I mean is not that it can be installed in Docker, but that it can work in Docker - that is, when the user presses `c => c` to copy the path, Yazi will invoke the clipboard command listed above, but it runs in Docker, and I'm not sure if the copied content can be somehow passed to the host machine, as I'm not familiar with how it works. 

If this doesn't work, then I can't accept this PR because it would cause these features to not work.

sxyazi:
I'm going to close this because it still needs some work and hasn't seen any activity recently. @sinawic thanks for the effort!

Issue #766, opened 2024-03-03
Provide information about renamed files for editor integrations


sp3ctum:
### Please describe the problem you're trying to solve

Hi! I am working on a neovim integration at https://github.com/sp3ctum/yazi.nvim.

It opens yazi in a floating terminal and then returns to neovim once yazi has been closed.
It works nicely for simple cases, but sometimes if I rename files or directories, neovim doesn't know about these changes and still has the old file path open.

### Would you be willing to contribute this feature?

- [X] Yes, I'll give it a shot

### Describe the solution you'd like

# Proposition: Write information about the renamings into a file on exit

Currently yazi has the `--chooser-file` option which will contain the chosen files after yazi has exited.

Do you think adding a new option like `--renamed-files` could work? It be similar but for files that were renamed, and could contain a flat list of files that were renamed like so:

```jsonc
// note: the file doesn't necessarily have to be json, but a structured format
// of some kind would eliminate the need for writing extra parsing logic
[
  // example 1: the first renaming operation in the yazi session
  // (the user can do many renaming operations in a single session)
  {
    // this indicates a file was renamed to another file
    "/home/user/project/file_A": "/home/user/project/file_BBBBB",
    // the editor integration can use this information to update the buffer
    // path in the editor
  },

  // example 2: renaming a directory
  {
    "/home/user/project/dir1": "/home/user/project/dir2",
    // the editor integration can update the file paths for all buffers that
    // are open for files in the directory
  },

  // example 3: moving a file into a directory
  {
    "/home/user/project/file_BBBBB": "/home/user/project/dir1/file_BBBBB",
  },
  
  // example 4: moving a directory into another directory
  {
    "/home/user/project/dir1": "/home/user/project/dir2/dir1",
  },
]
```

The editor integration could read this file in after yazi has exited, and update its knowledge about the buffers' files in the same order that yazi did. This should give reasonable guarantees that the editor now tracks the correct file paths.

### Additional context

If you think this approach could work, I would like to give this a shot myself.

I'm not experienced in rust beyond a few advent of code problems that I did last summer out of interest, but this solution seems quite simple because similar functionality already exists 🤞🏻

mikavilpas:
Hmm can you explain how that would work in more detail?

I can read stdout from a subprocess, here's a demo:

https://github.com/sxyazi/yazi/assets/300791/6988477f-5215-4cae-9c53-e5b4194114cb

[...]

Issue #639, opened 2024-02-07
[Feature request] server/client architecture


oredaze:
### Is your feature request related to a problem? Please describe

I love the idea of switching everything I use to rust. So next on the line is switching away from [Lf](https://github.com/gokcehan/lf).

I appear to be missing a killer feature that is present in Lf, which is the ability to have multiple instances of your file manager communicate with each other (copy/paste from one to the other and such).

The idea is that your window manager is responsible for tabbing/arranging multiple windows, etc., not the file manager.

### Will you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

The first run of yazi starts the server/daemon and each subsequent launches of an instance try to connect if server is found.

### Additional context

I am not a programmer, so don't know if my request is reasonable...
Would totally understand if it is too late to implement such a thing.

og900aero:
Ohh noo. Dont be the server/client operation. This mode of operation is also one of the biggest disadvantages of lf, and it cannot use tabs feature.

Abebe123000:
I want this feature too. It is very useful.

rrveex:
> Ohh noo. Dont be the server/client operation. This mode of operation is also one of the biggest disadvantages of lf, and it cannot use tabs feature.

Why is that a disadvantage?

I think it's a matter of getting used to (workflow). When I switched from `ranger` to `lf`, I missed the tabs. Now on `yazi` I miss the "common" or "global" yanked/cut from `lf` and almost don't use tabs at all (yet)

My brain thinks in terms of "where am I now?" and somehow correlates say "a project" (like `yazi` git tree) or "a place" (like `~/.config`) with something both visual and "action to go there" (like `cd` or "switch to workspace 3, look at left side of screen").

Coincidentally,  during my `ranger` times, I didn't use workspaces much, so having several places (== tabs) in the same window was an advantage. The "where" was "tab number" in the single file manager instance, (almost) always on screen.

Now on tiling WMs, the "were" has become a workspace on which I usually have one file manager + some other project related window(s). So I don't need (that many) tabs, but the "global" yanked would be nice.

Doesn't have to be "server/client" though. `lf` actually just keeps the list in `~/.local/share/files`. Something similar is easily doable in `yazi` with a plugin. Just pipe the yank/cut/paste operations through a plugin which does the "accounting" and then `ya.manager_emit(yank/...)`.

sxyazi:
[...]

PR #826, opened 2024-03-18
[WIP] feat: DDS (Data Distribution Service)

sxyazi:
This PR aims to address:

- Closes https://github.com/sxyazi/yazi/issues/766
- Closes https://github.com/sxyazi/yazi/issues/639

By introducing a new DDS (Data Distribution Service), allows communication and data synchronization between multiple Yazi instances through the plugin system.

It also provides a new mechanism to report Yazi events in real-time to external programs (such as Neovim) via `stdout`.

It's somewhat akin to a client-server architecture for `lf` but doesn't require running an additional server process and it's a Lua-based publish-subscribe pattern.

## TODO

- [ ] Windows support
- [ ] Plugin system API


mikavilpas:
I like the idea! 👍🏻 

I wonder if I understand the architecture correctly - here's what I think is happening:
- a global pubsub system is setup and communication between all running yazi instances (or other applications who want to listen in, perhaps) will happen via `/tmp/yazi.sock`, a unix socket (possibly something else for windows)
- when yazi instances perform actions, such as copying files, these are broadcast to the global socket
- actions of interest to editor integrations will, in addition, be written to stdout

Seems like stdout is added as a means of filtering events originating from a specific yazi instance only, as a sort of convenience feature. Or is the motivation performance?

As an example, Neovim's [RPC](https://neovim.io/doc/user/remote.html) has one socket per editor:

```sh
❯ file $NVIM
/var/folders/23/y7gf7hz129l56_83w_g49jl80000gn/T/nvim.mikavilpas/Ngixlj/nvim.74060.0: socket
# neovim's remote documentation says this is a named pipe (which would indicate same system communication, not over the network), but I'm not sure if this is a mistake
```

But then again, neovim supports access over the network IIRC (I think some use it to talk to docker devcontainers from neovim), but I have not tried this out.

sxyazi:
Thank you for summarizing, that's roughly how it works. But `stdout` is only used for communication with external processes and is read-only.

This is the current idea and I'm still exploring if there's a better way and need to do some experimentation at the code level.

Issue #847, opened 2024-03-28
More readable config errors


teto:
### Please describe the problem you're trying to solve

I tried to search for an existing ticket but to my surprise found none.

For instance I spread this from one line
 	`{ mime = "application/octet-stream",              use = [ "extract_zlib", "reveal" ] }`
to 2 lines:
```
{ mime = "application/octet-stream",              
use = [ "extract_zlib", "reveal" ] }
```
and this resulted into
```
yazi

Backtrace omitted. Run with RUST_BACKTRACE=1 to display it.
Run with RUST_BACKTRACE=full to include source snippets.

The application panicked (crashed).
  called `Result::unwrap()` on an `Err` value: Error { inner: Error { inner: TomlError { message: "invalid inline table\nexpected `}`", original: Some("# A TOML linter such as https://taplo.tamasfe.dev/ can use this schema to validate your config.\n# If you encounter any issues, please make an issue at https://github.com/yazi-rs/schemas.\n\"$schema\" = \"https://yazi-rs.github.io/schemas/yazi.json\"\n\n[manager]\nratio          = [ 1, 4, 3 ]\nsort_by        = \"alphabetical\"\nsort_sensitive = false\nsort_reverse   = false\nsort_dir_first = true\n# display information associated with the file on the right side of the file list row.\nlinemode       = \"size\"\nshow_hidden    = true  # toggled with .\nshow_symlink   = true\n# The number of files to keep above and below the cursor when moving through the file list.\nscrolloff      = 5\n\n[preview]\ntab_size        = 2\nmax_width       = 600\nmax_height      = 900\n# cache_dir       = \"\"\nimage_filter    = \"triangle\"\nimage_quality   = 75\nsixel_fraction  = 15\nueberzug_scale  = 1\nueberzug_offset = [ 0, 0, 0, 0 ]\n\n[opener]\nedit = [\n\t{ run = '${EDITOR:=nvim} \"$@\"', desc = \"$EDITOR\", block = true, for = \"unix\" },\n]\nopen = [\n\t{ run = 'xdg-open \"$@\"',                de 
```
It's not very user friendly.

### Would you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

 If the main error could be isolated that would be fantastic, otherwise even mentioning that the crash is not a "real" crash but just a badly formatted config would be helpful. For instance 
```
yazi

[...]

PR #854, opened 2024-03-31
feat: readable toml parsing error

SoloJacobs:
See commit message for the new error message.
For more information see issue GH-847. Closes https://github.com/sxyazi/yazi/issues/847

Because this is my first pull request in yazi, I've tried to keep it as minimal as possible. Here are some things I noticed while working on this issue:

- Only one file is parsed at a time, this means the user does not get information about `keymap.toml`, if `yazi.toml` has an error.
- No unit test was written for the new functionality. I would need some pointers, if this is needed.
- If `yazi.toml` is valid, but contains an unknown key, then the key appears to silently omitted.
  ```
  [manager]
  sortt_by = "natural"
  ```
- If `yazi.toml` could not be read, due to an IO error, then the file is silently omitted.
- It looks like there is some slight shotgun parsing happening in `Preset` (explanation of shotgun parsing: https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/). This may be a personal preference of mine, but I would have checked that `yazi.toml` matches the format specified the by docs *before* merging it with the default values.

### DIFF
--- a/yazi-config/src/lib.rs
+++ b/yazi-config/src/lib.rs
@@ -47,11 +47,11 @@ pub static INPUT: RoCell<popup::Input> = RoCell::new();
 pub static SELECT: RoCell<popup::Select> = RoCell::new();
 pub static WHICH: RoCell<which::Which> = RoCell::new();
 
-pub fn init() {
+pub fn init() -> anyhow::Result<()> {
 	let config_dir = Xdg::config_dir();
-	MERGED_YAZI.init(Preset::yazi(&config_dir));
-	MERGED_KEYMAP.init(Preset::keymap(&config_dir));
-	MERGED_THEME.init(Preset::theme(&config_dir));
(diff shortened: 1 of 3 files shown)


sxyazi:
Thank you for the PR, it's very neat.

Regarding shotgun parsing, yes, currently, the user's configuration is first merged with the default configuration without any checks; however, the parsing is done based on the merged new configuration (user configuration + default configuration) with checks.

Are you meaning doing the second step's checks in the first step as well? Or moving the checks of the second step to the first step?

SoloJacobs:
Ha, you are fast.

I would try to move the checks in the second step to the first step. If you are open to a refactoring commit like that, I will give it a shot. However, it might turn out that such a refactoring improves the code only by little.

sxyazi:
Sure, I'm open to it. I'd be happy to accept a refactoring PR if it can make the code tidier and won’t be much performance impact ;)

PR #880, opened 2024-04-06
feat: add new `move`, `trash`, and `delete` event kinds to DDS

mikavilpas:
Previously, only the rename command sent `rename` events. Now a cut and paste operation also sends events.

Demo:

https://github.com/sxyazi/yazi/assets/300791/31711f7f-dc5f-4977-a8d8-07bed123702b


sxyazi:
Thanks for the PR!

Could you create a new event type `move` for it and add the code to that event?

mikavilpas:
Sure!

mikavilpas:
Okay, there is now a new `move` event instead.

image

I don't know how to test the lua side of things yet. I copied the code from the `rename` events so I guess it might work without any changes 🙂

yazi-core/src/manager/commands/paste.rs
```diff
@@ -21,6 +22,12 @@ impl Manager {
 		if self.yanked.cut {
 			tasks.file_cut(&src, dest, opt.force);
 
+			src.iter().for_each(|source_file| {
```
sxyazi, line 25: I think it should be moved to the scheduler, https://github.com/sxyazi/yazi/blob/357f87aecbe52f1f36efff69af67558f7cf730a2/yazi-scheduler/src/file/file.rs#L28, as moving is not immediately completed (for example, when it involves moving across mount points). Some files may fail to move due to permissions or because the mount point is full, and I think such cases should not be reported as events.

yazi-dds/src/body/mod.rs
```diff
@@ -7,6 +7,7 @@ mod custom;
 mod hey;
 mod hi;
 mod hover;
+mod move_event;
```
sxyazi, line 10:
```suggestion
mod move_;
```

yazi-dds/src/payload.rs
```diff
@@ -82,6 +88,7 @@ impl Display for Payload<'_> {
 			Body::Cd(b) => serde_json::to_string(b),
 			Body::Hover(b) => serde_json::to_string(b),
 			Body::Rename(b) => serde_json::to_string(b),
+			Body::Move(b) => serde_json::to_string(b),
 			Body::Bulk(b) => serde_json::to_string(b),
 			Body::Yank(b) => serde_json::to_string(b),
```
sxyazi, line 93:
```suggestion
			Body::Bulk(b) => serde_json::to_string(b),
			Body::Yank(b) => serde_json::to_string(b),
			Body::Move(b) => serde_json::to_string(b),
```

yazi-core/src/manager/commands/paste.rs
mikavilpas: I like this idea. I tried implementing it but ran into some issues and could use some help.

# Issue 1
In the scheduler I don't seem to have access to the current tab though. The tab is present in the rename event, and I think it would need some refactoring to get access to the current tab.

# Issue 2
Publishing events from the scheduler doesn't cause the events to be visible for me. I pushed the commit 410bf4a8852ccedafb5f9f4b05d95b717d8629dd as a work-in-progress commit to show what I tried.

I could not figure this out. Any ideas on what could be done?

sxyazi:
Thanks for the work you've done so far; I'll finish the remaining part.

mikavilpas:
Thanks, I appreciate that!

I also tried to add support for a new `deleted` event which I could use to detect files that were deleted in yazi in neovim.

I kinda ran into the same issue - it seemed like it's possible to request the scheduler to delete files, but I couldn't see if it's possible to know if the request has been successfully completed or not.

Maybe this is a feature that's not implemented yet? It could also be that I missed it.

sxyazi:
Hi, I think I've completed the required work. I tested it by using `yazi --local-events=move,delete` and got the expected output:

CleanShot 2024-04-10 at 08 57 44

mikavilpas:
Yeah it seems to work! As a suggestion, I also added sending a `delete` event when a file is trashed as that seemed to not emit anything for me.

yazi-scheduler/src/scheduler.rs
```diff
@@ -174,7 +174,8 @@ impl Scheduler {
 		let file = self.file.clone();
 		_ = self.micro.try_send(
 			async move {
-				file.trash(FileOpTrash { id, target, length: 0 }).await.ok();
+				file.trash(FileOpTrash { id, target: target.clone(), length: 0 }).await.ok();
+				Pump::push_delete(target);
```
sxyazi, line 178: I don't really like adding `trash` to `delete` because they're two different concepts, I think it deserves a new `trash` event kind.

mikavilpas: Ok, I'll add a new event kind. It should be much simpler after the latest changes.

mikavilpas: Ok it's done now and seems to work 
image

sxyazi:
Thanks, let me do some minor refactors

sxyazi:
All looks good to me now, let me merge it now. 

Issue #782, opened 2024-03-05
Path completion improvement on Windows


ndtoan96:
### Please describe the problem you're trying to solve

I see several things can be improved for path completion (`g <space>`) on Windows:
1. Use `/` separator to trigger folder completion.
2. Smart case matching (i.e typing "prog" can matches "Program Files").
3. On Windows, somehow pressing `tab` for completion turns the `\` to `/`. For example, I'm at `.github\` and the completion popup shows `ISSUE_TEMPLATES\` and `workflows\`. If I select `workflow\` and press `tab` then the input becomes `.github/workflows\`

### Would you be willing to contribute this feature?

- [X] Yes, I'll give it a shot

### Describe the solution you'd like

As described above.

### Additional context

I was trying to implement the features but can't quite understand the code (e.g I don't know what is `word` and `ticket`, and what the `Cmd` is supposed to do). It would help if you can briefly explain the `completion` module.

sxyazi:
Are you saying the two parameters of `CompletionProxy.trigger()`?

https://github.com/sxyazi/yazi/blob/e3505b717c53a27b3deae20a911ef1bb8eecd5b2/yazi-proxy/src/completion.rs#L12-L17

- `word` is used to trigger the completion of this word, for example `"C:\"`.
- `ticket` is an incrementing unique identifier for this completion, since the completion is asynchronous, the completion result will be discarded if the identifier has changed when confirming the completion, indicating that the completion is no longer valid.

This method is an asynchronous proxy for `Completion.trigger()`:

https://github.com/sxyazi/yazi/blob/e3505b717c53a27b3deae20a911ef1bb8eecd5b2/yazi-core/src/completion/commands/trigger.rs#L23

sxyazi:
Hi, are you still interested in implementing this feature?

ndtoan96:
> Hi, are you still interested in implementing this feature?

Sorry, I almost forgot about this. It was put pretty deep in my backlog, hahaha.

I've implemented 2 of the 3 issues. I'll make another PR for smart case completion later.

sxyazi:
Done in https://github.com/sxyazi/yazi/pull/909 and https://github.com/sxyazi/yazi/pull/910, thank you so much for contributing to this awesome feature!

PR #909, opened 2024-04-14
feat: trigger path completion with both `/` and `\` on Windows

ndtoan96:
Fix some issues described in #782:
- Trigger completion with both `/` and `\`.
- Normalized path with `MAIN_SEPARATOR` when pressing `tab` for completion.

I've implemented them in Windows specific code. Because Linux apparently allows `\` in file name.


ndtoan96:
Demo:

https://github.com/sxyazi/yazi/assets/33489972/9230a762-a37b-427f-a3ce-b2822ce541f7

sxyazi:
Thanks for the PR! I've made some simplifications

PR #910, opened 2024-04-14
feat: Smart case completion

ndtoan96:
Implement smart case for path completion (issue #782)

https://github.com/sxyazi/yazi/assets/33489972/e9c503e1-0eea-4407-b379-b6c5f2882640


sxyazi:
Thanks!

PR #915, opened 2024-04-15
fix: add rustup to nix shell for building yazi

colbs255:
By adding rustup to the nix flake, I was able to build yazi with these steps:

``` bash
nix develop
scripts/build.sh yazi-x86_64-unknown-linux-gnu
```

### DIFF
--- a/nix/shell.nix
+++ b/nix/shell.nix
@@ -16,6 +16,7 @@ pkgs.mkShell {
     ripgrep
     fzf
     zoxide
+    rustup
   ];
 
   buildInputs = with pkgs;


sxyazi:
Thanks for the PR!

It seems there's already `rustToolchain`, why do we need `rustup`?

https://github.com/sxyazi/yazi/blob/64c5e854578b04907717b0f6b3e5e5246d6fd236/nix/shell.nix#L5

colbs255:
Ahh, did not see that!

What prompted this is building with scripts/build.sh. The build failed because it requires rustup. When I added this to the nix file, it built. 

But I agree, rustToolchain is probably sufficient for development.

sxyazi:
That script is expected to be used in CI, not used for users directly. If you want to build Yazi, please follow the user documentation https://yazi-rs.github.io/docs/installation

PR #920, opened 2024-04-16
fix: allow opening interactively with the --chosen-file flag

mikavilpas:
This commit fixes a bug where the user was unable to open a file interactively using when the `--chosen-file` flag was used.

This is provided by default in yazi for the `O` (capital o) keybinding.

(keymap.toml)

```toml
[manager]
keymap = [
  { on = [ "O", ], run = "open --interactive", desc = "Open the selected files interactively" },
]
```

This came to me today as a question from a user in https://github.com/mikavilpas/yazi.nvim/issues/30


sxyazi:
Thanks!

PR #925, opened 2024-04-18
fix: prevent pasting a directory into itself

### LINKED ISSUE #894: Protection from copying a directory into itself

240410_16h23m30s_screenshot

cp has protection against copying a directory into itself. 

240410_16h23m56s_screenshot

This is not present in yazi resulting in an operation running forever. 
Of course you could just quit yazi, so it is not an urgent feature, just a nice addition to yazi. 

A warning message when you are about to copy a directory into itself. 

### Additional context

_No response_

Rolv-Apneseth:
Closes #894 

Just the first approach I could think of but let me know if you had something else in mind.

Edit: Figured out a much easier way - did not know about `path.startswith` haha


sxyazi:
Thanks for the fix!

PR #926, opened 2024-04-18
feat: preserve files' modified at timestamp while copying

### LINKED ISSUE #863: Preserve timestamp

`y` and `x` changes file attribute like timestamp after `p`.

All actions in yazi should always keep all original attribute.

### Additional context

_No response_

Rolv-Apneseth:
Closes #863 

I couldn't find how to do this without a dependency, but if you like I could try copy how it's done in `filetime` (think it uses unsafe code though), or if you can think of another way that it can be done, do let me know.


sxyazi:
Thanks!

Rolv-Apneseth:
No problem

PR #932, opened 2024-04-20
feat: owner user and group linemode

AidanV:
Added m => o linemode keymap that displays the user and group of the owner of a file or directory. 

Adds dependency in yazi-shared
users = "0.11.0"
https://crates.io/crates/users

### DIFF
--- a/yazi-plugin/preset/components/folder.lua
+++ b/yazi-plugin/preset/components/folder.lua
@@ -22,11 +22,11 @@ function Folder:linemode(area, files)
 		elseif mode == "permissions" then
 			spans[#spans + 1] = ui.Span(f.cha:permissions() or "")
 		elseif mode == "owner" then
-      local username = ya.user_name(f.cha.uid)
-      local groupname = ya.group_name(f.cha.gid)
-      if username and groupname then
-        spans[#spans + 1] = ui.Span(username .. ":" .. groupname)
-      end
+			local username = ya.user_name(f.cha.uid)
+			local groupname = ya.group_name(f.cha.gid)
+			if username and groupname then
+				spans[#spans + 1] = ui.Span(username .. ":" .. groupname)
+			end
 		end
 
 		spans[#spans + 1] = ui.Span(" ")
--- a/yazi-plugin/preset/components/folder.lua
+++ b/yazi-plugin/preset/components/folder.lua
@@ -22,7 +22,11 @@ function Folder:linemode(area, files)
 		elseif mode == "permissions" then
 			spans[#spans + 1] = ui.Span(f.cha:permissions() or "")
 		elseif mode == "owner" then
-			spans[#spans + 1] = ui.Span(f.cha:owner() or "")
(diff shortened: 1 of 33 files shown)


yazi-plugin/preset/components/folder.lua
```diff
@@ -21,6 +21,8 @@ function Folder:linemode(area, files)
 			spans[#spans + 1] = ui.Span(time and os.date("%y-%m-%d %H:%M", time // 1) or "")
 		elseif mode == "permissions" then
 			spans[#spans + 1] = ui.Span(f.cha:permissions() or "")
+		elseif mode == "owner" then
+			spans[#spans + 1] = ui.Span(f.cha:owner() or "")
```
sxyazi, line 25: Please use the existing `user_name()` instead, https://yazi-rs.github.io/docs/plugins/utils#ya.user_name

AidanV:
Made suggested change

yazi-plugin/preset/components/folder.lua
```diff
@@ -21,6 +21,12 @@ function Folder:linemode(area, files)
 			spans[#spans + 1] = ui.Span(time and os.date("%y-%m-%d %H:%M", time // 1) or "")
 		elseif mode == "permissions" then
 			spans[#spans + 1] = ui.Span(f.cha:permissions() or "")
+		elseif mode == "owner" then
+			local username = ya.user_name(f.cha.uid)
```
sxyazi, line 25: `ya.user_name`, `ya.group_name`, `cha.uid`, and `cha.gid` are Unix-specific attributes/methods, they will break the UI rendering on the Windows platform.

Perhaps we can expose them to the Windows platform, but just return a `nil` value? Or add this code snippet to the [Tips page](https://yazi-rs.github.io/docs/tips) so Unix users can copy it from there? Curious your thoughts!

AidanV: I think it is pretty important on Unix to be able to see the user and group of files by default. Additionally this functionality somewhat [already exists](https://yazi-rs.github.io/docs/tips/#user-group-in-status) through the status bar.

It would be nice if this functionality could work on Windows as well. It looks doable [here](https://users.rust-lang.org/t/how-can-i-get-the-owner-of-a-file-on-windows/47563), but more complicated than it should be.

If we expose these methods to the Windows platform we should remove the ability to see the ownership key-bind in the help page. It would be confusing for a Windows user to see the option and have it do nothing.

I am happy to implement either safely hiding it on Windows or having the ownership functionality on Windows, up to you.

sxyazi: Great, please implement the ownership functionality on Windows!

sxyazi:
I'm going to close this because it still needs some work and hasn't seen any activity recently. @AidanV thanks for the effort!

AidanV:
Sounds good, I do not have access to a Windows computer at the moment. I will pick it back up when I do.

PR #937, opened 2024-04-21
feat: add new `bulk` event kind to DDS

mikavilpas:
This makes bulk renaming send one event per rename, exactly how the existing rename events are sent.

https://github.com/sxyazi/yazi/assets/300791/b43ef535-a38d-46cc-9055-241ef4bc12d1

I have a couple of ideas related to bulk renaming that I want to explore in neovim. I think I'll open an issue for discussion separately.


sxyazi:
Hi, thanks for your PR. 

Could you create a new kind `bulk` for the bulk rename?

mikavilpas:
Ok, now that is done. 

Looks like there already was a `bulk` event, but it seemed to be unused. I presume it might have been intended for this, so I just used that 🙂

sxyazi:
Thanks!

sxyazi (approved):

PR #943, opened 2024-04-23
feat: allow building with the system Lua

Kladki:
Still defaults to vendored lua, but that can be overridden by using `--no-default-features`

### DIFF
--- a/yazi-fm/Cargo.toml
+++ b/yazi-fm/Cargo.toml
@@ -8,6 +8,10 @@ description = "Yazi File Manager"
 homepage    = "https://yazi-rs.github.io"
 repository  = "https://github.com/sxyazi/yazi"
 
+[features]
+default      = [ "vendored-lua" ]
+vendored-lua = [ "mlua/vendored" ]
+
 [dependencies]
 yazi-adaptor = { path = "../yazi-adaptor", version = "0.2.5" }
 yazi-boot    = { path = "../yazi-boot", version = "0.2.5" }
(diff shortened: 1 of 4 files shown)


sxyazi:
Thanks for the PR! 

Could you add `default = ["vendored-lua"]` to each crate instead of just adding it to `yazi-fm`?

I think this would be more robust, as for example `yazi-dds` is not only depended on by `yazi-fm` but also by `yazi-cli`, `yazi-core`, `yazi-scheduler`, etc.

Kladki:
Ah ok. That would require building with `--workspace` to be able to build with system lua, but I can see that the robustness may be worth it.

yazi-fm/Cargo.toml
```diff
@@ -8,6 +8,10 @@ description = "Yazi File Manager"
 homepage    = "https://yazi-rs.github.io"
 repository  = "https://github.com/sxyazi/yazi"
 
+[features]
+default      = [ "vendored-lua" ]
+vendored-lua = [ "mlua/vendored", "yazi-dds/vendored-lua", "yazi-plugin/vendored-lua", "yazi-proxy/vendored-lua" ]
```
sxyazi, line 13:
```suggestion
vendored-lua = [ "mlua/vendored" ]
```

sxyazi (approved):

Kladki:
Sorry about this, but it seems the method you suggested doesn't work. I'll follow up later to see if I can get it working in a way you like.

24 Apr 2024 11:50:08 三咲雅 · Misaki Masa ***@***.***>:

Issue #963, opened 2024-04-27
Add some common `[filetype]` rules to the default theme configuration


sxyazi:
### Please describe the problem you're trying to solve

Yazi has added support for the [new `is` rule](https://yazi-rs.github.io/docs/configuration/theme#filetype), which is used to match specific file types (such as executable files or orphaned link files). I think we should add them by default because they are very useful for many users.

Perhaps we can refer to other popular terminal file managers, such as ranger, lf, nnn, etc., to optimize the existing `[filetype]`. In other words, I am open to changing the existing `[filetype]` as long as it can bring a better user experience. 

This is a complex aesthetic work, but I don't have much confidence in my own aesthetics, so I hope to see someone interested in trying it out :)

### Would you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

.

Brixy:
This is what I use. Would this be a start? I came to yazi from nnn, hence the colors.

```
	# Empty files
	{ mime = "inode/x-empty", fg = "red" },

	# Is rules
	{ name = "*",  is = "orphan", fg = "red" },
	{ name = "*",  is = "link",   fg = "cyan" },
	{ name = "*/", is = "link",   fg = "cyan" },
	{ name = "*",  is = "exec",   fg = "green" },

```

sxyazi:
Looks good! Would you like to raise a PR for it?

Brixy:
With pleasure! I’m on the road for a few hours now, but I’ll get back to it today.

sxyazi:
Done in https://github.com/sxyazi/yazi/pull/966, thanks!

PR #966, opened 2024-04-27
feat: add more rules to `[filetype]` and `[icon]`

Brixy:
This adds rules for empty files, links and executables.

Coloring is based on [eza's default theme](https://github.com/eza-community/eza/blob/main/src/theme/default_theme.rs). This also seems to be a default across  terminal file managers.

### DIFF
--- a/yazi-config/preset/theme.toml
+++ b/yazi-config/preset/theme.toml
@@ -170,10 +170,10 @@ icon_error = ""
 
 rules = [
 	# Images
-	{ mime = "image/*", fg = "cyan" },
+	{ mime = "image/*", fg = "yellow" },
 
 	# Media
-	{ mime = "{audio,video}/*", fg = "yellow" },
+	{ mime = "{audio,video}/*", fg = "yellow", bold = true },
 
 	# Archives
 	{ mime = "application/*zip", fg = "magenta" },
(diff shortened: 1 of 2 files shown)


Brixy:
Closes #963

yazi-config/preset/theme.toml
```diff
@@ -182,6 +182,15 @@ rules = [
 	# Documents
 	{ mime = "application/{pdf,doc,rtf,vnd.*}", fg = "green" },
 
+	# Empty files
+	{ mime = "inode/x-empty", fg = "red" },
+
+	# Is-rules
+	{ name = "*" , is = "orphan", fg = "red" },
+	{ name = "*" , is = "link"  , fg = "cyan" },
```
sxyazi, line 190: "cyan" and "green" are already being used by the images and documents, perhaps we can choose other colors? Or change the images and documents to use different colors?

I think using a strike-through for "orphan" instead of a color might also be a consideration, as this would give us more colors to work with (i.e., red in this case).

Brixy:
OK. Changed.

- Cyan is used for links which is common.
- Red is just used once for empty files.
- Images and audios/videos use the same color, but audio and video filetypes are boldened. (This is approximately what eza does, yet with purple instead of yazi’s yellow.)

I also tested underlining links, but it was confusing as underlining is used in the preview windows.

Brixy:
Excellent! Thank you.

PR #968, opened 2024-04-27
fix: improve accessibility by avoiding hex color code for white

Brixy:
Visibility of all hard coded colors depend on the terminal emulators' background-color. Accessibility can be addressed by using one’s own theme.

Yet, IMHO accessibility can be improved by removing hard hex color codes for white (and black).


sxyazi:
Ah fair enough - merged!

PR #969, opened 2024-04-28
feat: allow building with the system Lua

Kladki:
As per https://github.com/sxyazi/yazi/pull/943#issuecomment-2075013872

I tried to use `default-features = false` in the deps of `yazi-fm` and `yazi-cli` and have `vendored-lua` to be default in the dependencies, but when I did that it would just build with vendored lua every time. This is the only method that I could get to actually work, sorry if it's not ideal. :(

### DIFF
--- a/yazi-core/Cargo.toml
+++ b/yazi-core/Cargo.toml
@@ -12,9 +12,9 @@ repository  = "https://github.com/sxyazi/yazi"
 yazi-adaptor   = { path = "../yazi-adaptor", version = "0.2.5" }
 yazi-boot      = { path = "../yazi-boot", version = "0.2.5" }
 yazi-config    = { path = "../yazi-config", version = "0.2.5" }
-yazi-dds       = { path = "../yazi-dds", version = "0.2.5" }
-yazi-plugin    = { path = "../yazi-plugin", version = "0.2.5" }
-yazi-proxy     = { path = "../yazi-proxy", version = "0.2.5" }
+yazi-dds       = { path = "../yazi-dds", version = "0.2.5", default-features = false }
+yazi-plugin    = { path = "../yazi-plugin", version = "0.2.5", default-features = false }
+yazi-proxy     = { path = "../yazi-proxy", version = "0.2.5", default-features = false }
 yazi-scheduler = { path = "../yazi-scheduler", version = "0.2.5" }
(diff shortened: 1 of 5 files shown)


sxyazi:
Is it possible to still keep each crate's `default = [ "vendored-lua" ]`, but prevent them from being referenced in e.g. `yazi-fm` and `yazi-cli` using something like `yazi-dds = { path = "../yazi-dds", version = "0.2.5", default-features = false }`?

Kladki:
I can try, but that would likely require setting `default-features = false` in every `Cargo.toml` as some of the other yazi dependencies depend on other ones which use the ones which depend on `mlua`.

Kladki:
Done.

yazi-core/Cargo.toml
```diff
@@ -12,9 +12,9 @@ repository  = "https://github.com/sxyazi/yazi"
 yazi-adaptor   = { path = "../yazi-adaptor", version = "0.2.5" }
 yazi-boot      = { path = "../yazi-boot", version = "0.2.5" }
 yazi-config    = { path = "../yazi-config", version = "0.2.5" }
-yazi-dds       = { path = "../yazi-dds", version = "0.2.5" }
-yazi-plugin    = { path = "../yazi-plugin", version = "0.2.5" }
-yazi-proxy     = { path = "../yazi-proxy", version = "0.2.5" }
+yazi-dds       = { path = "../yazi-dds", version = "0.2.5", default-features = false }
```
sxyazi, line 15: Why do we need to add `default-features = false` to `yazi-core`? `yazi-core` is not the entry point of the program, is this necessary?

Kladki: Yes, because that dependency would otherwise expect to have those features, and we cannot override that without making enabling that feature not the default.

Kladki: `cargo tree` seems unreliable for these tests, so I just `eza -l target/release` and compare the sizes. Without this there is no size difference between the default and `--no-default-features`, but with this there is a difference.

sxyazi: What command did you use for building?

Kladki: `cargo build` with and without `--release` and with and without `--no-default-features`

Issue #955, opened 2024-04-26
Redefining default keybinds in submap displays duplicates


SolitudeSF:
### What system are you running Yazi on?

Linux X11

### What terminal are you running Yazi in?

kitty

### Did you try the latest code to see if this problem got fixed?

Tried, but the problem still

### `yazi --debug` output

<details>
<summary>Details</summary>

Yazi
    Version: 0.2.5 (c65a14c 2024-04-23)
    OS: linux-x86_64 (unix)
    Debug: false

Emulator
    Emulator.via_env: ("xterm-kitty", "")
    Emulator.via_csi: Ok(Kitty)
    Emulator.detect: Kitty

Adaptor
    Adaptor.matches: Kitty

Desktop
    XDG_SESSION_TYPE: Some("tty")
    WAYLAND_DISPLAY: None
    DISPLAY: Some(":1")

SSH
    shared.in_ssh_connection: false

WSL
    /proc/sys/fs/binfmt_misc/WSLInterop: false

Variables
    SHELL: Some("/bedrock/cross/bin/bash")
    EDITOR: Some("kak")
    ZELLIJ_SESSION_NAME: None
    YAZI_FILE_ONE: None
    YAZI_CONFIG_HOME: None

file(1)
    Version: Ok(Output { status: ExitStatus(unix_wait_status(0)), stdout: "file-5.45\nmagic file from /usr/share/misc/magic\n", stderr: "" })

Text Opener
    default: Some(Opener { run: "${EDITOR:=vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block: Some(Opener { run: "${EDITOR:=vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })

tmux
    TMUX: false

Ueberzug++
    Version: Err(Os { code: 2, kind: NotFound, message: "No such file or directory" })

</details>

### Describe the bug

![image](https://github.com/sxyazi/yazi/assets/10168645/078c8be4-3aed-44c0-8154-3ff6367a9055)

### Expected Behavior

No duplicates displayed

### To Reproduce

add duplicate submap keybinds via prepend_keymap

### Configuration

```toml
[manager]
prepend_keymap = [
  { on = [ "g", "d" ], run = "plugin shellcd --args='xdg-user-dir DOWNLOAD'", desc = "Go to downloads" },
[...]

PR #975, opened 2024-04-29
fix: displaying duplicate command suggestions

mikavilpas:
Fixes https://github.com/sxyazi/yazi/issues/955

Also added a test for this bug as a suggestion.

Let me know how you would prefer testing to be done in this project!


sxyazi:
Thanks for the PR!

I don't think this test is necessary because `KEYMAP.init` can [only be called once](https://github.com/sxyazi/yazi/blob/3a091553283de223dc664a012ae5cd8e1c35933e/yazi-shared/src/ro_cell.rs#L16), and it's difficult to assume whether it will appear in other tests. 

It contains unsafe code, so currently, it's restricted to only being used in the `init` method of each crate (like this https://github.com/sxyazi/yazi/blob/3a091553283de223dc664a012ae5cd8e1c35933e/yazi-adaptor/src/lib.rs#L33) and I don't want to spread unsafe code into tests, so could you remove the test and only modify the minimum code?

mikavilpas:
I think that makes sense. I actually noticed there were issues trying to add another test, probably because the `RoCell` was already allocated, making for weird errors 😄 

The test has now been removed.

mikavilpas:
Yes, duplicates are removed every time.

@sxyazi I just noticed this doesn't yet preserve the order of the suggestions, so the suggestions can appear in a random order now. I'll fix this soon.

mikavilpas:
Ok, order is now preserved 🙂

sxyazi:
Thanks!

mikavilpas:
I like your refactoring, nicely done 👍🏻

PR #980, opened 2024-04-30
Add `ya.clipboard()` lua api

### LINKED ISSUE #964: Add a new `ya.clipboard()` Lua API

Yazi has integrated clipboard functionality, it would be great if users could access the clipboard in a unified and simple way through the plugin API. 

1. Add a new `ya.clipboard()` method in https://github.com/sxyazi/yazi/blob/681612f9764d45f61bbac622c57e0a48342b05c1/yazi-plugin/src/utils/text.rs#L9
2. Call [`CLIPBOARD.get()`](https://github.com/sxyazi/yazi/blob/681612f9764d45f61bbac622c57e0a48342b05c1/yazi-core/src/clipboard.rs#L15) / [`CLIPBOARD.set()`](https://github.com/sxyazi/yazi/blob/681612f9764d45f61bbac622c57e0a48342b05c1/yazi-core/src/clipboard.rs#L56) in that method to mutate the clipboard.

The `ya.clipboard()` should be used as:

```lua
-- Get clipboard content
local content = ya.clipboard()  

-- Set clipboard content
ya.clipboard("new content") 
```

### Additional context

_No response_

mikavilpas:
Fixes #964 

It allows accessing the clipboard content as well as setting it.

```lua
-- get contents of the clipboard
local content = ya.clipboard()

-- set contents of the clipboard
ya.clipboard("new clipboard content")
```

Note: I moved the clipboard to a different crate because of circular dependency issues. Here's the commit message:

```
refactor: move CLIPBOARD from yazi-core -> yazi-shared

This needs to be done because we want to make it possible to use the
clipboard in lua plugins.

Before this change, the clipboard was only available in the yazi-core
crate, which depends on yazi-plugin.

The lua plugins are configured in the yazi-plugin crate, but it can't
use anything from yazi-core, because it would create a circular
dependency.
```

---

I should still look into testing this with the lua plugin api, so I'll keep this as a draft for now.


sxyazi:
Thank you! 

Please move the `clipboard.rs` to the `yazi-plugin` crate, as only `yazi-core` and `yazi-plugin` need it and `yazi-plugin` is already a dependency of `yazi-core`.

mikavilpas:
Ok, that is now done. I'll try to write a test plugin to verify this works in lua, next.

mikavilpas:
I was able to test setting the clipboard and that works fine. I had trouble testing getting the clipboard contents, however.

I think I don't have a deep enough understanding of yazi plugins to solve this yet. Would you have the time to take a look?

```lua
-- /Users/mikavilpas/.config/yazi/plugins/clipboard-test.yazi/init.lua
return {
	entry = function(_, args)
		-- test setting the clipboard
		--
		local input = ya.input({
			title = "give me some input",
			position = { "top-center", y = 3, w = 40 },
		})

		ya.clipboard(input)

		-- test getting the clipboard
		--
		-- TODO this does not work as expected
		ya.dbg("testing clipboard")
		local contents = ya.clipboard()
		ya.dbg("The clipboard contents are: " .. contents)
	end,
}
```

sxyazi:
Just tested it, and it works great! 👍🏻 

(Since it's an async function, so it needs to be ran in an async context, i.e. `plugin` command without `--sync` flag). 

I think we can merge it now.

mikavilpas:
Oh, I see. Do you think this case could be detected, and a warning could be logged, for example?

mikavilpas:
Ok, gotcha! I don't remember if I saw it in the log file or not, but I will keep my eyes on that in the future 🙂

PR #986, opened 2024-05-01
feat(lua): add basic LuaCATS types for internal lua code

mikavilpas:
This commit adds rudimentary types for the internal lua code. This does not have a big effect on its own, but it can be a part of a bigger effort to add types to the codebase.

The types are only usable within yazi itself and are not published for plugin developers to use (yet).

I only added types for the things that are defined in the lua side of the codebase. This means most of the useful typing is still missing. I want to add that separately, as it would be difficult to keep the types in sync with the actual codebase.

The plan:
- [x] add lua types for lua code
- [ ] add lua types for rust code with a new tool
- [ ] export both as LuaCATS Language Server Addons (meta files)
- [ ] finally, publish the types for plugin developers to use 

See https://github.com/sxyazi/yazi/discussions/981 for more discussion.

### DIFF
--- a/yazi-plugin/preset/components/current.lua
+++ b/yazi-plugin/preset/components/current.lua
@@ -1,7 +1,11 @@
+---@class yazi.Current
+---@field area? unknown
 Current = {
 	area = ui.Rect.default,
 }
 
+---@param area unknown
+---@return table
 function Current:empty(area)
 	local folder = Folder:by_kind(Folder.CURRENT)
 
@@ -17,6 +21,8 @@ function Current:empty(area)
 	}
 end
 
+---@param area unknown
+---@return table
 function Current:render(area)
 	self.area = area
 
--- a/yazi-plugin/preset/components/file.lua
+++ b/yazi-plugin/preset/components/file.lua
@@ -1,5 +1,8 @@
+---@class yazi.File
 File = {}
 
+---@param file unknown
+---@return table
 function File:icon(file)
 	local icon = file:icon()
 	if not icon then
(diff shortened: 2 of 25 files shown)


sxyazi:
Thanks. I think it's better to extract them into a separate file, similar to TypeScript's [`.d.ts` file](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html) for dedicated type definitions. I'm not sure if Lua can do this.

mikavilpas:
I agree, and it's probably the easiest way to accomplish the goal of adding types to yazi. It's actually what I would like to do eventually, but I see I did not explain this very well:

1. (this PR) generate types for `.lua` files
    - have them be next to the code so that they are very easy to keep up to date
2. generate LuaCATS types from rust code and put them into a file
3. combine both using some tool that extracts the lua types into a separate file

However, for defining the types I think it would be the easiest to define them where they are used. I am a bit worried that by defining the types separately, 
- keeping the lua types up to date will be difficult
- module private helper functions and variables cannot be typed at all

Is your motivation performance?

bvr-yr:
oh, I suppose that's it? https://github.com/sxyazi/yazi/discussions/934

mikavilpas:
@bvr-yr yes, that is exactly what I am after. Although currently I'm just getting some discussion started on a way this could be done. This PR could be part of a larger effort to do it, so this will alone not solve everything 🙂

bvr-yr:
nice then! cause currently I just ignore them in `.luarc.json`
as I can understand, final goal is to expose all stuff, not only lua presets, like `ya.manager_emit` and etc for example?

mikavilpas:
Yeah, that's exactly it. Ideally everything would be fully typed.

Issue #983, opened 2024-05-01
Add `--force-window` flag to mpv on `play` opener in the preset configuration


fmorroni:
### Please describe the problem you're trying to solve

With the default configuration, audio files that have no cover art will open in the background with no way to interact with them.

### Would you be willing to contribute this feature?

- [X] Yes, I'll give it a shot

### Describe the solution you'd like

Audio files should always open the mpv window so we can interact with it. This should be fixed by just adding the `--force-window` flag to the opener command in `yazi.toml`.
```toml
play = [
	{ run = 'mpv --force-window "$@"', orphan = true, for = "unix" },
]
```

sxyazi:
Sounds good to me, please raise a PR for it

PR #998, opened 2024-05-03
feat: add `--force-window` option to mpv

GOWxx:
close https://github.com/sxyazi/yazi/issues/983

### DIFF
--- a/yazi-config/preset/yazi.toml
+++ b/yazi-config/preset/yazi.toml
@@ -46,8 +46,8 @@ extract = [
 	{ run = 'unar "%1"', desc = "Extract here", for = "windows" },
 ]
 play = [
-	{ run = 'mpv "$@"', orphan = true, for = "unix" },
-	{ run = 'mpv "%1"', orphan = true, for = "windows" },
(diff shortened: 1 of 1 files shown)


yazi-config/preset/yazi.toml
```diff
@@ -46,8 +46,8 @@ extract = [
 	{ run = 'unar "%1"', desc = "Extract here", for = "windows" },
 ]
 play = [
-	{ run = 'mpv "$@"', orphan = true, for = "unix" },
-	{ run = 'mpv "%1"', orphan = true, for = "windows" },
+	{ run = 'mpv --force-window "$@"', orphan = true, for = "unix" },
+	{ run = 'mpv --force-window  "%1"', orphan = true, for = "windows" },
```
sxyazi, line 50:
```suggestion
	{ run = 'mpv --force-window "%1"', orphan = true, for = "windows" },
```

GOWxx:
image

After configuring like this, the result is a pop-up window. Is this expected?

I think at least it's better than not using `--force-window`, as it provides control.

sxyazi:
Yup this can control it more conveniently. Thanks for the PR!

sxyazi (approved):

PR #1003, opened 2024-05-04
feat: support `cargo binstall yazi-fm` and `cargo binstall yazi-cli`

slowsage:
This will enable binstall installs work directly from github. Required for raspberry pi.
This is needed since crate name is `yazi-fm` but the archive files are named `yazi-`


sxyazi:
Thanks for the PR, is it possible to also add it for the `yazi-cli` crate?

slowsage:
> Thanks for the PR, is it possible to also add it for the `yazi-cli` crate?

Done,

sxyazi (approved):

PR #1012, opened 2024-05-07
feat: `ya pack` displays help if no arguments are given

mikavilpas:
It used to do nothing before this change.


sxyazi (approved):
Nice catch, thanks!

Issue #962, opened 2024-04-27
Add a new `--args` parameter to the `search` command


sxyazi:
### Please describe the problem you're trying to solve

With the new `--args` parameter, users can pass custom arguments to `fd` or `rg`, for example, `search fd --args='-e -H'`.

You can parse the `--args` string into an array using the [`shell-words`](https://crates.io/crates/shell-words) crate. Anyone want to give it a shot?

### Would you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

.

LatentDream:
I'll take a look!

sxyazi:
Thank you! @LatentDream

LatentDream:
@sxyazi
PR here: https://github.com/sxyazi/yazi/pull/1013

PR #1013, opened 2024-05-07
feat: new `--args` parameter for `fd` an `rg` search

LatentDream:
## Please describe the problem you're trying to solve
With the new --args parameter, users can pass custom arguments to fd or rg, for example, search fd --args='-e -H'.

Related issue: Closes https://github.com/sxyazi/yazi/issues/962

## Would you be willing to contribute this feature?

## Describe the solution
Parse the --args string into an array using the shell-words crate.
- rebuilt the input without the `--args` parameter
- pass the args to the `rg` and `fd` commands

## Additional context
- [x] Tested on Windows and MacOS


sxyazi (approved):
Thank you for contributing to this feature!

PR #1019, opened 2024-05-09
fix: `ya pack` supports yazi packages that end in `.yazi`

mikavilpas:
This fixes the following issue:

image

This is what it looks like with this fix applied:

image

### DIFF
--- a/yazi-cli/src/package/deploy.rs
+++ b/yazi-cli/src/package/deploy.rs
@@ -22,8 +22,9 @@ impl Package {
 		let tracker = to.join(TRACKER);
 		if maybe_exists(&to).await && !must_exists(&tracker).await {
 			bail!(
-				"A user package with the same name `{name}` already exists.
-For safety, please manually delete it from your plugin/flavor directory and re-run the command."
+				"A user package with the same name `{}` already exists.
+For safety, please manually delete it and re-run the command.",
+				to.display()
 			);
 		}
 
--- a/yazi-cli/src/package/package.rs
+++ b/yazi-cli/src/package/package.rs
@@ -81,22 +81,6 @@ mod tests {
 	use super::*;
 
 	#[test]
-	fn test_adds_yazi_suffix() {
-		// if a package doesn't have a suffix, it should add ".yazi" to the end
-		let package = Package::new("abcd/my-plugin.yazi", None);
-		assert_eq!(package, Package {
-			repo:      "abcd/my-plugin.yazi".to_owned(),
-			child:     "".to_owned(),
-			commit:    "".to_owned(),
-			is_flavor: false,
-		});
-
-		assert_eq!(package.use_(), "abcd/my-plugin.yazi");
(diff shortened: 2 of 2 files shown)


mikavilpas:
Note: I tried to apply this fix without breaking any existing features, but I'm not sure I covered all the use cases.

I added tests for the existing use cases that I (guessed) are desired, and then made the fix while ensuring the tests continue to pass.

Please check if I missed something! 🙏🏻 🙂

mikavilpas:
Hmm I just realized this actually changes the existing behaviour slightly. Now the user must manually add `.yazi` to the end of their package, while previously it was not required. I'm not sure if this is a good or bad thing 🤔

sxyazi:
Yeah, every package is a directory ending with `.yazi`, so I think adding `.yazi` is a bit redundant and I prefer the previous behavior. So I'm going to close this PR, sorry

mikavilpas:
Do you think both could be supported? With the current behaviour, if I copy the name of the repository from my browser (`user/package.yazi`), I have to manually remove `.yazi` from the end.

Also if I'm not mistaken, yazi has to add `.yazi.git` to get the url to the repository.

What I'm suggesting is:
- if the user wants to add `user/package`, make it `user/package.yazi`
- if the user wants to add `user/package.yazi`, keep it as it is (no need to change)

sxyazi:
I want to keep consistency in this style as much as possible, for example, the [`plugin` command](https://yazi-rs.github.io/docs/plugins/overview#functional-plugin) does not need and cannot have `.yazi` added to it and the same applies to preloader and previewer rules. 

If we support another style here (allowing `.yazi`), we must also consider supporting it elsewhere; otherwise, it will cause further confusion. Users might think, "I used `foo.yazi` when installing, so I might also use `plugin foo.yazi` command"

Also we need to consider what style the `use` in the `package.toml` file should be, if a user runs `ya pack -i foo/bar.yazi`, should the configuration file be `use = "foo/bar"` or `use = "foo/bar.yazi"`? If it's the latter, we'll need to allow for two different styles in the configuration file, which seems messy.

> if I copy the name of the repository from my browser (user/package.yazi), I have to manually remove .yazi from the end.

I don't think this would be a problem because plugin READMEs should provide full installation commands, like

```
ya pack -a foo/bar
```

rather than providing only part of it for users to manually assemble, as this would be more inconvenient and prone to errors.

mikavilpas:
Ok, I think that makes sense. Thanks for the explanation.

Would you consider adding a warning message to make this more explicit? In my case this mistake resulted in this: 
image

SolitudeSF:
why is `.yazi` suffix even required?

sxyazi:
This is to increase recognition and indicate that it is a Yazi package format rather than a traditional Lua module. In traditional Lua modules, `require("foo")` can semantically represent both `foo.lua` and `foo/init.lua` files, while Yazi does not support single files.

Issue #1017, opened 2024-05-08
Update file permission after shell command


raffaem:
### Please describe the problem you're trying to solve

Hover over a file, open shell with `;`, and type `chmod +x $1`. The file is rendered executable, but the permissions shown on the bottom right corner are not updated.

### Would you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

Update file permissions after shell command is run

raffaem:
[20240509T042024.webm](https://github.com/sxyazi/yazi/assets/54762742/12a65fba-e2e0-4f4a-9122-82e237e1f1d9)

sxyazi:
Please try https://github.com/sxyazi/yazi/pull/1024

raffaem:
> Please try #1024

seems to work

kohane27:
I had the same problem in `ya 0.2.5` (stable version installed from archlinux extra package 0.2.5-2).

Confirmed it's fixed in `Yazi 0.2.5 (0ff4835 2024-05-18)`

Thank you!

PR #1024, opened 2024-05-11
fix: broaden file watcher event types to accommodate permission changes on certain platforms

sxyazi:
Fixes https://github.com/sxyazi/yazi/issues/1017


raffaem:
seems to work

PR #1029, opened 2024-05-12
ci: consistently enforce Lua coding style

mikavilpas:
This commit adds a GitHub Actions workflow that checks the lua files against the yazi lua coding style.

The workflow uses the `stylua` tool, which can be installed using cargo:

```sh
cargo install stylua --features lua54
```

I picked the options for the formatter in a way that forced minimal changes to the existing style.

<https://github.com/JohnnyMorganz/StyLua?tab=readme-ov-file#options>

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -39,3 +39,15 @@ jobs:
 
       - name: Rustfmt
         run: cargo +nightly fmt --all -- --check
+
+  stylua:
+    name: stylua
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+      - uses: JohnnyMorganz/stylua-action@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          version: latest
+          # CLI arguments
+          args: --color always --check .
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -41,7 +41,6 @@ jobs:
         run: cargo +nightly fmt --all -- --check
 
   stylua:
-    name: stylua
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
@@ -49,5 +48,4 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           version: latest
-          # CLI arguments
           args: --color always --check .
--- a/stylua.toml
+++ b/stylua.toml
@@ -1,0 +1,3 @@
+indent_width = 2
+call_parentheses = "NoSingleTable"
+collapse_simple_statement = "FunctionOnly"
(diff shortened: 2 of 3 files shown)


mikavilpas:
Hmm, looks like lua 5.4 is not picked up correctly. I will investigate.

mikavilpas:
The formatting is failing because, as I have now learned, stylua does not support `::continue::` because it can't decide whether it's a jump label or a luau type cast operator (https://create.roblox.com/docs/luau/type-checking#type-casts)

This can be solved (at least) the following ways:

1. yazi developers need to install (and use) a stylua version that does not include luau features built in
2. lua code using labels can be refactored to not use them to work around this
3. lua files using labels can be ignored from stylua
4. yazi can have a breaking change and start to use a lua version that's better supported in stylua (this is a bit drastic but listing for completeness)
5. another formatter can be taken into use instead with hopefully better support
6. autoformatting can simply not be implemented at all
7. autoformatting can be implemented on a "best effort" basis: simply let it fail for some files, and don't require autoformatting in CI (can still help ensure consistency)

What would you prefer? My favorite options are 2, 7, and 3.

sxyazi:
I think we can simply ignore it, i.e. `./yazi-plugin/preset/plugins/mime.lua`

.github/workflows/check.yml
```diff
@@ -39,3 +39,15 @@ jobs:
 
       - name: Rustfmt
         run: cargo +nightly fmt --all -- --check
+
+  stylua:
+    name: stylua
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - uses: JohnnyMorganz/stylua-action@v3
```
sxyazi, line 48:
```suggestion
      - uses: actions/checkout@v4
      - uses: JohnnyMorganz/stylua-action@v4
```

Any reason not to use `v4`?

stylua.toml
```diff
@@ -0,0 +1,5 @@
+indent_type = "Tabs"
+indent_width = 2
+column_width = 120
+call_parentheses = "NoSingleTable"
+collapse_simple_statement = "FunctionOnly"
```
sxyazi, line 5:
```suggestion
indent_width = 2
call_parentheses = "NoSingleTable"
collapse_simple_statement = "FunctionOnly"
```

`indent_type = "Tabs"` and `column_width = 120` is the default value

yazi-plugin/preset/plugins/mime.lua
```diff
@@ -1,4 +1,5 @@
-local SUPPORTED_TYPES = "application/audio/biosig/chemical/font/image/inode/message/model/rinex/text/vector/video/x-epoc/"
```
sxyazi, line 1: This file shouldn't be changed since is already ignored

.github/workflows/check.yml
mikavilpas: Hmm not really, let's use `v4` instead 👍🏻

sxyazi (approved):
Thanks!

Issue #903, opened 2024-04-12
Show file extensions if file names are truncated


Brixy:
### Please describe the problem you're trying to solve

Hi guys,

Sometimes it would be important to see file extensions when file names are long and extensions are truncated.

This is particularly important if file icons are ambiguous or deactivated.

### Would you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

I could think of two solutions:

1. truncate file names like `this is a very long f…EXTENSION` or
2. add an `extension` option to `linemode`. (Extensions of untruncated files are shown twice in this case.)

### Additional context

Current workarounds:

- rename, then escape
- toggle full screen
- use plugin to toggle preview pane’s visibility.

Thanks for considering this!

sxyazi:
Tracked by https://github.com/sxyazi/yazi/issues/51

boydaihungst:
https://github.com/boydaihungst/relative-motions.yazi/commit/6bc65a028300dee4ebc07e73c830bca3685a1d23
My way to handle this problem. Calculate dynamically, not fixed length.
![2024-10-18T22:07:32,845587590+07:00](https://github.com/user-attachments/assets/7e815eae-1c51-4730-be0e-5c9a48e04494)
![image](https://github.com/user-attachments/assets/b2f3f8ab-f409-4be8-90ac-8b0428202c4a)

sxyazi:
Supported in https://github.com/sxyazi/yazi/pull/2754

sxyazi:
`"…file extension"` is now supported in https://github.com/sxyazi/yazi/pull/2759

PR #1036, opened 2024-05-14
feat: add "extension" linemode (`m e`) to show extensions of files

mikavilpas:
Solves https://github.com/sxyazi/yazi/issues/903

image


sxyazi:
Thanks for the PR, but I don't like solving it in this way:

- When the file name is fully displayed, two extensions appear, which is a waste of space
- Still causing unexpected truncation when the filename is long enough (possibly not just the extension, but also the filename)

I want to implement it like "long-file-name.mp4" to be "long-fil~.mp4"

mikavilpas:
I see. Do you think this should be the new default behavior, or toggled on with the linemode setting?

sxyazi:
Yeah it should be the default behavior

PR #1037, opened 2024-05-14
feat: add deunicoded, natural sorting

### LINKED ISSUE #616: Sort_by function not work very well with hungarian letters

Linux X11

alacritty 0.11.0

### Yazi version

yazi 0.2.2 (e607943 2024-01-31)

### Did you try the latest main branch to see if the problem has already been fixed?

Not tried, and I'll explain why below

### Describe the bug

Yazi does not take into account accented characters when sorting files and folders. He always puts it at the bottom of the list, like this:

screenshot_1706856499

### Expected Behavior

The correct order would be the following for accented characters (Hungarian):

Alpha
Átrium
Beta
.
Elemér
Érvényes
.
Hedvig
Ilona
Írott
.
Olga
Órmotlan
Öveges
Őzike
.
Ubul
Útvonal
Üveg
Űr
.

### To Reproduce

Create files:
Alpha
Átrium
Beta
Elemér
Érvényes
Hedvig
Ilona
Írott
Olga
Órmotlan
Öveges
Őzike
Ubul
Útvonal
Üveg
Űr

and the problem will be visible

### Configuration

```
[manager]
ratio          = [ 0, 2, 2 ]
# ratio          = [ 0, 2, 2 ]
sort_by        = "natural"
sort_sensitive = false
sort_reverse   = false
sort_dir_first = true

```
### Anything else?

_No response_

clispios:
This resolves #616.

In order to have this work, I pulled in an external dependency: [deunicode](https://docs.rs/deunicode/latest/deunicode/).

This does add slight overhead to sorting as a de-unicoded string can grow in number of characters, for example: Æneid -> AEneid. This means converting the left and right byte vectors into new, de-unicoded byte vectors from strings.

https://github.com/sxyazi/yazi/assets/79223278/d4f24038-42c8-482f-a22a-9ffe7ae9fa6d


sxyazi:
Thanks for the PR, I have never used the `deunicode` crate, can you explain why you chose it?

clispios:
> Thanks for the PR, I have never used the `deunicode` crate, can you explain why you chose it?

Sure! I've used this crate a handful of times both at work and personal projects, so it came to mind when I saw the feature request for sorting special characters.

Reasons why I chose it for this purpose:
- The crate itself has no external dependencies, it can even work with `#[no-std]` if needed. 
- It is single purpose and lightweight (only three files). 
- It has been consistently maintained via the author of the crate (last update was just a couple days ago).
- It is an expanded port of the Perl module `Text::Unidecode` with the added benefit of working with emojis (which are valid characters in file names).
- We could have rolled our own sorting algorithm for dealing with special characters, but I would argue that is non-trivial and shouldn't be a priority of Yazi right now, and this crate solves this need pretty easily.

This is also straight from the description of the crate repository:

> It's a better alternative than just stripping all non-ASCII characters or letting them get [mangled](https://en.wikipedia.org/wiki/Mojibake) by some encoding-ignorant system. It's be okay for one-way conversions for things like search indexes and tokenization, as a stronger version of Unicode NFKD. It may be used for generating nice identifiers for file names and URLs, which aren't too user-facing.

I hope this helps.

sxyazi:
Thank you for the explanation! 

I pulled the code locally for testing, and it works well with Hungarian characters, but when the file list contains not only them but also Chinese characters and emojis, the sorting results become somewhat strange - Chinese characters and emojis are interspersed among files containing Hungarian characters:




And this is the result of natural sorting before:




I think this is due to `deunicode` converting Chinese characters to ASCII Pinyin and emojis to English words, but Chinese characters and emojis themselves are not ordered intrinsically, so I think `deunicode`'s handling scope is too broad.

And I tested another file manager `lf`, I found that it only deals with these Hungarian characters by [removing diacritics](https://github.com/gokcehan/lf/blob/master/diacritics.go#L13-L23), without changing the original sorting rules of Chinese characters and emojis:



So is it possible to adopt the same behavior here and add an option like `sort_transliteration`, which, when `true`, will replace these characters instead of creating a new sorting method `deunicode_natural`?

This is the file list I used for testing:

```
1
Alpha
Átrium
Beta
Elemér
Érvényes
Hedvig
Ilona
Írott
Olga
Órmotlan
Öveges
Ubul
Útvonal
Üveg
一
三
二
叁
贰
🎨
🔥
Űr
Őzike
```

I'm not good at Unicode. If there's a better way to handle it, please let me know!

clispios:
These are all valid points and observations. `deunicode` does indeed translate characters and emojis like you mentioned, and then the natural sorting would sort them as if they were Latin script, which maybe is not what you would want for the application.

If that's the case, which is understandable, I think a simple mapping approach of all known diacritics (like the example you linked) would work just fine. I am also no Unicode expert. Just thought leveraging a broader scope may be useful for future features (like searching for emoji or more complex character files with plan Latin characters, granted I am sure that is more of an edge feature and not a core value use for the scope of the project).

Let's maybe close this pull request, and I can open a new pull request with the implementation of this finer scope feature. I appreciate your input!

PR #1038, opened 2024-05-15
feat: support mouse event

qsdrqs:
Hi. Thanks for this amazing project! This is an attempt to add mouse support to yazi. I make it as a draft PR because I think it still needs tasks to be done before it can be merged. But I would appreciate it if I could get some early feedback from reviewing and help me to know what is the next step.

This PR tries to enable capturing `MouseEvent` in the terminal and spawn it to the `App` struct, then `App` will try to dispatch this event and route this event to the `mouse` method in different components (`Header`, `Manager`, and `Status`), based on coordination comparison between mouse event and `Rect` from `Layout`. The `Manager` is a special case. Because in `LAYOUT`, `Manager` is divided into `Current`, `Parent` and `Preview`. So I create a new `enum ManagerComponent` to indicate which specific component in Manager this MouseEvent belongs to and pass it as an argument. The `mouse` method will finally encapsulate the `MouseEvent` as lua object and invoke the corresponding `mouse` method in lua field.

In the lua field, I create a new type named `MouseEvent`, which contains the following attributes:

| Field     | Description                                                                                           |
|-----------|-------------------------------------------------------------------------------------------------------|
| x         | Column offset relative to the triggered component                                                     |
| y         | Row offset relative to the triggered component                                                        |
| kind      | Kind of mouse event                                                                                   |
| button    | Triggered button of mouse event, will be `nil` for events that are not `Down`, `Up`, and `Drag`       |
| modifiers | Key Modifiers of mouse event, passed as `u8` value                                                    |

The different `kind` and `button` are indicated as different numbers, specifically:

For `kind`:
| Kind        | Value |
|-------------|-------|
| Down        | 0     |
| Up          | 1     |
| Drag        | 2     |
| Moved       | 3     |
| ScrollDown  | 4     |
| ScrollUp    | 5     |
| ScrollLeft  | 6     |
| ScrollRight | 7     |

For `button`
| Button  | Value |
|---------|-------|
| Left    | 0     |
| Right   | 1     |
| Middle  | 2     |

`button` will be `nil` for events other than `Down`, `Up`, and `Drag`.

--------------------------------------------

Currently, only "left click to jump" is fully supported by lua script, which is implemented by `ya.manager_emit("reveal", { f.url })`. For scrolling, only `Current` window supports item-based scrolling. Scrolling in `Preview` will invoke the `seek` in the preview window.

Here are some potential things that people may like to see:
- [ ] Support page-based scrolling like `ranger`
- [ ] Support scrolling in `Preview` and `Parent` window

AFAIK, both of them need a new API for `ya.manager_emit` to invoke. If you think that's a good idea, I will try to implement these. Before that, it would be great if you could review the current implementation and provide some feedback.


sxyazi:
Thank you for the awesome work! 

Mouse support is a feature I've been wanting to add for a while, and I'm very happy to see it implemented now! It definitely deserves a separate version to be released ;)

sxyazi:
Here are some change requests:

- Dispatch `Down(MouseButton)` and `Up(MouseButton)` to the `click` method of *each* component, e.g.:

  ```lua
  function Current:click(event, up)
    -- event = MouseEvent { column = 10, row = 20, modifiers = 0 }
    -- up = true/false
  end
  ```

- Dispatch `ScrollDown`, `ScrollUp`, `ScrollLeft`, `ScrollRight` to the `scroll` method of *each* component, e.g.:

  ```lua
  function Current:scroll(event, step)
    -- event = MouseEvent { column = 10, row = 20, modifiers = 0 }
    -- step = 1 (Down), step = -1 (Up)
  end
  ```

- Dispatch `Moved` to the `move` method of the `Manager` component:

  ```lua
  function Manager:move(event)
    -- event = MouseEvent { column = 10, row = 20, modifiers = 0 }
  end
  ```

- Dispatch `Drag(MouseButton)` to the `drag` method of the `Manager` component:

  ```lua
  function Manager:drag(event, start)
    -- event = MouseEvent { column = 10, row = 20, modifiers = 0 }
    -- start = MouseEvent { column = 10, row = 20, modifiers = 0 }, we need to save the position of the last `Down(MouseButton)` here
  end
  ```

And the `MouseEvent` userdata contains 3 additional fields:

- `is_left` (Boolean)
- `is_right` (Boolean)
- `is_middle` (Boolean)

When the event type is `Moved`, `ScrollDown`, `ScrollUp`, `ScrollLeft`, or `ScrollRight`, the values of these 3 fields are all `false` since these types do not support mouse buttons.

qsdrqs:
>     * Dispatch `ScrollDown`, `ScrollUp`, `ScrollLeft`, `ScrollRight` to the `scroll` method of _each_ component, e.g.:
>       ```lua
>       function Current:scroll(event, step)
>         -- event = MouseEvent { column = 10, row = 20, modifiers = 0 }
>         -- step = 1 (Down), step = -1 (Up)
>       end
>       ```
How to deal with `ScrollLeft`, `ScrollRight` in this `scoll` function?

sxyazi:
I've updated the comment

qsdrqs:
I've finished the requested change, could you have a look at it?

yazi-fm/src/components/manager.rs
```diff
@@ -18,3 +30,85 @@ impl Widget for Manager {
...
+
+impl Manager {
+	pub fn mouse(
+		self,
+		event: crossterm::event::MouseEvent,
+		layout: &Layout,
+		cx: &crate::context::Ctx,
+	) {
+		let position = Position { x: event.column, y: event.row };
+		let comp = if layout.preview.contains(position) {
```
sxyazi, line 42: The `Manager` component should only include the logic for `Moved` and `Drag(MouseButton)`, all others should be dispatched to specific components (`Parent`, `Current`, `Preview`). There are no `parent.rs` and `current.rs` currently, please create them.

yazi-config/src/layout.rs
```diff
@@ -4,6 +4,8 @@ use ratatui::layout::Rect;
 pub struct Layout {
 	pub header: Rect,
 
+	pub manager: Rect,
```
sxyazi, line 7: After dispatching events to specific components (`Current`, `Parent`, `Preview`), this will no longer be necessary, please remove it

yazi-config/preset/yazi.toml
```diff
@@ -12,6 +12,7 @@ linemode       = "none"
 show_hidden    = false
 show_symlink   = true
 scrolloff      = 5
+mouse          = true
```
sxyazi, line 15:
```suggestion
mouse_events          = [ "click", "scroll" ]
```

This allows us to selectively ignore certain minor events, such as `touch`, `move`, `drag`, which can be manually enabled by the user when needed. Due to the high frequency of `move` events, this would be a significant optimization.

yazi-config/src/manager/manager.rs
```diff
@@ -20,6 +20,7 @@ pub struct Manager {
 	pub show_hidden:  bool,
 	pub show_symlink: bool,
 	pub scrolloff:    u8,
+	pub mouse:        bool,
```
sxyazi, line 23:
```suggestion
	pub mouse:        MouseEvents,
```

`MouseEvents` looks like:

```rust
bitflags! {
	#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
  #[serde(try_from = "Vec<String>")]
	pub struct MouseEvents: u8 {
		const CLICK  = 0b00001;
		const SCROLL = 0b00010;
		const TOUCH  = 0b00100;
		const MOVE   = 0b01000;
		const DRAG   = 0b10000;
	}
}
```

You can check out 

https://github.com/sxyazi/yazi/blob/28dfe728ab541f71be27f131e985723253bcff6e/yazi-config/src/popup/offset.rs#L13

 for implementing it.

yazi-config/src/layout.rs
qsdrqs: However, the `Manager` component still need this `Rect` to calculate the relative mouse position when invoking the `Moved` and `Drag` in the manager component. I think we still need to keep this for the manager when handling these events.

sxyazi: The `Manager` component is the whole terminal window, should its relative coordinates and absolute coordinates be the same, or am I getting something wrong?

qsdrqs: I don't think so? At least the `Manager` does not overlap with `Status` and `Header`. I tried to  print all Rect in the terminal and get:
```
manager: Rect { x: 0, y: 1, width: 101, height: 53 }
header: Rect { x: 0, y: 0, width: 101, height: 1 }
status: Rect { x: 0, y: 54, width: 101, height: 1 }
preview: Rect { x: 64, y: 1, width: 36, height: 53 }
current: Rect { x: 13, y: 1, width: 50, height: 53 }
parent: Rect { x: 1, y: 1, width: 11, height: 53 }
```

sxyazi: Ah you're right!

It's been so long since I looked at that part that I forgot haha.

qsdrqs:
Finished the review requests as described (if I understand correctly).

sxyazi:
Thanks for the change! I will review it in the next 1 or 2 days

sxyazi (approved):
Everything looks good now, let me merge it then. 

Thank you so much for contributing this awesome feature!

aspauldingcode:
Does clock and drag work for selecting items? Like spacebar selection

qsdrqs:
> Does click and drag work for selecting items? Like spacebar selection

I didn't implement these in this PR. This PR only has built-in support for scroll and click. However I think it should be easy to implement through the current Lua API.

PR #1042, opened 2024-05-15
feat: allow quitting the app with escape when there's no selection

janbuchar:
This allows making a key binding that first cancels visual mode, search, selections, etc., and if there is no such thing, it quits Yazi. 

I use Yazi in neovim inside a floating window and my other tools can be closed with escape as well, so this is nice for me usability-wise, and should be opt-in.

How can I update the docs, do I just submit another PR to the website repo?

### DIFF
--- a/yazi-core/src/tab/commands/escape.rs
+++ b/yazi-core/src/tab/commands/escape.rs
@@ -1,16 +1,21 @@
 use bitflags::bitflags;
 use yazi_proxy::{AppProxy, ManagerProxy};
-use yazi_shared::{event::Cmd, render, render_and};
+use yazi_shared::{
+	event::{Cmd, Event, EventQuit},
+	render, render_and,
+};
 
 use crate::tab::Tab;
 
 bitflags! {
 	pub struct Opt: u8 {
-		const FIND   = 0b00001;
-		const VISUAL = 0b00010;
-		const SELECT = 0b00100;
-		const FILTER = 0b01000;
-		const SEARCH = 0b10000;
+		const FIND    = 0b0000001;
+		const VISUAL  = 0b0000010;
+		const SELECT  = 0b0000100;
+		const FILTER  = 0b0001000;
+		const SEARCH  = 0b0010000;
+		const QUIT    = 0b0100000;
+		const CASCADE = 0b1000000;
 	}
 }
 
@@ -18,12 +23,14 @@ impl From<Cmd> for Opt {
 	fn from(c: Cmd) -> Self {
 		c.args.iter().fold(Opt::empty(), |acc, (k, v)| {
 			match (k.as_str(), v.as_bool().unwrap_or(false)) {
-				("all", true) => Self::all(),
+				("all", true) => acc | (Self::all() ^ Self::QUIT ^ Self::CASCADE),
 				("find", true) => acc | Self::FIND,
 				("visual", true) => acc | Self::VISUAL,
 				("select", true) => acc | Self::SELECT,
 				("filter", true) => acc | Self::FILTER,
 				("search", true) => acc | Self::SEARCH,
+				("quit", true) => acc | Self::QUIT,
+				("cascade", true) => acc | Self::CASCADE,
 				_ => acc,
 			}
 		})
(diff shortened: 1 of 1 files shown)


yazi-core/src/tab/commands/escape.rs
```diff
@@ -1,29 +1,36 @@
...
-		const SELECT = 0b00100;
-		const FILTER = 0b01000;
-		const SEARCH = 0b10000;
+		const FIND    = 0b0000001;
+		const VISUAL  = 0b0000010;
+		const SELECT  = 0b0000100;
+		const FILTER  = 0b0001000;
+		const SEARCH  = 0b0010000;
+		const QUIT    = 0b0100000;
+		const CASCADE = 0b1000000;
```
janbuchar, line 18: This is to make it possible to explicitly invoke the default behavior (what happened previously when `opt` was empty)

sxyazi:
Sorry, I'm afraid I can't accept this PR. 

I prefer to keep each command as simple as possible, and this new `escape --quit` command is essentially a combination of the `escape` and `quit` commands, which leads to overlapping functionality. 

I prefer each command to do only one thing. If you want to implement exiting by pressing `Esc`, I suggest implementing it through a simple plugin:

```lua
-- ~/.config/yazi/plugins/smart-escape.yazi/init.lua
return {
  entry = function()
    local active = cx.active
    local current = active.current

    local esc = active.mode.is_visual or #active.selected > 0 or current.files.filter or current.cwd.is_search
    ya.manager_emit(esc and "escape" or "quit", {})
  end,
}
```

Then bind it to your `<Esc>` key:

```toml
[[manager.prepend_keymap]]
on  = [ "<Esc>" ]
run = "plugin --sync smart-escape"
```

Note: There's a missing check for `escape_find` because [`finder`](https://github.com/sxyazi/yazi/blob/28dfe728ab541f71be27f131e985723253bcff6e/yazi-core/src/tab/commands/escape.rs#L62) is currently not exposed to the Lua API, but I can add it if you need.

janbuchar:
> I prefer to keep each command as simple as possible, and this new `escape --quit` command is essentially a combination of the `escape` and `quit` commands, which leads to overlapping functionality. 

Fair enough.

> I prefer each command to do only one thing. If you want to implement exiting by pressing `Esc`, I suggest implementing it through a simple plugin:

Thanks for the snippet! Can I find a detailed reference for the plugin api anywhere? When I checked, it was hard for me to find out what functionality is exposed through the context.

> Note: There's a missing check for `escape_find` because [`finder`](https://github.com/sxyazi/yazi/blob/28dfe728ab541f71be27f131e985723253bcff6e/yazi-core/src/tab/commands/escape.rs#L62) is currently not exposed to the Lua API, but I can add it if you need.

Well, I'm not sure if I need it now. However, it seems wrong to me to duplicate (and keep in sync) internals of the escape command in a plugin. Could we expose the return value of the command to the plugin API? Or is it already there?

sxyazi:
> Can I find a detailed reference for the plugin api anywhere? 

Here it is https://yazi-rs.github.io/docs/plugins/types#app-data

> it seems wrong to me to duplicate (and keep in sync) internals of the escape command in a plugin

I didn't quite understand this, could you explain it in more detail?

> Could we expose the return value of the command to the plugin API? Or is it already there?

Not supported at the moment, but it may be added in the future.

I still need to think it through carefully because currently, all commands do not return a value, including the `escape` command here. Its function signature is:

```rust
pub fn escape(&mut self, opt: impl Into<Opt>) {
```

Instead of something like:

```rust
pub fn escape(&mut self, opt: impl Into<Opt>) -> bool {
```

This would be a large change to the event system.

janbuchar:
> > it seems wrong to me to duplicate (and keep in sync) internals of the escape command in a plugin
> 
> I didn't quite understand this, could you explain it in more detail?

Well, in the plugin, we basically rewrite a part of the escape command in lua. If a new UI element that is controlled with the escape key is added, I will have to reflect that in the plugin. That doesn't seem ideal.

> > Could we expose the return value of the command to the plugin API? Or is it already there?
> 
> Not supported at the moment, but it may be added in the future.
> 
> I still need to think it through carefully because currently, all commands do not return a value, including the `escape` command here. 
> This would be a large change to the event system.

I understand, and it may be too much of a niche use case anyway. Still, it would make it trivial to "extend" the escape command.

sxyazi:
> If a new UI element that is controlled with the escape key is added, I will have to reflect that in the plugin

Yes, this is currently a problem with the plugin system. However, it's unlikely that the behavior of the escape key will be changed because I consistently refuse to add new behaviors to it. An example is the 'unyank' command, which was added as a new command rather than being integrated into the 'escape' sequence. 

Adding too many behaviors would cause inconvenience. Currently, some actions require pressing `Esc` twice in a row to reach the desired state, and adding more behaviors might increase this to three presses, which is not good for user experience.

> it would make it trivial to "extend" the escape command

I plan to add a method parallel to `ya.manager_emit` in the future called `ya.manager_exec`. This method will allow for obtaining the return value of commands (if any). However, this involves redesigning the event and plugin system, so it's not a priority right now. The current priority is to ensure the stability of the plugin API.

---

I am going to close this PR. If you have any other questions, feel free to reply.

janbuchar:
> I plan to add a method parallel to `ya.manager_emit` in the future called `ya.manager_exec`. This method will allow for obtaining the return value of commands (if any). However, this involves redesigning the event and plugin system, so it's not a priority right now. The current priority is to ensure the stability of the plugin API.

Okay, sounds good!

PR #1043, opened 2024-05-16
fix: remove `ignore` options from `rg` and `fd` search

chriszarate:
When `yazi`'s `hidden` option is true, it passes `--no-ignore` to `fd` and `rg`. However, ignore files have no actual relation to hidden files. They often contain entries like `node_modules` that users want to ignore in all situations.

Example use case: In a code project, I turn on the `hidden` option to navigate to configuration files like `.editorconfig` and search their contents with <kbd>S</kbd>. I do not want the `hidden` option to expose thousands of search results in `node_modules`, `.gocache`, or `dist`.

### DIFF
--- a/yazi-plugin/src/external/fd.rs
+++ b/yazi-plugin/src/external/fd.rs
@@ -16,7 +16,7 @@ pub fn fd(opt: FdOpt) -> Result<UnboundedReceiver<File>> {
 		.arg("--base-directory")
 		.arg(&opt.cwd)
 		.arg("--regex")
-		.args(if opt.hidden { ["--hidden"] } else { ["--no-hidden"] })
+		.arg(if opt.hidden { "--hidden" } else { "--no-hidden" })
 		.args(opt.args)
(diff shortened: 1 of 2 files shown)


yazi-plugin/src/external/fd.rs
```diff
@@ -16,7 +16,7 @@ pub fn fd(opt: FdOpt) -> Result<UnboundedReceiver<File>> {
 		.arg("--base-directory")
 		.arg(&opt.cwd)
 		.arg("--regex")
-		.args(if opt.hidden { ["--hidden", "--no-ignore"] } else { ["--no-hidden", "--ignore"] })
+		.args(if opt.hidden { ["--hidden"] } else { ["--no-hidden"] })
```
sxyazi, line 19:
```suggestion
		.arg(if opt.hidden { "--hidden" } else { "--no-hidden" })
```

Please use `arg()` instead. `rg.rs` is also.

sxyazi (approved):

Issue #1023, opened 2024-05-10
Confusing error on plugin installation 


mikavilpas:
### What system are you running Yazi on?

macOS

### What terminal are you running Yazi in?

Wezterm 

### Did you try the latest code to see if this problem got fixed?

Tried, but the problem still

### `yazi --debug` output

<details>

```sh

```

</details>

### Describe the bug

.

### Expected Behavior

.

### To Reproduce

See picture 

### Anything else?

Would you consider adding a warning message to make this more explicit? In my case this mistake resulted in this: 
![image](https://github.com/sxyazi/yazi/assets/300791/deef339a-033c-48c6-a00b-edf5591cc9a2)

_Originally posted by @mikavilpas in https://github.com/sxyazi/yazi/issues/1019#issuecomment-2103292905_

mikavilpas:
The issue is that I tried to install the package `glow.yazi`, which is converted to `glow.yazi.yazi` (which doesn't exist).

mikavilpas:
lazy.nvim takes the opposite approach to what you described in https://github.com/sxyazi/yazi/pull/1019#issuecomment-2103288700. It doesn't change the url when you want to add a new package. This can result in confusion when you actually use the package in lua code.

PR #1060, opened 2024-05-19
fix: ya pack --upgrade not being able to upgrade existing packages

mikavilpas:
With the following packages, running the `ya pack --upgrade` command results in a prompt for a username:

```sh
❯ cat ~/dotfiles/.config/yazi/package.toml
[plugin]
deps = [{ use = "DreamMaoMao/keyjump.yazi", commit = "06383de" }, { use = "Rolv-Apneseth/starship.yazi", commit = "6197e4c" }]

[flavor]
deps = []
❯ ya pack --upgrade

  Upgrading package `keyjump.yazi.yazi`

Cloning into '/Users/mikavilpas/.local/state/yazi/packages/109b2bd4c6591dcb62b40b25202be168'...
Username for 'https://github.com':
```

The problem seems to be that the plugin `DreamMaoMao/keyjump.yazi` is being resolved to `https://github.com/DreamMaoMao/keyjump.yazi.yazi` with an extra `.yazi` at the end.

We discussed this behaviour (and decided it's preferred to add `.yazi` when adding new packages) in

- https://github.com/sxyazi/yazi/pull/1019
- https://github.com/sxyazi/yazi/issues/1023

...but I did not realize at the time that it also affects upgrading packages.

### DIFF
--- a/yazi-cli/src/package/package.rs
+++ b/yazi-cli/src/package/package.rs
@@ -12,6 +12,8 @@ pub(crate) struct Package {
 }
 
 impl Package {
+	/// Create a new Package when adding a new package to the config.
+	/// Note that the package must end in `".yazi"`.
 	pub(super) fn new(url: &str, commit: Option<&str>) -> Self {
 		let mut parts = url.splitn(2, '#');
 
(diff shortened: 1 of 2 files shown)


sxyazi:
Hi, no `.yazi` should be added anywhere in the package manager domain. Because `use` is not strictly a repository name/URL; it could also refer to a sub-package within a repository, such as `yazi-rs/flavors#catppuccin-mocha`. This is also ultimately why I decided to change from `repo` to `use`. Closing as it's expected behavior.

mikavilpas:
Oh, I see. It must have been the case that my configuration is actually wrong!

In that case I agree, this should be closed as expected.

Issue #1062, opened 2024-05-20
[Bug] yazi extracts .epub files instead of opening them


DestroyerAce:
### What system are you running Yazi on?

Linux Wayland

### What terminal are you running Yazi in?

foot 1.17.2

### Did you try the latest code to see if this problem got fixed?

Tried, but the problem still

### `yazi --debug` output

<details>

```sh

Yazi
    Version: 0.2.5 (Arch Linux 2024-04-27)
    OS: linux-x86_64 (unix)
    Debug: false

Emulator
    Emulator.via_env: ("foot-extra", "")
    Emulator.via_csi: Ok(Foot)
    Emulator.detect: Foot

Adaptor
    Adaptor.matches: Sixel

Desktop
    XDG_SESSION_TYPE: Some("wayland")
    WAYLAND_DISPLAY: Some("wayland-1")
    DISPLAY: Some(":0")

SSH
    shared.in_ssh_connection: false

WSL
    /proc/sys/fs/binfmt_misc/WSLInterop: false

Variables
    SHELL: Some("/usr/bin/fish")
    EDITOR: Some("nvim")
    ZELLIJ_SESSION_NAME: None
    YAZI_FILE_ONE: None
    YAZI_CONFIG_HOME: None

file(1)
    Version: Ok(Output { status: ExitStatus(unix_wait_status(0)), stdout: "file-5.45\nmagic file from /usr/share/file/misc/magic\nseccomp support included\n", stderr: "" })

Text Opener
    default: Some(Opener { run: "${EDITOR:=vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block: Some(Opener { run: "${EDITOR:=vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })

tmux
    TMUX: false

Ueberzug++
    Version: Ok(Output { status: ExitStatus(unix_wait_status(0)), stdout: "ueberzugpp 2.9.5\n", stderr: "" })


--------------------------------------------------
When reporting a bug, please also upload the `yazi.log` log file - only upload the most recent content by time.
You can find it in the "/home/akash/.local/state/yazi" directory.


```

</details>

### Describe the bug

`.epub` files not opening in the default app set by `xdg-mime` and instead getting extracted.

### Expected Behavior

The `.epub` file opening with the default app.

### To Reproduce

1. Launch `yazi`
2. Navigate to any file with `.epub` extension
3. Press enter

[...]

PR #1063, opened 2024-05-20
fix: ePUB file mime-type matching for the opener rule

sxyazi:
Fixes https://github.com/sxyazi/yazi/issues/1062


DestroyerAce:
It seems to work fine without any issue

sxyazi:
Alright lemme merge then

PR #1082, opened 2024-05-23
feat: support case insensitive special keys in keymappings

mikavilpas:
This enables mapping `"<right>"` instead of the current behavior, which requires `"<Right>"` and silently ignores the mapping in lowercase.


mikavilpas:
Forgot to add that case insensitive mappings are the default in vim, which has very similar syntax.

yazi-config/src/keymap/key.rs
```diff
@@ -77,40 +77,40 @@ impl FromStr for Key {
...
+				"f7" => key.code = KeyCode::F(7),
+				"f8" => key.code = KeyCode::F(8),
+				"f9" => key.code = KeyCode::F(9),
+				"f10" => key.code = KeyCode::F(10),
+				"f11" => key.code = KeyCode::F(11),
+				"f12" => key.code = KeyCode::F(12),
+				"esc" => key.code = KeyCode::Esc,
 
 				c if it.peek().is_none() => {
 					key.code = KeyCode::Char(c.chars().next().unwrap());
```
sxyazi, line 116: This will cause `KeyCode::Char` to be converted to lowercase as well

mikavilpas: Hmm, good point. I applied what I think will fix this, but was not able to detect a difference in the behaviour of `G` (go to last item) and `gg` (go to first item).

I think the way I tested the fix might be unreliable. Do you know what could be used to test it?

sxyazi: Only key expressions wrapped in `<>` will enter this branch, so key combinations like `<C-z>` and `<C-Z>` should be usable for testing

sxyazi:
The new code looks fine, I'll test it later and then merge it

sxyazi (approved):
Thanks!

Issue #1130, opened 2024-06-06
yazi process gets killed when you switch to shell and got a nonzero return value


ih8d8:
### What system are you running Yazi on?

Linux Wayland

### What terminal are you running Yazi in?

Alacritty + tmux

### Did you try the latest code to see if this problem got fixed?

Not tried, and I'll explain why below

### `yazi --debug` output

<details>

```
Yazi
    Version: 0.2.5 (Arch Linux 2024-04-27)
    OS: linux-x86_64 (unix)
    Debug: false

Emulator
    Emulator.via_env: ("alacritty", "tmux")
    Emulator.via_csi: Ok(Unknown([]))
    Emulator.detect: Unknown([])

Adaptor
    Adaptor.matches: Wayland

Desktop
    XDG_SESSION_TYPE: Some("wayland")
    WAYLAND_DISPLAY: Some("wayland-1")
    DISPLAY: Some(":1")

SSH
    shared.in_ssh_connection: false

WSL
    /proc/sys/fs/binfmt_misc/WSLInterop: false

Variables
    SHELL: Some("/bin/bash")
    EDITOR: Some("vim")
    ZELLIJ_SESSION_NAME: None
    YAZI_FILE_ONE: None
    YAZI_CONFIG_HOME: None

file(1)
    Version: Ok(Output { status: ExitStatus(unix_wait_status(0)), stdout: "file-5.45\nmagic file from /usr/share/file/misc/magic\nseccomp support included\n", stderr: "" })

Text Opener
    default: Some(Opener { run: "${EDITOR:=vim} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block: Some(Opener { run: "${EDITOR:=vim} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })

tmux
    TMUX: true

Ueberzug++
    Version: Ok(Output { status: ExitStatus(unix_wait_status(0)), stdout: "ueberzugpp 2.9.5\n", stderr: "" })


--------------------------------------------------


```

</details>

### Describe the bug

I have these lines in my keymap.toml file:

	{ on = [ "'", "r" ], 		run = 'shell "rsync -auvP --info=progress2 \"$@\" \"$PWD\"" --block --confirm', 							desc = "Copy with rsync" },
	{ on = [ "'", "x" ], 		run = 'shell "7z x \"$@\"" --block --confirm',																desc = "Extract password-protected archives" },

[...]

PR #1131, opened 2024-06-06
fix: block `SIGINT` signal from the spawned subprocess

sxyazi:
Fixes https://github.com/sxyazi/yazi/issues/1130


ArtyomArtamonov:
@sxyazi hey, I just found out that this commit changes something in preview. If file contains carriage return (ascii code 13), it will show some characters on the next line at the start (left side of the window). I can file an issue if you want, and describe it there.

Issue #1089, opened 2024-05-26
Word wrapping in preview


mskvsk:
### Please describe the problem you're trying to solve

Word wrapping in preview has been a pretty contentious issue in the discussions of at least two terminal file managers. Some people like to scroll it, some people prefer it to word wrap.

Personally, I would like some of my documents to word wrap, while the others and code to not to.

I suspect, many people have their own expectations around that which I will try to address in the solution.

### Would you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

An optional "wrap" field in the previewers array in yazi.toml

True for word wrapping, false for long lines. The default is false. Ignored if the active preview is not wrappable.

Example:

```
previewers = [
	{ name = "*/", run = "folder", sync = true },
	# Code
	{ mime = "text/*", run = "code" },
	{ mime = "*/{xml,javascript,x-wine-extension-ini}", run = "code" },
	# JSON
	{ mime = "application/json", run = "json", wrap = true},
]

```

mskvsk:
Also, I have never written any Rust but can give it a shot if @sxyazi deems this feature good to have.

We can also downscale the scope of this feature to a single boolean (wordwrap = true/false) in the config if it seems an overkill.

All I need is my text files to word wrap as I use text preview on them daily.

sxyazi:
Nice to have!

Would you like to raise a PR for it? Here is the method that handles code highlighting: https://github.com/sxyazi/yazi/blob/9df256f7f4abc312ae94b59c948502310ecb2c87/yazi-plugin/src/external/highlighter.rs#L59

Maybe we can change its function signature to:

```rust
pub async fn highlight(&self, skip: usize, area: Rect, wrap: bool) -> Result<Text<'static>, PeekError> {
```

And handle text wrapping when `wrap = true`.

mskvsk:
Sure, will do. I'm half-way through the solution alredy.

sxyazi:
If anyone is interested in implementing it, https://github.com/sxyazi/yazi/pull/1142 would be a great starting point.

mskvsk:
I might also add that I am ready to pay a small bounty of $50 in crypto (USDC/USDT) to anyone who solves this issue without using an external crate (I don't mind if you copypaste the solution from the crate I have used in my PR).

This feature is quite nice to have to me, as I use Yazi to browse a lot of personal content that is stored as separate text files which don't look good without wrapping.

[...]

PR #1142, opened 2024-06-08
feat: word wrap in preview

mskvsk:
As per https://github.com/sxyazi/yazi/issues/1089

I figured that we start with a simple flag in yazi.toml and see if there is a demand for the more fine solution outlined in the issue.

Word wrapping turned out to be a deeper problem than I expected so I had to utilize an external crate for that.

### DIFF
--- a/yazi-plugin/Cargo.toml
+++ b/yazi-plugin/Cargo.toml
@@ -39,6 +39,7 @@ tokio         = { version = "1.38.0", features = [ "full" ] }
 tokio-stream  = "0.1.15"
 tokio-util    = "0.7.11"
 unicode-width = "0.1.13"
+textwrap	  = "0.14.2"
 yazi-prebuild = "0.1.2"
 
 # Logging
--- a/yazi-plugin/src/external/highlighter.rs
+++ b/yazi-plugin/src/external/highlighter.rs
@@ -6,6 +6,8 @@ use syntect::{dumps, easy::HighlightLines, highlighting::{self, Theme, ThemeSet}
 use tokio::{fs::File, io::{AsyncBufReadExt, BufReader}};
 use yazi_config::{PREVIEW, THEME};
 use yazi_shared::PeekError;
+use ratatui::prelude::Rect;
+use textwrap::wrap as textwrap_wrap;
 
 static INCR: AtomicUsize = AtomicUsize::new(0);
 static SYNTECT_SYNTAX: OnceLock<SyntaxSet> = OnceLock::new();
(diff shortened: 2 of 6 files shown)


yazi-plugin/Cargo.toml
```diff
@@ -39,6 +39,7 @@ tokio         = { version = "1.38.0", features = [ "full" ] }
 tokio-stream  = "0.1.15"
 tokio-util    = "0.7.11"
 unicode-width = "0.1.13"
+textwrap	  = "0.14.2"
```
sxyazi, line 42: I want to avoid introducing new crates as much as possible. Is it possible to write a simple function to handle line breaks? 

We can use the [`unicode-width`](https://crates.io/crates/unicode-width) crate to get the character width and it's already in the dependency tree.

yazi-plugin/src/external/highlighter.rs
```diff
@@ -84,23 +86,40 @@ impl Highlighter {
...
-				before.push(String::from_utf8_lossy(&buf).into_owned());
+				before.push(line);
 			}
 			buf.clear();
 		}
 
-		if skip > 0 && i < skip + limit {
-			return Err(PeekError::Exceed(i.saturating_sub(limit)));
+		if skip > 0 && i < skip + area.height as usize {
+			return Err(PeekError::Exceed(i.saturating_sub(area.height as usize)));
```
sxyazi, line 100: Here we need to consider the number of lines after wrapping; otherwise, it will cause the [`seek` command](https://yazi-rs.github.io/docs/configuration/keymap/#manager.seek) abnormal, i.e. making it impossible to scroll to the very end of the file.

mskvsk: Good catch! Unfortunately, after analyzing the issue I have to admit that it is way above my pay grade. The complexity introduced by preview scrolling makes the implementation much more difficult than my close to zero Rust skills supported by LLM can handle.

I'm sorry but I have to capitulate here and leave the baton for the more skilled developer to pick up.

yazi-plugin/Cargo.toml
mskvsk: I tried to push ChatGPT to do that but after about 50 iterations concluded that it's pretty hard due to various edge cases. This crate solved the issue momentarily.

Although, I admit that probably my Rust skills or the lack thereof has considerably contributed to the difficulty :)

sxyazi:
No worries, let me close it and reassign the issue. @mskvsk thanks for the effort!

PR #1143, opened 2024-06-09
feat: support `enable` for which-key

gaesa:
### Description

This PR adds an `enable` feature for the "which key" functionality. When `enable` is set to `false`, the "which key" feature will be disabled, and the corresponding key hints will not be shown. The `enable` defaults to `true` to keep backward compatibility.

### Changes Made

1. Added `enable` field in the `Which` struct in `yazi-config/src/which/which.rs`.
2. Updated the default configuration in `yazi-config/preset/yazi.toml` to include the `enable` field.
3. Modified the `show_with` method in `yazi-core/src/which/commands/show.rs` to respect the `enable` setting.

### DIFF
--- a/yazi-config/preset/yazi.toml
+++ b/yazi-config/preset/yazi.toml
@@ -185,10 +185,11 @@ open_origin = "hovered"
 open_offset = [ 0, 1, 50, 7 ]
 
 [which]
+enable           = true
 sort_by      	 = "none"
-sort_sensitive = false
+sort_sensitive   = false
 sort_reverse 	 = false
-sort_translit  = false
+sort_translit    = false
 
 [log]
 enabled = false
--- a/yazi-config/src/which/which.rs
+++ b/yazi-config/src/which/which.rs
@@ -6,6 +6,7 @@ use crate::MERGED_YAZI;
 
 #[derive(Debug, Deserialize, Serialize, Validate)]
 pub struct Which {
+	pub enable:         bool,
 	// Sorting
 	pub sort_by:        SortBy,
 	pub sort_sensitive: bool,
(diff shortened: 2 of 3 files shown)


sxyazi:
Sorry, but I'm afraid I can't accept this PR for the following reasons:

1. It breaks existing functionality – when `enable = false`, all consecutive key presses stop working, such as copying file names (`c` => `n`):
  https://github.com/sxyazi/yazi/blob/189cb81db335650889be6f02406526fecf8e0fb1/yazi-config/preset/keymap.toml#L93-L97

2. I don't see the point in disabling the `which` component. The `which` component is closely linked to Yazi's routing system, and disabling it would cause functionality abnormal.

3. I want to keep Yazi's configuration as simple as possible and provide a set of reasonable default presets. Adding every possible decision as an option would not only bloat the code but also raise the entry barrier for new users.

sxyazi:
> As a file manager, Yazi should also provide users with the choice to enable or disable this component

AFAIK, ranger, lf, and joshuto do not support disabling it - these are the file managers I know of that support key hints, although they only support displaying a single column vertically.

> After all, unlike Emacs and Neovim's which-key plugins, the current "which" component in Yazi does not support a timeout setting.

This is why I didn't consider disabling it in the design - I wanted to avoid introducing a timeout mechanism, as it would increase design complexity and is not always reliable (considering high-latency scenarios like SSH).

The "which" component is only applicable to multiple consecutive key presses. Disabling it would leave users unaware of their key press status and require an additional timeout mechanism to cancel the key press (for example, for consecutive key presses like `c` => `n`, if the user presses `c` but does not press `n`, the `c` state needs to be canceled after a certain time). However, if the "which" component is always available, it provides users with assurance that they can always proceed to press the next key.

> This is just an option and does not significantly increase the complexity or the entry barrier for new users

This would increase the amount of configuration options and documentation new users need to read and understand when getting started with and using Yazi.

PR #1152, opened 2024-06-13
feat: Introduce remote scheme

Xuanwo:
This PR intends to introduce the new remote scheme concept in yazi. After this new feature introduced, users can use yazi to visit and manage files store in remote storage services like s3, gcs, webdav, dropbox and so on.

This PR just a start of this PR and not integrate with file operations yet.


Xuanwo:
cc @sxyazi, I'm not sure if this PR is the right size for merging. Could you first review the overall design?

sxyazi:
Thanks for the PR!

I think we should create a new crate `yazi-fs` and move `yazi-shared/src/fs/` into this new crate - all integration work can be done within this new crate. 

As for the scheme configuration, I don't think it's reasonable to keep it in `yazi.toml` because people often host their `~/.config` directory on Git services, which could lead to sensitive information (like usernames and passwords) being leaked. I think the `state` directory (`~/.local/state`) would be a better location for it.

sxyazi:
> I'm not sure if this PR is the right size for merging

Typically, I expect a complete feature to be merged. Perhaps we can create a separate branch for ongoing work, or you could just give me write access to your fork so I can participate in the development?

Xuanwo:
> Perhaps we can create a separate branch for ongoing work, or you could just give me write access to your fork so I can participate in the development?

Yes, I have enabled `Allow edits and access to secrets by maintainers`. You can push to this branch directly.



> I think we should create a new crate `yazi-fs` and move `yazi-shared/src/fs/` into this new crate - all integration work can be done within this new crate.

Nice idea, I will do this.

Xuanwo:
> As for the scheme configuration, I don't think it's reasonable to keep it in `yazi.toml` because people often host their `~/.config` directory on Git services, which could lead to sensitive information (like usernames and passwords) being leaked. I think the `state` directory (`~/.local/state`) would be a better location for it.

DONE, please take a look.


> I think we should create a new crate `yazi-fs` and move `yazi-shared/src/fs/` into this new crate

I tried but failed. `yazi-shared` and `yazi-fs` will depend on each other.

sxyazi:
> DONE, please take a look.

Thanks, the config for remote services should be lazily initialized on use – I'll refactor it. 

> I tried but failed. yazi-shared and yazi-fs will depend on each other.

Why does `yazi-shared` need to depend on `yazi-fs`? 

> I'm trying to make yazi mys3://path/to/file works but find that there are many places need to hack with:

Yeah currently Yazi's filesystem isn't abstracted and is full of local fs API calls. I think the first step should be to abstract the common APIs into the `yazi-fs` crate and implement it in a cross-scheme way, so it can be used everywhere uniformly.

PR #1159, opened 2024-06-16
feat: word wrapping in `code` previewer

ArtyomArtamonov:
This PR tries to add feature described in #1089 

I tried to implement feature without introducing new dependencies, like it was asked in #1142. Also I used #1142 commits as baseline to my changes (configuration from yazi.toml, highlight signature). 

I tried to separate new logic with two methods: `before_after` and `before_after_wrapped`. `before_after` is almost the same with minor change in readability `no_more_scroll`.
`before_after_wrapped` is more complicated. I had to read whole file first and apply logic afterwards. We have to decide if file is plaintext or not because we only apply indentation (tab_size in config) to plaintext files (in this method at least).
After that I chunk every line of file in chunks by column width using [unicode-width](https://crates.io/crates/unicode-width) crate as was said. I tried my solution in some cases including ascii characters, cyrrilic alphabet, emojis. All of them looks fine.

It's my first PR using rust, I tried to come up with optimized solution, but this is as far as I could go. If you have any suggestions how we can improve it, please tell, I'm open to suggestions.

### DIFF
--- a/yazi-plugin/src/external/highlighter.rs
+++ b/yazi-plugin/src/external/highlighter.rs
@@ -1,10 +1,10 @@
-use std::{io::Cursor, mem, path::{Path, PathBuf}, sync::atomic::{AtomicUsize, Ordering}};
+use std::{borrow::Cow, io::Cursor, mem, path::{Path, PathBuf}, sync::atomic::{AtomicUsize, Ordering}};
 
 use anyhow::{anyhow, Result};
-use ratatui::text::{Line, Span, Text};
+use ratatui::{layout::Rect, text::{Line, Span, Text}};
 use syntect::{dumps, easy::HighlightLines, highlighting::{self, Theme, ThemeSet}, parsing::{SyntaxReference, SyntaxSet}, LoadingError};
 use tokio::{fs::File, io::{AsyncBufReadExt, BufReader}, sync::OnceCell};
-use yazi_config::{PREVIEW, THEME};
+use yazi_config::{preview::PreviewWrap, PREVIEW, THEME};
 use yazi_shared::PeekError;
 
 static INCR: AtomicUsize = AtomicUsize::new(0);
@@ -41,23 +41,20 @@ impl Highlighter {
 	#[inline]
 	pub fn abort() { INCR.fetch_add(1, Ordering::Relaxed); }
 
-	pub async fn highlight(&self, skip: usize, limit: usize) -> Result<Text<'static>, PeekError> {
+	pub async fn highlight(&self, skip: usize, area: Rect) -> Result<Text<'static>, PeekError> {
 		let mut reader = BufReader::new(File::open(&self.path).await?);
 
 		let syntax = Self::find_syntax(&self.path).await;
 		let mut plain = syntax.is_err();
 
 		let mut before = Vec::with_capacity(if plain { 0 } else { skip });
-		let mut after = Vec::with_capacity(limit);
+		let mut after = Vec::with_capacity(area.height as _);
 
 		let mut i = 0;
 		let mut buf = vec![];
 		let mut inspected = 0u16;
-		while reader.read_until(b'\n', &mut buf).await.is_ok() {
-			i += 1;
-			if buf.is_empty() || i > skip + limit {
-				break;
-			} else if Self::is_binary(&buf, &mut inspected) {
+		while reader.read_until(b'\n', &mut buf).await.is_ok_and(|n| n > 0) {
+			if Self::is_binary(&buf, &mut inspected) {
 				return Err("Binary file".into());
 			}
 
@@ -72,17 +69,27 @@ impl Highlighter {
 				buf.push(b'\n');
 			}
 
-			if i > skip {
+			i += if i >= skip {
 				buf.iter_mut().for_each(Self::carriage_return_to_line_feed);
 				after.push(String::from_utf8_lossy(&buf).into_owned());
+				Self::line_height(&after[after.len() - 1], area.width)
 			} else if !plain {
 				before.push(String::from_utf8_lossy(&buf).into_owned());
-			}
+				Self::line_height(&before[before.len() - 1], area.width)
+			} else if PREVIEW.wrap == PreviewWrap::Yes {
+				Self::line_height(&String::from_utf8_lossy(&buf), area.width)
+			} else {
+				1
+			};
+
 			buf.clear();
+			if i > skip + area.height as usize {
+				break;
+			}
 		}
 
-		if skip > 0 && i < skip + limit {
-			return Err(PeekError::Exceed(i.saturating_sub(limit)));
+		if skip > 0 && i < skip + area.height as usize {
+			return Err(PeekError::Exceed(i.saturating_sub(area.height as _)));
 		}
 
 		Ok(if plain {
@@ -151,6 +158,31 @@ impl Highlighter {
 		} else {
 			false
 		}
+	}
+
+	fn line_height(s: &str, width: u16) -> usize {
+		if PREVIEW.wrap != PreviewWrap::Yes {
+			return 1;
+		}
+
+		let pad = PREVIEW
+			.tab_size
+			.checked_sub(1)
+			.map(|n| s.bytes().filter(|&b| b == b'\t').count() * n as usize)
+			.map(|n| yazi_config::preview::Preview::indent_with(n))
+			.unwrap_or_default();
+
+		let line = Line {
+			spans: vec![Span { content: pad, style: Default::default() }, Span {
+				content: Cow::Borrowed(s),
+				..Default::default()
+			}],
+			..Default::default()
+		};
+
+		ratatui::widgets::Paragraph::new(line)
+			.wrap(ratatui::widgets::Wrap { trim: false })
+			.line_count(width)
 	}
 
 	#[inline(always)]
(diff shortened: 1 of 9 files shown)


mskvsk:
Thank you for the PR! It works perfectly, the only thing I would change is hard wrap to soft wrap.

Hard wrap example:
The quick brown fox jum
ps over the lazy dog

Soft wrap example:
The quick brown fox 
jumps over the lazy dog

As an extra suggestion, the crate I have used previously provides some algorithm that tries to optimize the wrapping so it looks more even. But that's completely optional, just regular soft wrap will be enough.

Appreciate the solution!

ArtyomArtamonov:
Instead of hard wrap I tried to implement soft wrap. It tries to wrap lines on characters `,.; `. If a line does not contain any of these characters, we use the old logic - hard wrapping

mskvsk:
Works like a charm, thank you.

Please, DM me your Ethereum wallet so I could send you 50 USDT (let me know if you prefer USDC).

As far as I am concerned, the problem is solved so looking forward to @sxyazi 's review.

yazi-plugin/src/external/highlighter.rs
```diff
@@ -56,51 +59,235 @@ impl Highlighter {
...
+			return true;
+		}
+
+		if line.ends_with(b"\r\n") {
+			line.pop();
+			line.pop();
+			line.push(b'\n');
+		} else if !line.ends_with(b"\n") {
+			line.push(b'\n')
+		}
```
sxyazi, line 205: It seems like this branch isn't in the original code - can you explain why you added it?

yazi-plugin/src/external/highlighter.rs
```diff
@@ -56,51 +59,235 @@ impl Highlighter {
...
-				buf.pop();
-				buf.push(b'\n');
+		let mut lines_handled = 0;
+		for mut long_line in long_lines {
+			if long_line.is_empty() || lines_handled > skip + area.height as usize {
+				break;
+			}
+			Self::replace_tabs_with_spaces(&mut long_line, PREVIEW.tab_size as usize);
+			for line in Self::chunk_by_width(long_line, area.width as usize) {
+				let mut line = line.to_vec();
```
sxyazi, line 121:
```suggestion
			for mut line in Self::chunk_by_width(long_line, area.width as usize) {
```

This can avoid unnecessary memory reallocations.

yazi-plugin/src/external/highlighter.rs
```diff
@@ -56,51 +59,235 @@ impl Highlighter {
...
+		let text = String::from_utf8_lossy(line).into_owned();
+		if lines_handled > skip {
+			after.push(text);
+		} else if !plain {
+			before.push(text);
+		}
+		false
+	}
+
+	fn chunk_by_width(line: Vec<u8>, width: usize) -> Vec<Vec<u8>> {
```
sxyazi, line 216: Please combine all instances of `String::from_utf8` and `String::from_utf8_lossy` in this function into a single one, as it is a time-consuming operation.

yazi-plugin/src/external/highlighter.rs
```diff
@@ -56,51 +59,235 @@ impl Highlighter {
...
+					match buf_line_width.cmp(&width) {
+						std::cmp::Ordering::Equal => {
+							resulted_lines.push(buf_line.clone());
+							buf_line.clear();
+							last_break_idx = i + 1;
+						}
+						std::cmp::Ordering::Greater => {
+							let take_bytes = buf_line.len() - buf_char.len();
+							buf_line = buf_line[..take_bytes].to_vec();
+							resulted_lines.push(buf_line.clone());
```
sxyazi, line 254: Please remove all `.clone()` and use `mem::take()` instead.

yazi-plugin/src/external/highlighter.rs
sxyazi, line 216:
```suggestion
	fn chunk_by_width(line: Vec<u8>, width: usize) -> Vec<String> {
```

Perhaps `String` is a better return type as we need to convert `Vec<u8>` to `String` anyway in order to break the content.

yazi-plugin/src/external/highlighter.rs
ArtyomArtamonov: Every line in original code had \n at the end, but if we want to wrap long line, we have to place \n at the end of each piece of smaller line. There should not be any issues with old solution because all lines there already have \n at the end

yazi-plugin/src/external/highlighter.rs
ArtyomArtamonov: Done

yazi-plugin/src/external/highlighter.rs
ArtyomArtamonov: I tried to optimize this solution a bit by using two buffers: one is string and second is Vec<char>. First one is used only to calculate width every iteration. Second is used to take slices when slicing a line

yazi-plugin/src/external/highlighter.rs
ArtyomArtamonov: Done

yazi-plugin/src/external/highlighter.rs
ArtyomArtamonov: Changed return type to Vec<String>.

If you see there is a big optimization opportunity which I don't see, please feel free to commit, I don't mind

yazi-plugin/src/external/highlighter.rs
```diff
@@ -56,51 +59,238 @@ impl Highlighter {
...
+		plain: &mut bool,
+		mut reader: BufReader<File>,
+	) -> Result<(Vec<String>, Vec<String>), PeekError> {
+		let mut before = Vec::with_capacity(if *plain { 0 } else { skip });
+		let mut after = Vec::with_capacity(area.height as usize);
+
+		let mut long_lines = vec![];
 		let mut buf = vec![];
+		// If we want to indent plain text, we have to decide if it is plain
 		while reader.read_until(b'\n', &mut buf).await.is_ok() {
```
sxyazi, line 103: Why do we need both `while` and `for` loops here? It seems like `while` will keep reading until the end of the file without stopping early, I think we should exit the loop as soon as we have enough lines for preview. 

Is it possible to move `if long_line.is_empty() || lines_handled > skip + area.height as usize {` up into the `while` loop and combine the `while` and `for` into one loop?

yazi-plugin/src/external/highlighter.rs
```diff
@@ -56,51 +59,238 @@ impl Highlighter {
...
 			}
-
-			if !plain && buf.len() > 6000 {
-				plain = true;
-				drop(mem::take(&mut before));
+			if !*plain && buf.len() > MAX_LINE_BYTES_TO_PLAINTEXT_FALLBACK {
+				*plain = true;
 			}
+			long_lines.push(mem::take(&mut buf));
+			buf.clear()
```
sxyazi, line 111:
```suggestion

```

This is unnecessary. After `mem::take`, `buf` will be initialized as a new `Vec`.

yazi-plugin/src/external/highlighter.rs
```diff
@@ -56,51 +59,238 @@ impl Highlighter {
...
+		if line.is_empty() || lines_handled > skip + limit {
+			return true;
+		}
+
+		if line.as_str().ends_with("\r\n") {
+			let mut chars = line.chars();
+			chars.next_back();
+			chars.next_back();
+			line = chars.as_str().to_string();
+		} else if !line.as_str().ends_with('\n') {
```
sxyazi, line 203:
```suggestion
		} else if !line.ends_with('\n') {
```

yazi-plugin/src/external/highlighter.rs
```diff
@@ -56,51 +59,238 @@ impl Highlighter {
...
+	) -> bool {
+		if line.is_empty() || lines_handled > skip + limit {
+			return true;
+		}
+
+		if line.as_str().ends_with("\r\n") {
+			let mut chars = line.chars();
+			chars.next_back();
+			chars.next_back();
+			line = chars.as_str().to_string();
```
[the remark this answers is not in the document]
sxyazi: i.e.

```
		if line.ends_with("\r\n") {
			line.pop();
			line.pop();
			line.push('\n');

```

yazi-plugin/src/external/highlighter.rs
ArtyomArtamonov: At first I thought we have to go line by line twice, but now it seems it's possible to do it once. I tried to refactor code according to this. Now we traverse a file once, like before

yazi-plugin/src/external/highlighter.rs
ArtyomArtamonov: done

yazi-plugin/src/external/highlighter.rs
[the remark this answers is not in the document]
ArtyomArtamonov: Fixed

yazi-plugin/src/external/highlighter.rs
ArtyomArtamonov: also fixed this

sxyazi (approved):
Thanks!

plejd-sebman:
It seems that the schema for yazi.toml files haven't been updated with the new wrap option yet. I've tried to find the schema file in this repo but couldn't find it. Where could I find it so that I could create an PR?

sxyazi:
@plejd-sebman here you go https://github.com/yazi-rs/schemas

PR #1167, opened 2024-06-18
feat: new `confirm` component

thelamb:
https://github.com/user-attachments/assets/ac08bd15-8335-44cd-9bb9-678df237bdc0

feat: introduce confirm popup
    
    Confirm popup that takes a '\n' separated string message renders
    a popup requesting user for confirmation
    
    The message is displayed as a Paragraph with vertical scrollbar (if
    needed)
    
      - Lua binding: ya.confrm(title, message, position)
      - Position::try_from now parses "h" value (default 3)


thelamb:
Thanks for your changes @sxyazi, they all make sense.

I think we can still improve the quit() scenario when there are running tasks. Now the confirm popup only shows after 3 seconds but there is no feedback to the user about this. 

Do you see a way to get the same behavior using the select! as with the Input popup? If not I suggest we at least show a Notify to let the user know that something is happening.

sxyazi (approved):
Thanks!

cbr9:
Could this be used for the bulk rename confirmation?

Issue #1181, opened 2024-06-21
Filtered view should have visual notification that there is a filter applied


VKondakoff:
### Please describe the problem you're trying to solve

Currently there is no way to say if you see all files in folder or the view was filtered. Can you guess was this folder view filtered or not?

<img width="466" alt="Снимок экрана 2024-06-21 в 17 44 55" src="https://github.com/sxyazi/yazi/assets/230824/aa18724f-3c7e-4a57-b190-45ba8655127a">

This is especially important as the filtered view is persistent (so, you can leave the folder and return later and the filter will still be applied).

### Would you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

There is a quite good notification for search modes (fd and ripgrep):

<img width="466" alt="Снимок экрана 2024-06-21 в 17 43 37" src="https://github.com/sxyazi/yazi/assets/230824/d2ecb318-b0fb-4ae1-9af6-341214c6167c">

<img width="466" alt="Снимок экрана 2024-06-21 в 17 44 22" src="https://github.com/sxyazi/yazi/assets/230824/e2e3cbb1-9250-47f0-a37a-0bce22081f98">

Pretty sure the filtered view need something like this.

<img width="466" alt="Снимок экрана 2024-06-21 в 17 44 55" src="https://github.com/sxyazi/yazi/assets/230824/67deb33f-06b9-4c22-8658-efb1e9a86c05">

VKondakoff:
Ah... And there is an interesting situation, where both search and filter are applied. So, probably, notification should look something like: `<folder path and name> (search: xxx filter: yyy)`.

thelamb:
Request makes sense to me to show the filter as we do with search. Opened PR for this

PR #1182, opened 2024-06-22
feat: include file filter state in the header

thelamb:
Header now shows filter when it is applied, as it did with search.

Closes https://github.com/sxyazi/yazi/issues/1181


sxyazi (approved):
Thanks!

Issue #1171, opened 2024-06-19
No sensitivity to the case when renaming files in editor


mskvsk:
### What system are you running Yazi on?

macOS

### What terminal are you running Yazi in?

kitty 0.34.1

### Did you try the latest code to see if this problem got fixed?

Tried, but the problem still

### `yazi --debug` output

<details>

```sh

Yazi
    Version: 0.2.5 (VERGEN_IDEMPOTENT_OUTPUT 2024-06-19)
    Debug  : false
    OS     : macos-aarch64 (unix)

Ya
    Version: No such file or directory (os error 2)

Emulator
    Emulator.via_env: ("xterm-kitty", "")
    Emulator.via_csi: Ok(Kitty)
    Emulator.detect : Kitty

Adapter
    Adapter.matches: Kitty

Desktop
    XDG_SESSION_TYPE: None
    WAYLAND_DISPLAY : None
    DISPLAY         : None

SSH
    shared.in_ssh_connection: false

WSL
    /proc/sys/fs/binfmt_misc/WSLInterop: false

Variables
    SHELL              : Some("/bin/zsh")
    EDITOR             : Some("nvim")
    ZELLIJ_SESSION_NAME: None
    YAZI_FILE_ONE      : None
    YAZI_CONFIG_HOME   : None

Text Opener
    default: Some(Opener { run: "${EDITOR:=vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block  : Some(Opener { run: "${EDITOR:=vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })

tmux
    TMUX   : false
    Version: No such file or directory (os error 2)

Dependencies
    file             : 5.41
    ueberzugpp       : No such file or directory (os error 2)
    ffmpegthumbnailer: 2.2.2
    magick           : No such file or directory (os error 2)
    fzf              : 0.52.1
    fd               : 10.1.0
    rg               : 14.1.0
    chafa            : No such file or directory (os error 2)
    zoxide           : No such file or directory (os error 2)
    unar             : 1.10.7
    jq               : No such file or directory (os error 2)
```

</details>

### Describe the bug

When renaming a file or files in an editor and changing just the case of some characters, Yazi acts as if it isn't sensitive to the case. So if trying to change a.txt to A.txt it produces the following error:
`Failed to rename:
a.txt -> A.txt: Destination already exists`

### Expected Behavior

[...]

PR #1185, opened 2024-06-22
fix: suppress warnings for different name representations of the same file in the case-insensitive file system when renaming

Xerxes-2:
Original issue here: https://github.com/sxyazi/yazi/issues/1171

Fixes https://github.com/sxyazi/yazi/issues/1171

This PR introduced a function `are_paths_equal` in `fns.rs` to determine if two paths points to the same node.
It fixes the rename and bulk_rename functions not working properly when only change the case of letters.
Previously these functions only check if the new path already exists, but not considering the old and new path could be the same node in case-insensitive systems like macOS and Windows.


yazi-shared/src/fs/fns.rs
```diff
@@ -1,4 +1,4 @@
-use std::{borrow::Cow, collections::{HashMap, VecDeque}, ffi::{OsStr, OsString}, fs::Metadata, path::{Path, PathBuf}};
+use std::{borrow::Cow, collections::{HashMap, VecDeque}, ffi::{OsStr, OsString}, fs::Metadata, os::unix::fs::MetadataExt, path::{Path, PathBuf}};
```
lxl66566, line 1: unresolved import on windows, please move it to cfg(unix)

lxl66566:
There's an unexpected behavior: when i rename a file to it's hardlink, it will overwrite (same as being deleted) without asking.
Before this PR, yazi would ask whether to Overwrite an existing file.

reproduce:
```
touch 1
ln 1 2
./yazi
(rename 2 to 1)
```

Xerxes-2:
> There's an unexpected behavior: when i rename a file to it's hardlink, it will overwrite (same as being deleted) without asking. Before this PR, yazi would ask whether to Overwrite an existing file.
> 
> reproduce:
> 
> ```
> touch 1
> ln 1 2
> ./yazi
> (rename 2 to 1)
> ```

I think new commit have resolved this issue, but in a slight dumb way.

sxyazi:
Thanks for the PR, I'll take care of the follow-up work on this.

sxyazi (approved):
Thanks!

Issue #1186, opened 2024-06-23
DDS `hover` event doesn't get triggered after entering a directory


samsze0:
### What system are you running Yazi on?

macOS

### What terminal are you running Yazi in?

kitty 0.35.1

### Did you try the latest code to see if this problem got fixed?

Tried, but the problem still

### `yazi --debug` output

<details>

```sh

Yazi
    Version: 0.2.5 (0f84717 2024-06-20)
    Debug  : false
    OS     : macos-aarch64 (unix)

Ya
    Version: 0.2.5

Emulator
    Emulator.via_env: ("xterm-kitty", "")
    Emulator.via_csi: Ok(Kitty)
    Emulator.detect : Kitty

Adapter
    Adapter.matches: Kitty

Desktop
    XDG_SESSION_TYPE: None
    WAYLAND_DISPLAY : None
    DISPLAY         : None

SSH
    shared.in_ssh_connection: false

WSL
    /proc/sys/fs/binfmt_misc/WSLInterop: false

Variables
    SHELL              : Some("/opt/homebrew/bin/zsh")
    EDITOR             : Some("nvim")
    ZELLIJ_SESSION_NAME: None
    YAZI_FILE_ONE      : None
    YAZI_CONFIG_HOME   : None

Text Opener
    default: Some(Opener { run: "${EDITOR:=vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block  : Some(Opener { run: "${EDITOR:=vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })

tmux
    TMUX   : false
    Version: No such file or directory (os error 2)

Dependencies
    file             : 5.41
    ueberzugpp       : No such file or directory (os error 2)
    ffmpegthumbnailer: No such file or directory (os error 2)
    magick           : 7.1.1-33
    fzf              : 0.53.0
    fd               : 10.1.0
    rg               : 14.1.0
    chafa            : 1.14.0
    zoxide           : 0.9.4
    unar             : No such file or directory (os error 2)
    jq               : No such file or directory (os error 2)


--------------------------------------------------
When reporting a bug, please also upload the `yazi.log` log file - only upload the most recent content by time.
You can find it in the "/Users/sam/.local/state/yazi" directory.
```

</details>

### Describe the bug

the `hover` event doesn't get triggered after entering a directory

### Expected Behavior

[...]

PR #1187, opened 2024-06-23
fix: accommodate all `hover` events for DDS

Xerxes-2:
Original issue here: https://github.com/sxyazi/yazi/issues/1186

Fixed the issue by add `pub_from_hover` in `cd` command.


sxyazi (approved):
Thank you!

PR #1191, opened 2024-06-23
feat: add `wv` (WavPack) extension

xfzv:


sxyazi:
Hi, thanks for the PR! 

The icons for Yazi comes from https://github.com/nvim-tree/nvim-web-devicons and is automatically generated through https://github.com/sxyazi/yazi/blob/main/scripts/icons/generate.lua script.

So it's best to submit this icon upstream so it will be easier to update.

PR #1238, opened 2024-06-30
feat: ownership linemode

AidanV:
Allowed uid and gid to be called on windows. When uid or gid is called on Windows it will return nil. This is similar to how "permissions" works.

This is a continuation of #932 I decided to not support ownership on Windows and instead have the same behaviour as permissions, which is to have the keybind do nothing.

Additionally added #[allow(unused_variables)] to permissions in order to suppress compiler warnings for unused variable "me" when compiling on Windows.


sxyazi (approved):
Thanks

Issue #1258, opened 2024-07-04
Command for creating hard links


Ape:
### Please describe the problem you're trying to solve

I'd like to have a keymappable command for creating a hard link of the yanked file. In case of multiple files or directories, I'd like it to create the directories and hard links for all of them in a tree structure.

### Would you be willing to contribute this feature?

- [x] Yes, I'll give it a shot

### Describe the solution you'd like

`hardlink` command would be similar to `link`.

sxyazi:
Nice to have, maybe we can add a `--hard` to the existing `link` command, i.e. `link --hard`.

Anyone want to give it a go? We need to add the `hard` argument to the `link` command here:

https://github.com/sxyazi/yazi/blob/11547eefe0346006a1a82455577784a34d67c9b7/yazi-core/src/manager/commands/link.rs#L5-L8

And this should be the only place we need to change in the scheduler:

https://github.com/sxyazi/yazi/blob/11547eefe0346006a1a82455577784a34d67c9b7/yazi-scheduler/src/file/file.rs#L84-L95

Ape:
I have a work-in-progress patch here: https://github.com/sxyazi/yazi/pull/1268

I still want to add support for directories.

Ape:
The pull request is now ready for review.

I also added an optional second commit for adjusting the default keymap. This is how I would configure the keys, but I understand that my opinion might differ from other people.

sxyazi:
Done in https://github.com/sxyazi/yazi/pull/1268

PR #1268, opened 2024-07-06
feat: new command `hardlink`

Ape:
Implementation for https://github.com/sxyazi/yazi/issues/1258.

~~I hope adding `walkdir` as a dependency is okay.~~


Ape:
I have only tested this on Linux, but my understanding is that it should work regardless of the OS.

yazi-scheduler/Cargo.toml
```diff
@@ -22,6 +22,7 @@ futures                = "0.3.30"
 parking_lot            = "0.12.3"
 scopeguard             = "1.2.0"
 tokio                  = { version = "1.38.0", features = [ "full" ] }
+walkdir                = "2.5.0"
```
sxyazi, line 25: Please remove the `walkdir` crate, I don't want to introduce new dependencies, and `walkdir` uses sync I/O instead of async I/O.

sxyazi (changes requested):

yazi-scheduler/Cargo.toml
Ape: Done!

yazi-scheduler/src/file/file.rs
```diff
@@ -81,16 +81,25 @@ impl File {
...
-				{
-					fs::symlink(src, &task.to).await?
-				}
-				#[cfg(windows)]
-				{
-					if meta.is_dir() {
-						fs::symlink_dir(src, &task.to).await?
+
+				if task.hard {
+					if !src.is_symlink() && src.is_dir() {
```
Ape, line 86: I added `!src.is_symlink()` here to cover the special case where the yanked file is a symlink to a directory. With this version, a hardlink of the symlink is created instead of following the original symlink and hardlinking the destination.

Regular directories and symlinks to non-directory files were working correctly even before this change.

sxyazi:
I've thought about it, and it's better to implement it as a new command `hardlink`, because I found some logic in `link` is redundant (like getting the relative path, which doesn't apply to hardlinks). 

I'll finish the remaining work on this PR.

sxyazi (approved):

sxyazi:
I did some tests and it looks great even when hardlinking to a broken softlink (points to a file don't exist). I think we can merge it now.

PR #1275, opened 2024-07-08
feat: show files with inaccessible metadata

Ape:
Some broken files cause `tokio::fs::DirEntry::metadata` to fail. The previous version would just hide those broken files with no way of viewing them. This commit shows those files, but marked as broken. File type information about being a directory or a symlink is used if available.

For example a broken FUSE mount looks like this in `ls`:
```
ls: cannot access 'failed_fuse': Transport endpoint is not connected
total 8.0K
drwxr-xr-x 4 ape ape 120 Jul  8 19:28 ./
drwx------ 3 ape ape  60 Jul  8 19:26 ../
d????????? ? ?   ?     ?            ? failed_fuse/
drwxr-xr-x 2 ape ape  40 Jul  8 19:26 regular_dir/
-rw-r--r-- 1 ape ape   2 Jul  8 19:28 regular_file1
-rw-r--r-- 1 ape ape   2 Jul  8 19:28 regular_file2
```

This is now shown in yazi like this:
yazi_failed_metadata


Ape:
We could also show the failure reason in yazi. For example "Transport endpoint is not connected" in my example above. That message is available in the result on `DirEntry::metadata`, but I don't know where to show it in the UI.

yazi-shared/src/fs/file.rs
```diff
@@ -62,6 +62,23 @@ impl File {
 		Self { url, cha: Cha::from(meta).with_kind(ck), link_to, icon: Default::default() }
 	}
 
+	pub async fn from_no_meta(url: Url, file_type: Option<FileType>) -> Self {
+		// Failed to get metadata for this file, so marking it broken
+		let mut ck = ChaKind::ORPHAN;
```
sxyazi, line 67: We can't use `ORPHAN` to mark it as a broken file, because `ORPHAN` indicates a normal link file whose file linked to doesn't exist, not a file that itself is inaccessible for metadata (but exists and is a non-link file).

yazi-core/src/folder/files.rs
```diff
@@ -76,8 +82,12 @@ impl Files {
 		async fn go(entities: &[DirEntry]) -> Vec<File> {
 			let mut files = Vec::with_capacity(entities.len() / 3 + 1);
 			for entry in entities {
-				if let Ok(meta) = entry.metadata().await {
-					files.push(File::from_meta(Url::from(entry.path()), meta).await);
+				match entry.metadata().await {
+					Ok(meta) => files.push(File::from_meta(Url::from(entry.path()), meta).await),
+					Err(_) => {
+						let file_type = entry.file_type().await.ok();
```
sxyazi, line 88: I'm wondering if this is necessary - if a file's metadata can't be got, can we just get its filetype? I've never seen a broken file, so I doubt its necessity.

yazi-plugin/preset/ya.lua
```diff
@@ -50,3 +50,14 @@ function ya.readable_path(path)
...
 	end
 end
+
+function ya.owner(cha)
+	if not cha.uid or not cha.gid then
+		return nil
+	end
+
+	user = ya.user_name(cha.uid) or cha.uid
+	group = ya.group_name(cha.gid) or cha.gid
```
Ape, line 60: I wanted to write this helper function to handle possibly missing `uid` and `gid` values. It incidentally fixes a bug in Yazi where the linemode would fail if `uid` was available, but the user name was not.

yazi-shared/src/fs/cha.rs
```diff
@@ -90,12 +92,52 @@ impl From<Metadata> for Cha {
...
+	fn from(file_type: FileType) -> Self {
+		let mut kind = ChaKind::empty();
+		let mut permissions = 0;
+
+		if file_type.is_dir() {
+			kind |= ChaKind::DIR;
+
+			#[cfg(unix)]
+			{
+				permissions |= libc::S_IFDIR;
```
sxyazi, line 105: Why does the permission here only consider `libc::S_IFDIR`? What about the others?

https://github.com/sxyazi/yazi/blob/afa5936bf5102865ef9d9456823ede8e24ecc42e/yazi-shared/src/fs/fns.rs#L271-L280

yazi-plugin/src/utils/user.rs
```diff
@@ -17,19 +17,31 @@ impl Utils {
...
+					.get_user_by_uid(user_id)
 					.map(|s| lua.create_string(s.name().as_encoded_bytes()))
 					.transpose()
 			})?,
 		)?;
 
 		ya.raw_set(
 			"group_name",
-			lua.create_function(|lua, gid: Option<u32>| {
+			lua.create_function(|lua, gid: Option<Option<u32>>| {
```
sxyazi, line 36: In what case would the value passed from Lua be `Option<Option<u32>>`? If you call it with `ya.group_name(nil)`, it would only be an `Option<u32>`.

Ape: See the match clause right after this line. I did it like this to not break the Lua API: 'group_name()' still returns the name of the current user. However, 'group_name(nil)' now returns 'nil'.

'Option<Option<>>' is how mlua differentiates 'nil' with no argument.

yazi-shared/src/fs/cha.rs
Ape: Thanks for pointing that out! I first thought 'FileType' didn't have that info, but it is there in 'FileTypeExt'.

I will add all the types.

Ape: Done!

yazi-plugin/src/utils/user.rs
sxyazi: This looks really confusing because in Lua, passing 0 arguments and passing one argument as `nil` both result in the parameter value being `nil`:

```lua
function foo(bar)
  print(bar)
end

foo()
foo(nil)
```

Ouptup:

```
nil
nil
```

Ape: Actually this change is not needed anymore since the current version of `ya.owner(cha)` checks if `uid` is `nil`.

There is a caveat that `ya.user_name(cha.uid)` might now return `"root"` (or the user for `uid 0`) if `cha.uid` is `nil`. The file might not be actually owned by `root`. With that `Option<Option<>>` code the return value would be `nil` in this case.

However, the Lua code I have for `folder.lua` / `ya.lua` is doing the necessary checks currently, so I would be happy to merge my current commit with this `Option<Option<>>` change removed. The caveat only applies to potential user plugins.

Incidentally, Lua is able to differentiate a `nil` and no arguments: https://stackoverflow.com/a/12625444

sxyazi: Thanks for the change!

> There is a caveat that ya.user_name(cha.uid) might now return "root" (or the user for uid 0) if cha.uid is nil. 

If `cha.uid` is `nil`, the expected behavior is for `ya.user_name(cha.uid)` to return the current user (not the user with `uid 0`), did you find any inconsistency with the current implementation? If so we need to fix it. 

> Lua is able to differentiate a nil and no arguments

Yeah you're right, I just want to avoid this as much as possible because it requires the caller to explicitly distinguish between `ya.user_name()` and `ya.user_name(nil)`, I find it often confusing.

Ape: Yeah sorry, it returns the current user instead of `root`.

sxyazi (approved):

PR #1279, opened 2024-07-08
feat: add `nlink` property to the `Cha` plugin API

Ape:
Provides `nlink` i.e. the reference count for links on unix.

This could maybe also be shown in the default status line. As this is pretty essential for me personally, so I will be adding it to my custom status line anyway.


sxyazi (approved):
LGTM, thanks!

PR #1281, opened 2024-07-10
feat: add `title_format` manager config option

chriszarate:
This PR introduces an `update_title` manager config option, which controls whether Yazi updates the terminal title using ANSI escape sequences. For some users, this behavior is undesirable since:

1. Yazi may be invoked within another process (like Vim). If Yazi updates the terminal title, it overrides the title set by / for the parent process.
2. Users may have their own mechanisms or preferences for setting the terminal title and may not want it overridden.

Apologies that this feature request is a PR and not an issue, but I was eager to experiment with Rust. :) Happy to have it closed in favor of another implementation.

Closes https://github.com/sxyazi/yazi/issues/1350

### DIFF
--- a/yazi-config/preset/yazi.toml
+++ b/yazi-config/preset/yazi.toml
@@ -14,7 +14,7 @@ show_hidden    = false
 show_symlink   = true
 scrolloff      = 5
 mouse_events   = [ "click", "scroll" ]
-update_title   = true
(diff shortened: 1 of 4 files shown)


SmartFinn:
It's really nice to have. I just faced with this PR because I was looking how to disable the title update.

sxyazi (approved):
Thanks!

PR #1290, opened 2024-07-13
allow config fullmatch for 'which'

fuwt:
when I call 
``` lua
ya.which { cands = {
  { on = {"w","k"}, desc = "/xxx/works"},
}}
```
I might want to get a return value only after pressing "w" and "k", rather than when pressing "d". Because the of only on match, so I added an 'emit_unique' option to configure this behavior.


sxyazi:
Thanks for the PR!

How about not auto-submitting when `silent=true`? I think it only makes sense when `silent` is set, because if the user can see the remaining candidates, they should have some expectations.

fuwt:
> Thanks for the PR!
> 
> How about not auto-submitting when `silent=true`? I think it only makes sense when `silent` is set, because if the user can see the remaining candidates, they should have some expectations.

actually, I think it should never auto-submit if user is not complete input. user can easyly see the whole cand and input it quickly, but it is difficult to distinguish whether options starting with some letter are unique

PR #1291, opened 2024-07-13
feat: add `random` option to the `sort` command

zooeywm:
This PR introduces `random` option to the `sort` command. To avoid including external crate, I use the [Linear congruential generator](https://en.wikipedia.org/wiki/Linear_congruential_generator), using the [Musl](https://en.wikipedia.org/wiki/Musl) source LCG parameters.

This option is also compatible with `sort_dir_first` and `reverse`


sxyazi (approved):
LGTM, thanks!

Issue #1292, opened 2024-07-13
Add `rust-toolchain.toml` config file


zooeywm:
### Please describe the problem you're trying to solve

The yazi terminal file manager is the best one I've ever seen. But the first time when I try to contribute, I have to manually switch channel to nightly.

And [there is one deprecated fmt option](https://rust-lang.github.io/rustfmt/?version=master&search=#hide_parse_errors), which recommended by using `show_parse_errors` instead of `hide_parse_errors`

### Would you be willing to contribute this feature?

- [X] Yes, I'll give it a shot

### Describe the solution you'd like

Add a `rust-toolchain.toml` in the project's root directory to automatically switch channel.

zooeywm:
I attempted it in #1293.

sxyazi:
Closed in https://github.com/sxyazi/yazi/pull/1293

PR #1293, opened 2024-07-13
feat: add `rust-toolchain.toml`

zooeywm:
Introduce #1292


rust-toolchain.toml
```diff
@@ -0,0 +1,2 @@
+[toolchain]
+channel = "nightly"
```
sxyazi, line 2: Will this affect [users building Yazi](https://yazi-rs.github.io/docs/installation#build-from-source)? Generally users should use the latest stable version of Rust to build it, not the nightly version.

zooeywm: sorry I didn't see this! I just found the rustfmt's nightly features doesn't work on stable channel, so I subconsciously think we are using nightly version of Rust. But now when I run format from editor, it won't add the +nightly option, so the format will be broken. I should find if there is any better way to fix this from my editor config or something else.

sxyazi: No problem! In that case I'm going to close this PR, @zooeywm thanks for the effort!

PR #1304, opened 2024-07-17
feat: help users migrate shell key bindings to v0.3

mikavilpas:
I hit an issue in my config and found the existing migration error messages slightly confusing.

My issue was that I had key bindings using `--interactive` and `--block` at the same time, which seems to be disallowed in v0.3.

https://github.com/mikavilpas/dotfiles/commit/d2c8040760e49bf74d29c62390f4ca8be1b488a9


sxyazi (approved):
Thanks!

Issue #1273, opened 2024-07-08
Show "Loading" instead of "No items" if the directory is not empty but the items are still loading


Ape:
### Please describe the problem you're trying to solve

Currently, if you navigate to a directory in a particularly slow file system, yazi will show "No items" until it has loaded some directory entries. This looks just like an empty directory would, but happens regardless of if the directory is actually empty.

### Would you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

Yazi should show something like "Loading" instead of "No items" until it actually knows that the directory is empty, or until we have some actual directory contents to show.

sxyazi:
Should be fixed in https://github.com/sxyazi/yazi/pull/1439

PR #1311, opened 2024-07-19
feat: show IO errors for files in the UI

### LINKED ISSUE #1137: Give some hint/message when the content of a directory cannot be shown due to lack of permission

When attempting to browse or preview a directory for which you lack permission to view the contents, the directory may appear empty. It would greatly improve user experience to display a message indicating "Permission denied" in such cases.
Also, when trying to preview password-protected archive files, no message is shown. A user might wonder whether the number of files inside the archive is huge and yazi is struggling to show the content or it is password-protected.

It would greatly improve user experience to display a message indicating "Permission denied" in such cases.

### Additional context

_No response_

Ape:
Resolves #1137.

### DIFF
--- a/yazi-core/src/folder/files.rs
+++ b/yazi-core/src/folder/files.rs
@@ -46,7 +46,11 @@ impl Deref for Files {
 
 impl Files {
 	pub async fn from_dir(url: &Url) -> std::io::Result<UnboundedReceiver<File>> {
-		let mut it = fs::read_dir(url).await?;
+		let mut it = fs::read_dir(url).await.map_err(|e| {
+			FilesOp::IOErr(url.clone(), e.kind(), e.to_string()).emit();
+			e
+		})?;
+
 		let (tx, rx) = mpsc::unbounded_channel();
 
 		tokio::spawn(async move {
@@ -100,13 +104,18 @@ impl Files {
 		match fs::metadata(url).await {
 			Ok(m) if !m.is_dir() => {
 				// FIXME: use `ErrorKind::NotADirectory` instead once it gets stabilized
-				FilesOp::IOErr(url.clone(), std::io::ErrorKind::AlreadyExists).emit();
+				FilesOp::IOErr(
+					url.clone(),
+					std::io::ErrorKind::AlreadyExists,
+					"Not a directory".to_string(),
+				)
+				.emit();
 			}
 			Ok(m) if mtime == m.modified().ok() => {}
 			Ok(m) => return Some(m),
 			Err(e) => {
 				if maybe_exists(url).await {
-					FilesOp::IOErr(url.clone(), e.kind()).emit();
+					FilesOp::IOErr(url.clone(), e.kind(), e.to_string()).emit();
 				} else if let Some(p) = url.parent_url() {
 					FilesOp::Deleting(p, vec![url.clone()]).emit();
 				}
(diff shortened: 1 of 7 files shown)


Ape:
This actually accidentally partially fixes #1273, too. Turns out there was `tostring` missing in `yazi-plugin/preset/components/current.lua`.

The loading message still doesn't work in the folder preview view (`yazi-plugin/preset/plugins/folder.lua`), but that's outside the scope of this PR.

yazi-core/src/folder/files.rs
```diff
@@ -46,7 +46,11 @@ impl Deref for Files {
 
 impl Files {
 	pub async fn from_dir(url: &Url) -> std::io::Result<UnboundedReceiver<File>> {
-		let mut it = fs::read_dir(url).await?;
+		let mut it = fs::read_dir(url).await.map_err(|e| {
```
sxyazi, line 49: It's unnecessary because `Files::assert_stale` has already checked it:

https://github.com/sxyazi/yazi/blob/d6081fbe6f57c8fa4b577fb3fbe1d392c057d94e/yazi-core/src/tab/preview.rs#L57-L58

yazi-core/src/folder/stage.rs
```diff
@@ -1,7 +1,7 @@
-#[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
 pub enum FolderStage {
 	#[default]
 	Loading,
 	Loaded,
-	Failed(std::io::ErrorKind),
+	Failed(std::io::ErrorKind, String),
```
sxyazi, line 6: Why don't we generate error messages through `std::io::ErrorKind` instead of saving them as strings?

yazi-plugin/preset/components/current.lua
```diff
@@ -14,8 +14,12 @@ function Current:empty()
 	local line
 	if self._folder.files.filter then
 		line = ui.Line("No filter results")
+	elseif tostring(self._folder.stage) == "loading" then
```
sxyazi, line 17: Nice catch!

sxyazi (changes requested):

yazi-core/src/folder/files.rs
Ape: This is needed at least for "Permission denied" on a directory. `assert_stale` currently only checks for the metadata, not accessing the contents.

yazi-core/src/folder/stage.rs
sxyazi: How about using the [`Display` trait](https://doc.rust-lang.org/std/io/enum.ErrorKind.html#impl-Display-for-ErrorKind) to do this?

Ape: That doesn't give the pretty human readable messages, just the name of the kind.

sxyazi: The error messages for the two cases `NotFound` and `PermissionDenied` are "entity not found" and "permission denied", which I think are quite readable. What does the current implementation look like?

Ape: The error strings in the current commit are the classic C / POSIX error messages with the os error code, e.g. "No such file or directory (os error 2)" or "Permission denied (os error 13)". I wanted to use these since they seem familiar and are easy to search details for especially when you encounter a rarer type of error that you might not understand.

However, I guess the ErrorKind strings are actually good enough and it makes the implementation simpler.

sxyazi: > However, I guess the ErrorKind strings are actually good enough and it makes the implementation simpler.

Nice, please change it! Let's implement it in the simplest way for now, if someone reports that it's not clear enough, we can create our own messages by `ErrorKind`.

sxyazi:
I'm going to close this because it still needs some work and hasn't seen any activity recently. @Ape thanks for the effort!

Issue #1316, opened 2024-07-20
[MacOS] `yazi` crashes with `reader source not set` when using `xargs`


joshuali925:
### What system are you running Yazi on?

macOS

### What terminal are you running Yazi in?

wezterm 20240212-073805-217dfb30

### `yazi --debug` output

<details>

```sh
##### ↓↓↓ Paste the output here: ↓↓↓ #####
❯ yazi --debug

Yazi
    Version: 0.2.5 (Homebrew 2024-07-16)
    Debug  : false
    OS     : macos-aarch64 (unix)

Ya
    Version: 0.2.5

Emulator
    Emulator.via_env: ("xterm-256color", "WezTerm")
    Emulator.via_csi: Ok(WezTerm)
    Emulator.detect : WezTerm

Adapter
    Adapter.matches: Iterm2

Desktop
    XDG_SESSION_TYPE: None
    WAYLAND_DISPLAY : None
    DISPLAY         : None

SSH
    shared.in_ssh_connection: false

WSL
    /proc/sys/fs/binfmt_misc/WSLInterop: false

Variables
    SHELL              : Some("/bin/zsh")
    EDITOR             : Some("nvim")
    ZELLIJ_SESSION_NAME: None
    YAZI_FILE_ONE      : None
    YAZI_CONFIG_HOME   : None

Text Opener
    default: Some(Opener { run: "${EDITOR:=vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block  : Some(Opener { run: "${EDITOR:=vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })

tmux
    TMUX   : false
    Version: 3.4

Dependencies
    file             : 5.41
    ueberzugpp       : No such file or directory (os error 2)
    ffmpegthumbnailer: No such file or directory (os error 2)
    magick           : No such file or directory (os error 2)
    fzf              : 0.54.0
    fd               : 8.7.0
    rg               : 14.0.3
    chafa            : No such file or directory (os error 2)
    zoxide           : No such file or directory (os error 2)
    unar             : No such file or directory (os error 2)
    jq               : 1.7


--------------------------------------------------
When reporting a bug, please also upload the `yazi.log` log file - only upload the most recent content by time.
You can find it in the "/Users/user/.local/state/yazi" directory.

```

yazi.log is empty when it crashes

</details>

### Did you try the latest nightly build to see if the problem got fixed?

[...]

PR #1317, opened 2024-07-20
fix: add `use-dev-tty` to fix piping to Yazi on macOS

joshuali925:
fixes https://github.com/sxyazi/yazi/issues/1316


sxyazi (approved):
Thanks!

PR #1321, opened 2024-07-22
feat: new `extract` builtin plugin for archive extracting

sxyazi:
TODO: add description later


leihaojun:
This change seems to make the sub handle function in the plug-in no longer use the plug-in `state` variable.

for example the ya.err output always is nil ,  from this commit.

```lua
local M = {
	setup = function(st,opts)
		ps.sub("cd",function ()
			st.test = cx.active.current.cwd
		end)
	end
}

local cd_last = ya.sync(function (st)
	ya.err(st.test)
end)

```

leihaojun:
also fail like this 
```
loacl handle_cd = ya.sync(function(st)
    st.test = cx.active.current.cwd
end)

local M = {
	setup = function(st,opts)
		ps.sub("cd",handle_cd)
	end
}

local cd_last = ya.sync(function (st)
	ya.err(st.test)
end)


```

sxyazi:
Thanks for the report @leihaojun! Please try https://github.com/sxyazi/yazi/commit/1e1ae637a8cd183b1e95f8c218b5022d547df7a9

leihaojun:
Although the above problem is solved, there is a new problem, the fetcher will be executed in an endless loop, even if nothing is done.
```
function M:fetch()
	ya.err("fetch")
	return 3	
end

```

image

sxyazi:
That commit doesn't include any changes to fetch, so I don't think it's relevant. 

Also, you should return 1 to indicate "Success, don't continue" instead of 3 for "Success, continue". This way, fetch will keep running, which is the expected behavior.

See https://yazi-rs.github.io/docs/plugins/overview#preloader

leihaojun:
I don't quite understand what this continuation means. If I return 1, it seems that the fetch will only be monitored once, and then it will not be executed regardless of whether the file changes, but I can continue the follow-up monitoring when I return 3, which is what I want.


Before the last commit, fetch would execute fetch only when the current file changed, and after this commit, fetch would continue to execute even if there was no file change.

In addition, it is not always possible to execute fetch, but the current path has at least one folder, and this is only possible if the fetch rule of the directory is enabled.
```toml
{ id = "test", name = "*/", run = "test",if = "mime", prio = "normal" },
```

leihaojun:
image

It was normal at the time of this submission

sxyazi:
I'm not quite sure I understand what you mean - do you want the fetcher to be triggered continuously or just once? If it's the former, then returning 3 is the expected behavior. If it's the latter, you should return 1.

leihaojun:
i mean that It can be triggered at any time when there is a file change

But now it keeps getting called when the file isn't changing

leihaojun:
If I return 1, I can't listen for the next file change

sxyazi:
Should be fixed in https://github.com/sxyazi/yazi/commit/23a56c24a4c15807bc74dc904da10ed4cb5c3e6e

leihaojun:
> Should be fixed in [23a56c2](https://github.com/sxyazi/yazi/commit/23a56c24a4c15807bc74dc904da10ed4cb5c3e6e)

Confirm that it have fixed the issue with the  endless loop fetch call.

leihaojun:
However, I found another strange problem, when the path rule of fetch is enabled,

```toml
{ id = "test", name = "*/", run = "test",if = "mime", prio = "normal" },
```

 if a file is deleted in a directory that has both folders and files, it will call fetch once. When I delete the folder and then delete the files, fetch will no longer be called.

I don't quite understand the logic of this design, or is it a bug?

leihaojun:
When I use these two rules at the same time and find that there is at least one folder in the current path, fetch will be called once to modify the file. Once there is no folder in the current path, fetch will not be triggered when I modify the file.

```

{ id = "test", name = "*/", run = "test",if = "mime", prio = "normal" },
{ id = "test", name = "*", run = "test",if = "!mime", prio = "normal" },

```

leihaojun:
i change the fetch config to this
```

{ id = "test", name = "*/", run = "test",if = "mime", prio = "normal" },
{ id = "test", name = "*", run = "test",if = "!mime", prio = "normal" },


```

if solved the problem above.

uncenter:
Why were the `J` and `K` cursor moving commands removed in the commit for this PR (https://github.com/sxyazi/yazi/commit/23a56c24a4c15807bc74dc904da10ed4cb5c3e6e#diff-24b902f41fbbd00060d73da09d41f1f7f29e9f81f89d5f078653473c6d40598e), but not in this PR and not mentioned in the release notes (as far as I can tell!)?

PR #1325, opened 2024-07-22
feat(yazi-cli/pack): support remotes other than github

uncenter:
Currently only GitHub plugins are supported, but with authors like [DreamMaoMao having to use Gitee](https://gitee.com/DreamMaoMao) or for other reasons supporting other Git hosts would be nice. This PR maintains the same behavior as before, but in addition if a URL is passed to `ya pack -a` the CLI will use that URL as the remote.


uncenter:
Upgrading is broken atm, let me fix...

uncenter:
Now packages with remotes other than GitHub will have a `remote` key added to the entry in `deps`, and that `remote` is restored when loading from config and used as the initializer string when creating a new `Package`.

uncenter:
Currently to add a subdirectory of a repo as a package you use the same syntax as before, e.g. `https://github.com/yazi-rs/plugins#smart-filter`. This is because I don't think we can guarantee that every Git forge will use the `/tree/main/smart-filter` URL pattern.

uncenter:
Open for too long and now conflicts are present. Not interested in resolving if this won't be merged in the near future.

Issue #1330, opened 2024-07-23
Cut files still appear on dir preview


wojciech-kulik:
### What system are you running Yazi on?

macOS

### What terminal are you running Yazi in?

kitty 0.35.2

### `yazi --debug` output

<details>

```sh
##### ↓↓↓ Paste the output here: ↓↓↓ #####

Yazi
    Version: 0.2.5 (Homebrew 2024-04-23)
    OS: macos-aarch64 (unix)
    Debug: false

Emulator
    Emulator.via_env: ("xterm-kitty", "")
    Emulator.via_csi: Ok(Kitty)
    Emulator.detect: Kitty

Adaptor
    Adaptor.matches: Kitty

Desktop
    XDG_SESSION_TYPE: None
    WAYLAND_DISPLAY: None
    DISPLAY: None

SSH
    shared.in_ssh_connection: false

WSL
    /proc/sys/fs/binfmt_misc/WSLInterop: false

Variables
    SHELL: Some("/bin/zsh")
    EDITOR: Some("nvim")
    ZELLIJ_SESSION_NAME: None
    YAZI_FILE_ONE: None
    YAZI_CONFIG_HOME: None

file(1)
    Version: Ok(Output { status: ExitStatus(unix_wait_status(0)), stdout: "file-5.41\nmagic file from /usr/share/file/magic\n", stderr: "" })

Text Opener
    default: Some(Opener { run: "${EDITOR:=vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block: Some(Opener { run: "${EDITOR:=vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })

tmux
    TMUX: false

Ueberzug++
    Version: Err(Os { code: 2, kind: NotFound, message: "No such file or directory" })
```

</details>

### Did you try the latest nightly build to see if the problem got fixed?

Yes, and I updated the debug information above (`yazi --debug`) to the nightly that I tried

### Describe the bug

I'm migrating from `lf` so I'm not yet fully fluent with yazi, but I noticed a weird issue that doesn't refresh dir preview after some operations.

Environment:
macOS 14.5
kitty 0.35.2
Yazi 0.2.5 (Homebrew 2024-04-23)

### Expected Behavior

Dir preview should be refreshed and it should reflect the current state.

### To Reproduce

1. Go to sub-folder
2. Select some files and cut them
[...]

PR #1331, opened 2024-07-23
fix: wrong directory loading optimization condition

sxyazi:
Fixes https://github.com/sxyazi/yazi/issues/1330


wojciech-kulik:
I verified this fix and now it works correctly 🔥. Thank you for such a quick reaction

Issue #1313, opened 2024-07-19
After a folder is deleted once, subdirectories of this folder are not watched for deletion events


leihaojun:
### What system are you running Yazi on?

Linux X11

### What terminal are you running Yazi in?

wezterm

### `yazi --debug` output

<details>

```sh
Yazi
    Version: 0.2.5 (ed6ae00 2024-07-19)
    Debug  : false
    OS     : linux-x86_64 (unix)

Ya
    Version: No such file or directory (os error 2)

Emulator
    Emulator.via_env: ("xterm-256color", "WezTerm")
    Emulator.via_csi: Ok(WezTerm)
    Emulator.detect : WezTerm

Adapter
    Adapter.matches: Iterm2

Desktop
    XDG_SESSION_TYPE: Some("x11")
    WAYLAND_DISPLAY : None
    DISPLAY         : Some(":0")

SSH
    shared.in_ssh_connection: false

WSL
    /proc/sys/fs/binfmt_misc/WSLInterop: false

Variables
    SHELL              : Some("/bin/fish")
    EDITOR             : Some("nvim ")
    ZELLIJ_SESSION_NAME: None
    YAZI_FILE_ONE      : None
    YAZI_CONFIG_HOME   : None

Text Opener
    default: Some(Opener { run: "$EDITOR \"$1\"", block: true, orphan: false, desc: "neovim", for_: None, spread: false })
    block  : Some(Opener { run: "$EDITOR \"$1\"", block: true, orphan: false, desc: "neovim", for_: None, spread: false })

tmux
    TMUX   : false
    Version: 3.4

Dependencies
    file             : 5.45
    ueberzugpp       : No such file or directory (os error 2)
    ffmpegthumbnailer: 2.2.2
    magick           : 7.1.1-27
    fzf              : 0.42.0
    fd               : 9.0.0
    rg               : 13.0.0
    chafa            : 1.14.0
    zoxide           : 0.9.2
    unar             : ExitStatus(unix_wait_status(32512)), "unar: error while loading shared libraries: libicuuc.so.74: cannot open shared object file: No such file or directory\n"
    jq               : 1.7


--------------------------------------------------
When reporting a bug, please also upload the `yazi.log` log file - only upload the most recent content by time.

```

</details>

### Did you try the latest nightly build to see if the problem got fixed?

Yes, and I updated the debug information above (`yazi --debug`) to the nightly that I tried

### Describe the bug

[...]

PR #1335, opened 2024-07-24
fix: can't rewatch a directory that has been deleted once before

leihaojun:
Fixes https://github.com/sxyazi/yazi/issues/1313

The retrun is_ok from to_unwatch a `deleted path` is always false. 
image

But the directory has actually been unwatch, because the return is false, so it was not actively removed from the `WATCHED`, 
image

so when the same name path is created, the path recognize as a watching path in  `old` var.  
image

so the new path with the same name will not add to the `to_watch` to rewatcher.

### DIFF
--- a/yazi-core/src/manager/watcher.rs
+++ b/yazi-core/src/manager/watcher.rs
@@ -72,7 +72,14 @@ impl Watcher {
 				(old.difference(new).cloned().collect(), new.difference(old).cloned().collect())
 			};
 
-			to_unwatch.retain(|u| watcher.unwatch(u).is_ok() || !u.exists());
+			to_unwatch.retain(|u| match watcher.unwatch(u) {
+				Ok(_) => true,
+				Err(e) if matches!(e.kind, notify::ErrorKind::WatchNotFound) => true,
+				Err(e) => {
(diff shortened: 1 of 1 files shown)


sxyazi:
Thanks for the PR!

yazi-core/src/manager/watcher.rs
```diff
@@ -72,7 +72,7 @@ impl Watcher {
 				(old.difference(new).cloned().collect(), new.difference(old).cloned().collect())
 			};
 
-			to_unwatch.retain(|u| watcher.unwatch(u).is_ok());
+			to_unwatch.retain(|u| watcher.unwatch(u).is_ok() || !u.exists());
```
sxyazi, line 75: `.unwatch()` should already return an error when it fails to unwatch a non-existent directory, why don't we just use it directly?

leihaojun: you mean to analyze the cause of the error before deciding whether to remove the path?

sxyazi: Yep, I'll fix it now

sxyazi (approved):
Thanks!

PR #1360, opened 2024-07-27
Keybinding for filtering help items was not listed among those help items

MarkZaytsev:
There is a command to filter help items. By default it was bind to "/" in `keymap.toml` for `help` section.

```rust
[help]

keymap = [
...
   # Filtering
   { on = "/", run = "filter", desc = "Apply a filter for the help items" },
]
```
But it was missing for other sections, like `manager`. Thus it was never displayed. And this is a very useful feature for newcomers. 
I've added it to all sections. I also had to replace `"/"` binding with `"F"`. By default, `"/"` is used for finding in manager, while `"f"` is used for filtering. Unfortunately we can't bind filtering for help items to `"f"` because it will collide with manager's command and only one will be displayed.

I also think that filtering help items should be displayed as the first item on the list since it's very useful when looking at big list of commands like in manager's section. But it's more like follow up suggestion and not included in this PR.

### DIFF
--- a/yazi-config/preset/keymap.toml
+++ b/yazi-config/preset/keymap.toml
@@ -153,7 +153,7 @@ keymap = [
 	# Help
 	{ on = "~",		run = "help", 	desc = "Open help" },
 	{ on = "<F1>",	run = "help", 	desc = "Open help" },
-	{ on = "/",		run = "filter", desc = "Apply a filter for the help items" },
+	{ on = "F",		run = "filter", desc = "Apply a filter for the help items" },
 ]
 
 [tasks]
@@ -176,7 +176,7 @@ keymap = [
 	# Help
 	{ on = "~",		run = "help", 	desc = "Open help" },
 	{ on = "<F1>",	run = "help", 	desc = "Open help" },
-	{ on = "/",		run = "filter", desc = "Apply a filter for the help items" },
+	{ on = "F",		run = "filter", desc = "Apply a filter for the help items" },
 ]
 
 [select]
@@ -202,7 +202,7 @@ keymap = [
 	# Help
 	{ on = "~",		run = "help", 	desc = "Open help" },
 	{ on = "<F1>",	run = "help", 	desc = "Open help" },
-	{ on = "/",		run = "filter", desc = "Apply a filter for the help items" },
+	{ on = "F",		run = "filter", desc = "Apply a filter for the help items" },
 ]
 
 [input]
@@ -273,7 +273,7 @@ keymap = [
 	# Help
 	{ on = "~",		run = "help", 	desc = "Open help" },
 	{ on = "<F1>",	run = "help", 	desc = "Open help" },
-	{ on = "/",		run = "filter", desc = "Apply a filter for the help items" },
+	{ on = "F",		run = "filter", desc = "Apply a filter for the help items" },
 ]
 
 [completion]
@@ -295,7 +295,7 @@ keymap = [
 	# Help
 	{ on = "~",		run = "help", 	desc = "Open help" },
 	{ on = "<F1>",	run = "help", 	desc = "Open help" },
-	{ on = "/",		run = "filter", desc = "Apply a filter for the help items" },
+	{ on = "F",		run = "filter", desc = "Apply a filter for the help items" },
 ]
 
 [help]
@@ -320,5 +320,5 @@ keymap = [
 	{ on = "<S-Down>", run = "arrow 5",  desc = "Move cursor down 5 lines" },
 
 	# Filtering
-	{ on = "/", run = "filter", desc = "Apply a filter for the help items" },
+	{ on = "F", run = "filter", desc = "Apply a filter for the help items" },
 ]
--- a/yazi-config/preset/keymap.toml
+++ b/yazi-config/preset/keymap.toml
@@ -151,8 +151,9 @@ keymap = [
 	{ on = [ "g", "<Space>" ], run = "cd --interactive", desc = "Go to a directory interactively" },
 
 	# Help
-	{ on = "~",    run = "help", desc = "Open help" },
-	{ on = "<F1>", run = "help", desc = "Open help" },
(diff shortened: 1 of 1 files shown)


yazi-config/preset/keymap.toml
```diff
@@ -173,8 +174,9 @@ keymap = [
 	{ on = "x",       run = "cancel",  desc = "Cancel the task" },
 
 	# Help
-	{ on = "~",    run = "help", desc = "Open help" },
-	{ on = "<F1>", run = "help", desc = "Open help" },
+	{ on = "~",		run = "help", 	desc = "Open help" },
+	{ on = "<F1>",	run = "help", 	desc = "Open help" },
+	{ on = "F",		run = "filter", desc = "Apply a filter for the help items" },
```
sxyazi, line 179: `[tasks]` doesn't have a command called `filter`, https://yazi-rs.github.io/docs/configuration/keymap#tasks

yazi-config/preset/keymap.toml
```diff
@@ -198,8 +200,9 @@ keymap = [
 	{ on = "<S-Down>", run = "arrow 5",  desc = "Move cursor down 5 lines" },
 
 	# Help
-	{ on = "~",    run = "help", desc = "Open help" },
-	{ on = "<F1>", run = "help", desc = "Open help" },
+	{ on = "~",		run = "help", 	desc = "Open help" },
+	{ on = "<F1>",	run = "help", 	desc = "Open help" },
+	{ on = "F",		run = "filter", desc = "Apply a filter for the help items" },
```
sxyazi, line 205: `[select]` doesn't have a command called `filter`, https://yazi-rs.github.io/docs/configuration/keymap#select

yazi-config/preset/keymap.toml
```diff
@@ -289,8 +293,9 @@ keymap = [
 	{ on = "<C-n>", run = "arrow 1",  desc = "Move cursor down" },
 
 	# Help
-	{ on = "~",    run = "help", desc = "Open help" },
-	{ on = "<F1>", run = "help", desc = "Open help" },
+	{ on = "~",		run = "help", 	desc = "Open help" },
+	{ on = "<F1>",	run = "help", 	desc = "Open help" },
+	{ on = "F",		run = "filter", desc = "Apply a filter for the help items" },
```
sxyazi, line 298: `[completion]` doesn't have a command called `filter`, https://yazi-rs.github.io/docs/configuration/keymap#completion

sxyazi (changes requested):

yazi-config/preset/keymap.toml
```diff
@@ -151,8 +151,9 @@ keymap = [
 	{ on = [ "g", "<Space>" ], run = "cd --interactive", desc = "Go to a directory interactively" },
 
 	# Help
-	{ on = "~",    run = "help", desc = "Open help" },
-	{ on = "<F1>", run = "help", desc = "Open help" },
+	{ on = "~",		run = "help", 	desc = "Open help" },
+	{ on = "<F1>",	run = "help", 	desc = "Open help" },
+	{ on = "F",		run = "filter", desc = "Apply a filter for the help items" },
```
sxyazi, line 156: https://github.com/sxyazi/yazi/blob/f024ce03e7b97f55f3bfad9da4b33458cea765d5/yazi-config/preset/keymap.toml#L101-L102

It's already in `[manager]` section, https://yazi-rs.github.io/docs/quick-start#filter-files

MarkZaytsev: This is filter for file list. I'm talking about interactive filtering of entries in help listing

sxyazi: > in help listing

So why did you add it to `[manager]` rather than `[help]`?

yazi-config/preset/keymap.toml
MarkZaytsev: How else then can you show to user that help list can be filtered when opened from tasks content?

sxyazi: You need to press `~` in task manager to open its own help

MarkZaytsev:
I'm adding the video to avoid misunderstanding. Currently there is no way for user to know that he can use filter in help widget.

[simplescreenrecorder-2024-07-27_20.18.32.webm](https://github.com/user-attachments/assets/64cb4063-3aa8-40ba-be7b-b94ee199d3c7)

sxyazi:
That's because there's no help for the help itself, you can't see the help's own keymappings. 

When you press `~` in the task manager, you're viewing the task manager's help

sxyazi:
If we want to prompt users that they can use `/` to filter in the help, we should change the help itself instead of adding it to other components (`[tasks]`, `[manager]`, etc.).

MarkZaytsev:
> That's because there's no help for the help itself, you can't see the help's own keymappings.

Yes, I understand that yazi shows keymap help based on a widget context. 

But I see no other way how you can let a user to know that he can filter the help listing. Making a user to open a help for help is weird and judging by code base it doesn't look as intended. 

My point is that this is a very useful feature that is not documented and not listed anywhere in help keymap. If you know a better way how to implement it - please let me know.

MarkZaytsev:
I agree that putting command that is not related to the widget is not the best solution in general.

How about showing this keybinding instead of `manager.help` or `tasks.help`? 

yazi-filter

sxyazi:
Sounds great to me! And it should also be simple to achieve:

https://github.com/sxyazi/yazi/blob/f024ce03e7b97f55f3bfad9da4b33458cea765d5/yazi-fm/src/help/layout.rs#L22

We just need to append a ` (Press f to filter)` after the `.help` string here

MarkZaytsev:
In that case I will revert my changes and reimplement this way. 
I will also try to have it based on keymap binding, in case if a user would want to override the key.

MarkZaytsev:
Completely reworked in https://github.com/sxyazi/yazi/pull/1361

PR #1361, opened 2024-07-27
feat: add filtering tip in help component

MarkZaytsev:
This is a follow up to https://github.com/sxyazi/yazi/pull/1360 conversation in attempt to get users more familiar with filtering help items feature.

[simplescreenrecorder-2024-07-27_23.35.40.webm](https://github.com/user-attachments/assets/d08d35b6-3b09-40d7-8f95-ae663638bf01)

It will work correctly in case if a user overrides it with combination of keys like that:
image

By default filtering key changed to `f` for consistency with manager keymap

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -1,6 +1,7 @@
 cargo publish -p yazi-shared
 cargo publish -p yazi-config
 cargo publish -p yazi-proxy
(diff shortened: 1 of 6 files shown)


sxyazi (approved):
Thanks!

PR #1375, opened 2024-07-31
feat(nix): set `VERGEN_*` build env vars

uncenter:
Similar to my recent https://github.com/NixOS/nixpkgs/pull/330520 (but without the need for https://github.com/NixOS/nixpkgs/pull/330619 since this is entirely automatic). Changes the `yazi --version` output from the not-so-helpful `Yazi 0.2.5 (VERGEN_IDEMPOTENT_OUTPUT 1980-01-01)` to `Yazi 0.2.5 (cdb34de 2024-07-31)`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -27,15 +27,13 @@
       (system:
         let
           pkgs = import nixpkgs { inherit system overlays; };
-          versionSuffix = "pre${
-            builtins.substring 0 8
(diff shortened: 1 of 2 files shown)


sxyazi (approved):
Awesome, thanks!

Never realized that Nix build times were like this. No wonder the bug reports that didnt have build times were probably from Nix users haha :D

PR #1376, opened 2024-07-31
refactor(nix): cleanup and switch to `nixfmt-rfc-style`

uncenter:
Cleans up the code with the `statix` and `deadnix` tools, and switches the formatter to `nixfmt-rfc-style`.

PR #1380, opened 2024-08-01
refactor(nix): home keeping

isabelroses:
changes and why:
- drop flake-utils: flake-utils was removed from rust-overlay, so pulling it in only to do something that a function could do seems uneeded
- sync "yazi" to be more in sync with upstream
- add "?" to "constructed" args since the ordinary `callPackage` won't pass these args
- remove the last `rec` this is following upstreams nixpkgs pattern on this pattern
- use `prev.stdenv.system` over `final.system` since users can disable aliases, we also have no need to use the final specifically either so `prev` makes more sense
-`shell.nix` is now constructed via `yazi-unwrapped.nix` meaning all dependencies are transfered, reducing some repetion

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,13 +1,9 @@
 {
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
-    flake-utils.url = "github:numtide/flake-utils";
     rust-overlay = {
       url = "github:oxalica/rust-overlay";
-      inputs = {
-        nixpkgs.follows = "nixpkgs";
-        flake-utils.follows = "flake-utils";
-      };
+      inputs.nixpkgs.follows = "nixpkgs";
     };
   };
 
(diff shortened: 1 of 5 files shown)


isabelroses:
As such completes this https://github.com/sxyazi/yazi/pull/1376#issuecomment-2261871740

cc: @uncenter

XYenon:
Are there any benefits to rewrite a `forEachSystem` instead of use `flake-utils`?

flake.nix
```diff
@@ -1,61 +1,87 @@
...
+              inherit system;
+              pkgs = import nixpkgs {
+                inherit system;
+                overlays = [
+                  rust-overlay.overlays.default
+
+                  (
+                    final: prev:
+                    let
+                      toolchain = final.rust-bin.stable.latest.default.override { extensions = [ "rust-src" ]; };
```
XYenon, line 42: `rust-src` is only used by `rust-analyzer`, this can be moved to `shell.nix`.

isabelroses:
> Are there any benefits to rewrite a `forEachSystem` instead of use `flake-utils`?

nix clones each repo that you have to the nix store so you end up with an extra storage being taken for such limited use, that can easily be done by any user.

XYenon:
> > Are there any benefits to rewrite a `forEachSystem` instead of use `flake-utils`?
> 
> nix clones each repo that you have to the nix store so you end up with an extra storage being taken for such limited use, that can easily be done by any user.
> 
> It wouldn't be so bad if rust-overlay we're still tracking it but if its not needed its better to just remove it.

I found flake-utils is less than 90KB. I don't think increase the complexity to save 90KB is worth it.

isabelroses:
> I don't think increase the complexity 

Most of that is just from what it looks like but most of what it does you were already doing with:
https://github.com/sxyazi/yazi/blob/4257b95e8d0df4acc4d88ab1a9f4a915c3b4b9cd/flake.nix#L29
and 
https://github.com/sxyazi/yazi/blob/4257b95e8d0df4acc4d88ab1a9f4a915c3b4b9cd/flake.nix#L17-24

Just now most of that complexity is merged into one function, which may make it look worse then it actually is

isabelroses:
you could also make `devShells` shorter and save 2 lines. But this nix "api" should be *mostly* (since flakes are experimental in of themselves) unchanging so rewriting it here is not so bad.

```nix
 devShells.default = pkgs.callPackage ./nix/shell.nix { };
 ```

**Comment by XYenon** [comment]
> you could also make `devShells` shorter and save 2 lines. 

Make it shorter is not the purpose. I just want to say, reimplement a nix function in a non strongly nix related project is not a good idea.

**Comment by isabelroses** [comment]
> I just want to say, reimplement a nix function in a non strongly nix related project is not a good idea.

Normally I would agree but this project has yourself, who works with [upstream](https://github.com/NixOS/nixpkgs/blob/52ec9ac3b12395ad677e8b62106f0b98c1f8569d/pkgs/by-name/ya/yazi-unwrapped/package.nix#L57) and [@uncenter](https://github.com/orgs/yazi-rs/people). Who are both more then capable of understanding and maintaining this.

I also have a bit of a personal grievance with flake-utils, which I won't explain but will point you towards: https://ayats.org/blog/no-flake-utils. As thus I am happy to drop that change though If I cannot change your mind.

**Comment by XYenon** [comment]
> Normally I would agree but this project has yourself, who works with [upstream](https://github.com/NixOS/nixpkgs/blob/52ec9ac3b12395ad677e8b62106f0b98c1f8569d/pkgs/by-name/ya/yazi-unwrapped/package.nix#L57) and [@uncenter](https://github.com/orgs/yazi-rs/people). Who are both more then capable of understanding and maintaining this.
> 
> I also have a bit of a personal grievance with flake-utils, which I won't explain but will point you towards: [ayats.org/blog/no-flake-utils](https://ayats.org/blog/no-flake-utils). As thus I am happy to drop that change though If I cannot change your mind.

I agree with the opinion of that post. And I'll be happy to maintain a function like this in a nix project, but I don't want to maintain it in every project which uses nix. 

Change it to flake-parts is also a great option. 😂

File: nix/yazi-unwrapped.nix [hunk 2]
```diff
@@ -1,67 +1,86 @@
...
 
   imagemagick,
 }:
+let
+  src = lib.fileset.toSource {
+    root = ../.;
+    fileset = lib.fileset.intersection (lib.fileset.fromSource (lib.sources.cleanSource ../.)) (
+      lib.fileset.unions [
+        ../assets
+        ../yazi-adapter
```
XYenon, line 21: Could we add all `yazi-*`?

https://github.com/sxyazi/yazi/blob/7a380c2f0fdaec87992836e21dc8b6cd251af980/Cargo.toml#L3

uncenter:
Anything else here? Would love to get this merged.

isabelroses:
> Anything else here? Would love to get this merged.

I think the last thing to do would be to revert back to flake-utils. I do have a way to make the function simpler (at least visually) but as @XYenon said its probably not best.

uncenter (approved):
LGTM

nix/yazi-unwrapped.nix
```diff
@@ -1,67 +1,76 @@
...
 }:
+let
+  src = lib.fileset.toSource {
+    root = ../.;
+    fileset = lib.fileset.intersection (lib.fileset.fromSource (lib.sources.cleanSource ../.)) (
+      lib.fileset.unions [
+        ../assets
+        ../Cargo.toml
+        ../Cargo.lock
+        (lib.fileset.fileFilter (file: file.hasExt "rs" || file.hasExt "toml") ../.)
```
XYenon, line 23: There are also `.lua` files should be included. I think we should add all `yazi-*` directories instead of filter by extension name.

isabelroses: when testing i found no `.lua` files. But if we did want to do `yazi-*` it would look something like this.

```nix
src = lib.fileset.toSource {
    root = ../.;
    fileset = lib.fileset.unions (
      builtins.concatLists [
        [
          ../assets
          ../Cargo.toml
          ../Cargo.lock
        ]

        (map (name: ../. + "/${name}") (
          builtins.filter (name: lib.hasPrefix "yazi-" name) (builtins.attrNames (builtins.readDir ../.))
        ))
      ]
    );
  };
```

Which is not awfully pleasant

XYenon: There are many `.lua` files in `yazi-plugin/preset`.

What about this?

```nix
  src = lib.fileset.toSource {
    root = ../.;
    fileset = lib.fileset.intersection (lib.fileset.fromSource (lib.sources.cleanSource ../.)) (
      lib.fileset.unions [
        ../assets
        ../Cargo.toml
        ../Cargo.lock
        (lib.fileset.fromSource (lib.sources.sourceByRegex ../. [ "^yazi-.*" ]))
      ]
    );
  };
```

XYenon (approved):
LGTM :)

sxyazi (approved):
LGTM, thanks everyone! 

I'll merge it now 🙂

sxyazi:
It seems the Cachex failed now https://github.com/sxyazi/yazi/actions/runs/10303852646/job/28520903582

@isabelroses please have a look

isabelroses:
> It seems the Cachex failed now https://github.com/sxyazi/yazi/actions/runs/10303852646/job/28520903582
> 
> @isabelroses please have a look

My best guess here is that rust_overlay is now using rust 1.80 and its unhappy with cargo-c's code thus is generating an error. We could test by rolling back a few versions.

sxyazi:
Should I revert this PR and you can submit a new one?

isabelroses:
> Should I revert this PR and you can submit a new one?

I can create a Pr in a second pinning the version, when i find a successful build.

PR #1388, opened 2024-08-01
chore: move common dependencies to workspace Cargo.toml

karimElmougi:
It's easier to keep versions in sync across the entire workspace this way.

### DIFF
--- a/yazi-adapter/Cargo.toml
+++ b/yazi-adapter/Cargo.toml
@@ -26,9 +26,7 @@ kamadak-exif = "0.5.5"
 ratatui      = { workspace = true }
 scopeguard   = { workspace = true }
 tokio        = { workspace = true }
-
-# Logging
-tracing = { workspace = true }
+tracing      = { workspace = true }
 
 [target.'cfg(target_os = "macos")'.dependencies]
 crossterm = { workspace = true, features = [ "use-dev-tty" ] }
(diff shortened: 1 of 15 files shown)


sxyazi:
Thanks for the PR! 

Every crate of Yazi gets published to crates.io, will this have any impact on it?

karimElmougi:
I don't believe it should cause any issues. [quickwit](https://github.com/quickwit-oss/quickwit/tree/main/quickwit) makes use of workspace dependencies as well, and [their crates get published as well](https://crates.io/search?q=quickwit).

sxyazi:
> I don't believe it should cause any issues. [quickwit](https://github.com/quickwit-oss/quickwit/tree/main/quickwit) makes use of workspace dependencies, and [their crates get published as well](https://crates.io/search?q=quickwit).

Nice, TIL about that.

yazi-adapter/Cargo.toml
```diff
@@ -13,22 +13,22 @@ yazi-config = { path = "../yazi-config", version = "0.3.0" }
...
-arc-swap     = "1.7.1"
-base64       = "0.22.1"
+ansi-to-tui  = { workspace = true }
+anyhow       = { workspace = true }
+arc-swap     = { workspace = true }
+base64       = { workspace = true }
 color_quant  = "1.1.0"
-crossterm    = "0.27.0"
-futures      = "0.3.30"
+crossterm    = { workspace = true }
```
sxyazi, line 21: We can't use `{ workspace = true }` here because the "use-dev-tty" feature can only be enabled on macOS systems.

sxyazi (changes requested):

yazi-adapter/Cargo.toml
karimElmougi: Ah, I see. I hadn't realized. I pushed a fix.

sxyazi (approved):
LGTM, thanks!

PR #1395, opened 2024-08-02
fix: preview ascii escape image in yazi#1391

### LINKED ISSUE #1391: Extra symbols remained in the window after trying to preview ASCII escape codes

macOS

kitty 0.35.2

### `yazi --debug` output

<details>
```sh
Yazi
    Version: 0.3.0 (Homebrew 2024-08-01)
    Debug  : false
    OS     : macos-x86_64 (unix)

Ya
    Version: 0.3.0

Emulator
    Emulator.via_env: ("xterm-kitty", "")
    Emulator.via_csi: Ok(Kitty)
    Emulator.detect : Kitty

Adapter
    Adapter.matches: Kitty

Desktop
    XDG_SESSION_TYPE: None
    WAYLAND_DISPLAY : None
    DISPLAY         : Some("/private/tmp/com.apple.launchd.RYRSLZAOnF/org.macosforge.xquartz:0")

SSH

WSL
    /proc/sys/fs/binfmt_misc/WSLInterop: false

Variables
    YAZI_FILE_ONE      : None
    YAZI_CONFIG_HOME   : None
    ZELLIJ_SESSION_NAME: None

Text Opener
    default: Some(Opener { run: "nvim \"$@\"", block: true, orphan: false, desc: "nvim", for_: None, spread: true })
    block  : Some(Opener { run: "nvim \"$@\"", block: true, orphan: false, desc: "nvim", for_: None, spread: true })

tmux
    TMUX   : false
    Version: No such file or directory (os error 2)

Dependencies
    file             : 5.41
    ffmpegthumbnailer: No such file or directory (os error 2)
    magick           : No such file or directory (os error 2)
    fzf              : 0.54.2
    fd               : 10.1.0
    rg               : 14.1.0
    zoxide           : 0.9.4
    7z               : No such file or directory (os error 2)
    jq               : No such file or directory (os error 2)
```
</details>

### Describe the bug

https://github.com/user-attachments/assets/2c7f0e86-4889-4152-8778-071547f026c1

### Expected Behavior

No extra symbols in the terminal window

### To Reproduce

1. `git clone https://github.com/nvim-telescope/telescope.nvim`
2. go to `./telescope.nvim/data/memes/planets`
3. hover one of the `planets/` files

### Configuration

_No response_

### Anything else?

If I exit yazi and open it again it resets the window.

Xerxes-2:
fix #1391

### DIFF
--- a/yazi-plugin/src/external/highlighter.rs
+++ b/yazi-plugin/src/external/highlighter.rs
@@ -97,8 +97,14 @@ impl Highlighter {
 		}
 
 		if plain {
-			let indent = " ".repeat(PREVIEW.tab_size as usize);
-			Ok(Text::from(after.join("").replace('\t', &indent)))
+			if after.iter().any(|l| l.contains('\x1b')) {
+				let bytes = after.join("").bytes().collect::<Vec<_>>();
+				use ansi_to_tui::IntoText;
+				bytes.into_text().map_err(|e| PeekError::Unexpected(e.to_string()))
+			} else {
+				let indent = " ".repeat(PREVIEW.tab_size as usize);
+				Ok(Text::from(after.join("").replace('\t', &indent)))
+			}
 		} else {
 			Self::highlight_with(before, after, syntax.unwrap()).await
 		}


sxyazi:
I think we should preview these ASCII sequences as code, rather than outputting and rendering them directly to the terminal because:

1. `ansi_to_tui` only supports a limited set of cases. If these escape codes include cursor movements, it could cause issues.
2. This is a [code previewer](https://github.com/sxyazi/yazi/blob/9d1fab84efdd78dcf102262601d6965ec760ad95/yazi-config/preset/yazi.toml#L106-L108), so the expected result should be treated as a code preview, not as actual rendering them.

Xerxes-2:
> I think we should preview these ASCII sequences as code, rather than outputting and rendering them directly to the terminal because:
> 
> 1. `ansi_to_tui` only supports a limited set of cases. If these escape codes include cursor movements, it could cause issues.
> 2. This is a [code previewer](https://github.com/sxyazi/yazi/blob/9d1fab84efdd78dcf102262601d6965ec760ad95/yazi-config/preset/yazi.toml#L106-L108), so the expected result should be treated as a code preview, not as actual rendering them.

Now the previewer should treat esc as space

yazi-plugin/src/external/highlighter.rs
```diff
@@ -84,6 +84,12 @@ impl Highlighter {
 				buf.push(b'\n');
 			}
 
+			buf.iter_mut().for_each(|b| {
+				if *b == 27 {
+					*b = b' '
```
sxyazi, line 89: Please use `"^["` instead of spaces

yazi-plugin/src/external/highlighter.rs
```diff
@@ -84,6 +84,8 @@ impl Highlighter {
 				buf.push(b'\n');
 			}
 
+			buf.iter_mut().filter(|&&mut b| b == b'\x1b').for_each(|b| *b = b'^');
```
sxyazi, line 87: No, it should be `"^["`, not `"^"`

Issue #1382, opened 2024-08-01
CI: Add automatic nightly build to the workflow


sxyazi:
### Please describe the problem you're trying to solve

This will allow users to download pre-built binaries from the release page instead of building them themselves. 

It will also benefit downstream projects like `yazi.nvim`, where automated testing in CI relies on the latest nightly version for some features, https://github.com/mikavilpas/yazi.nvim/issues/135#issuecomment-2205571261.

### Would you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

Add a workflow for nightly builds to the current GitHub Actions

uznog:
I'll open a PR soon!

sxyazi:
Done in https://github.com/sxyazi/yazi/pull/1397, thanks for contributing this feature @uznog!

PR #1397, opened 2024-08-02
feat: Add automatic nightly builds to the Github Actions workflow

uznog:
Named after and hopefully resolving #1382

Changes include:
- workflow running at a cron schedule (12:00 AM / 00:00 UTC)
- workflow scheduled by cron runs `nightly` job, generating a pre-release

Nightly build name is `yazi-nightly-<timestamp>`. Flow tested with dummy artifact created:
https://github.com/uznog/yazi/actions/runs/10215675482

And here's the resulting pre-release: https://github.com/uznog/yazi/releases/tag/yazi-nightly-20240802-1236

### DIFF
--- a/.github/workflows/draft.yml
+++ b/.github/workflows/draft.yml
@@ -5,7 +5,7 @@ on:
     tags:
       - "v[0-9]+.[0-9]+.[0-9]+"
   schedule:
-    - cron: '0 0 * * *'
+    - cron: "0 */8 * * *"
 
 jobs:
   build-unix:
@@ -147,7 +147,7 @@ jobs:
             yazi-*.zip
             yazi-*.snap
           generate_release_notes: true
-  
+
   nightly:
     permissions:
       contents: write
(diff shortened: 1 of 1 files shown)


uncenter:
If I'm reading this right, this creates a new nightly release every day? I think creating a single `nightly`-tagged rolling release might be better 🤔 , though I might not be understanding exactly what this is trying to solve.

uznog:
I totally agree. I pushed a commit changing tag name to be a static `nightly`, which will work as the latest version built off master branch every day

sxyazi:
Thanks for the PR! 

Could you also add a build time to it? Something like: 

```
From commit: d2ebadb
Generated on: 2024-08-04 23:00:00 UTC
```

uznog:
I guess this does the job:
https://github.com/uznog/yazi/releases/tag/nightly

sxyazi (approved):
Thanks!

PR #1402, opened 2024-08-03
Added config showing tabs always

Jelenkee:


sxyazi:
Thanks for the PR!

I usually say no to adding options like this because I want to keep the number of settings low and offer sensible defaults that work for most situations. This makes it easier for new users to get started with Yazi, as they only have to learn a few settings. 

In this case, I don't really see the need - when would you ever need to show just one tab? I think it only makes sense to display tabs when there's more than one to save some space.

PR #1442, opened 2024-08-08
fix(nix): pin `rust-overlay` to workaround `cargo-c`

isabelroses:
fix from https://github.com/sxyazi/yazi/pull/1380#issuecomment-2276039515


sxyazi (approved):
Thanks for the quick fix!

PR #1443, opened 2024-08-08
feat: start with multiple tabs with different paths

### LINKED ISSUE #1322: Start with multiple tabs with different paths

yazi only works with 1 path.
i dont see any options in the help pages that allows to open multiple tabs on startup.

eg.
`yazi ~/Downloads`

add option to start yazi with multiple tabs with different paths on startup

`yazi ~/Download ~/Documents ~/Public ~/Music`

### Additional context

_No response_

coolkiid:
1. CLI can receive more than one paths as entries now.
2. Modify BOOT in yazi-boot/ and Tabs in yazi-core to fit the change above.
3. Add a test for parse_entries().

Closes #1322

For example, to open multi tabs at once, run `yazi ~/Projects ~/Downloads ~/Documents`.


yazi-boot/src/args.rs
```diff
@@ -6,8 +6,8 @@ use clap::{command, Parser};
 #[command(name = "yazi")]
 pub struct Args {
 	/// Set the current working entry
-	#[arg(index = 1)]
-	pub entry: Option<PathBuf>,
+	#[arg(index = 1, value_delimiter = ' ', num_args = 1..)]
```
sxyazi, line 9: Please limit the maximum number of parameters to 9 as Yazi can only create up to 9 tabs.

coolkiid: done.

yazi-boot/src/args.rs
```diff
@@ -6,8 +6,8 @@ use clap::{command, Parser};
 #[command(name = "yazi")]
 pub struct Args {
 	/// Set the current working entry
-	#[arg(index = 1)]
-	pub entry: Option<PathBuf>,
+	#[arg(index = 1, value_delimiter = ' ', num_args = 1..=9)]
+	pub entries: Vec<PathBuf>,
```
sxyazi, line 10: Why do we need `value_delimiter = ' '` here?

coolkiid: yea actually it is not necessary, i think it can be deleted. the behavior is the same with or without it.

sxyazi (approved):
LGTM, thank you so much for contributing this feature!

PR #1461, opened 2024-08-10
feat: add keybinding for the `hardlink` command

ahanniga:
Ctrl-h proposed as key


sxyazi (approved):
Thanks, I've changed it to `Ctrl` + `-` to align it with the `-` for softlink.

Issue #1396, opened 2024-08-02
DDS kind for switching tabs


Rolv-Apneseth:
### `yazi --debug` output

N/A

### Please describe the problem you're trying to solve

I had [an issue](https://github.com/Rolv-Apneseth/starship.yazi/issues/6) opened for `starship.yazi` that pointed out that it wasn't updating when switching between tabs. When I was looking into it, I couldn't find a way to run the plugin when the tabs change.

### Would you be willing to contribute this feature?

- [X] Yes, I'll give it a shot

### Describe the solution you'd like

If I'm not missing something, would it be possible to add a DDS kind when switching tabs so plugins can subscribe to that event?

### Additional context

I can give it a shot if you point me in the right direction (haven't touched the code in a while).

### Validations

- [X] I have searched the existing issues
- [X] The latest nightly build of Yazi doesn't already have this feature

mikavilpas:
@Rolv-Apneseth I opened a PR that could help implement this. It currently sends the index of the tab that was switched to as an event.

Do you think this could be enough for implementing this feature in your plugin? I'm not sure if the current directory is available from yazi's APIs as long as we know the current tab.

Rolv-Apneseth:
Oh nice, thanks. I believe the cwd is based on the current tab anyway but yeah seems to work perfectly for me, just subbed to the `tab-switch` event and it updates no problem.

PR #1474, opened 2024-08-12
feat: new `tab` DDS event on tab switch

mikavilpas:
This is a simple suggestion on how https://github.com/sxyazi/yazi/issues/1396 could be solved - it simply sends the index of the tab that was switched to in yazi.

Example:

```js
$ ya sub tab-switch
tab-switch,0,1723480172088356,{"tab":1}
tab-switch,0,1723480172088356,{"tab":0}
```

Some observations for your consideration:
- the index in this implementation is currently 0-based while yazi displays tabs as 1-based (1-9). This seems to be what is happening inside yazi, but I'm not sure if this is okay or not.
- other tab commands seem to behave as expected (switch to tab, switch to next/previous tab, move tab, close tab all work without issues)

Closes https://github.com/sxyazi/yazi/issues/1396


sxyazi (approved):
Thanks!

Issue #1495, opened 2024-08-15
Compilation issues while building from source


AnirudhG07:
### What system are you running Yazi on?

macOS

### What terminal are you running Yazi in?

iTerm2

### `yazi --debug` output

```Shell
❯ yazi --debug                                                                                                                                                               ─╯

Yazi
    Version: 0.3.0 (7a380c2 2024-08-01)
    Debug  : false
    OS     : macos-aarch64 (unix)

Ya
    Version: 0.3.1

Emulator
    Emulator.via_env: ("xterm-256color", "iTerm.app")
    Emulator.via_csi: Ok(Iterm2)
    Emulator.detect : Iterm2

Adapter
    Adapter.matches: Iterm2

Desktop
    XDG_SESSION_TYPE: None
    WAYLAND_DISPLAY : None
    DISPLAY         : None

SSH
    shared.in_ssh_connection: false

WSL
    /proc/sys/fs/binfmt_misc/WSLInterop: false

Variables
    SHELL              : Some("/bin/zsh")
    EDITOR             : Some("nvim")
    YAZI_FILE_ONE      : None
    YAZI_CONFIG_HOME   : None
    ZELLIJ_SESSION_NAME: None

Text Opener
    default: Some(Opener { run: "${EDITOR:=vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block  : Some(Opener { run: "${EDITOR:=vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })

tmux
    TMUX   : false
    Version: 3.4

Dependencies
    file             : 5.41
    ueberzugpp       : No such file or directory (os error 2)
    ffmpegthumbnailer: 2.2.2
    magick           : 7.1.1-36
    fzf              : 0.54.3
    fd               : 10.1.0
    rg               : 14.1.0
    chafa            : No such file or directory (os error 2)
    zoxide           : 0.9.4
    7z               : 17.05
    7zz              : No such file or directory (os error 2)
    jq               : 1.6


--------------------------------------------------
When reporting a bug, please also upload the `yazi.log` log file - only upload the most recent content by time.
You can find it in the "/Users/anirudhgupta/.local/state/yazi" directory.
```

### Did you try the latest nightly build to see if the problem got fixed?

Yes, and I updated the debug information above (`yazi --debug`) to the nightly that I tried

### Describe the bug

```
[...]

PR #1499, opened 2024-08-16
fix: upgrade `ansi-to-tui` to `5.0.0-rc.1` to resolve `ratatui` version conflict

sxyazi:
Fixes https://github.com/sxyazi/yazi/issues/1495


AnirudhG07:
Great now building from source works without any issues!

Issue #62, opened 2023-08-15
Deleting files in search mode doesn't refresh tab


TD-Sky:
Version: https://github.com/sxyazi/yazi/commit/8fd3d15918b9ebba127fe11cff38b1bb73393f60

sxyazi:
Thanks for reporting, this is a known issue, as refreshes have to be re-searched, and that's a lot of overhead.

Perhaps a way needs to be found to partially update user changed items.

aidanmontare-fed:
Might be similar behavior: changing a file permission by running a shell command `;` then `chmod +x script.sh` doesn't update the displayed file permissions.

PR #1501, opened 2024-08-16
fix: refresh after bulk renaming (#1418)

### LINKED ISSUE #1418: No re-sort after mass name change

Linux Wayland

kitty 32.3

### `yazi --debug` output

<details>

```sh
##### ↓↓↓ Paste the output here: ↓↓↓ #####


```

</details>

### Describe the bug

When I select several filenames and change them. I then save and return to yazi. Unfortunately, the display does not re-sort to reflect new file order. I need to leave the directory and enter again to get a re-sort.

### Minimal reproducer

Select several files and select "r" to change the names. Change the names in a way that would change the sort order. Save and return.

### Anything else?

I suggest forcing a re-sort after a mass name change. Or at least providing a command to refresh/reload a directory.

TobisLee:
Fixes #1418

### DIFF
--- a/yazi-core/src/manager/commands/bulk_rename.rs
+++ b/yazi-core/src/manager/commands/bulk_rename.rs
@@ -5,7 +5,7 @@ use scopeguard::defer;
 use tokio::{fs::{self, OpenOptions}, io::{stdin, AsyncReadExt, AsyncWriteExt}};
 use yazi_config::{OPEN, PREVIEW};
 use yazi_dds::Pubsub;
-use yazi_proxy::{AppProxy, TasksProxy, HIDER, WATCHER};
+use yazi_proxy::{AppProxy, ManagerProxy, TasksProxy, HIDER, WATCHER};
 use yazi_shared::{fs::{max_common_root, maybe_exists, paths_to_same_file, File, FilesOp, Url}, terminal_clear};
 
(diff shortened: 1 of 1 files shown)


AnirudhG07:
can we have refresh, not just after bulk-renaming, but many more places? For example, after mactag, you have to quit and refresh for things to show up. Various issues pointing to exiting and coming back for effects to show up, should be solved if refresh can be placed somewhere more general in my opinion...

TobisLee:
> should be solved if refresh can be placed somewhere more general in my opinion...

I agree with you, but I'm not sure if this will become a performance bottleneck since the `refresh()` function resorts files.

As for this PR, it actually fixes two issues:
1. The lack of re-sorting after bulk renaming.
2. When bulk renaming files in both the current directory and a subdirectory, the files from the subdirectory appear in the current directory after renaming.

Here is the video for the second issue.

[Kooha-2024-08-19-11-56-40.webm](https://github.com/user-attachments/assets/2b04e761-9daa-4a1d-864d-231a1aa94bef)

AnirudhG07:
I was seeing some issues and i found #62, might be why it wasnt implemented so far. Also wondering why is it still open?

TobisLee:
Yeah, seems like `refresh()` does reduce the performance. I'll close this pr.

PR #1505, opened 2024-08-17
feat: add `--dir` option to `create` command

abstrakct:
This simple pull request adds a new option to the 'create' command: --dir

With this option 'create' will always create a directory.
This allows you to bind 'create --dir' to a key, so you can easily create directories without having to remember to put / or \ at the end of the name.

Personally it makes sense to me to have a command for this. I very rarely need to create empty files, but I often create directories, and used to have a keybinding for this in ranger. After switching to yazi I very often forget to add the / when I want to make a directory.

I've tested this change on Linux. 'create --dir' now always creates a directory, trailing / is ignored, trailing \ is treated as part of the directory name (which is ok in Linux). Not tested on Windows, because I have no Windows machines :)

Let me know if I've missed something, or if there is a better way to add this functionality.

### DIFF
--- a/yazi-core/src/manager/commands/create.rs
+++ b/yazi-core/src/manager/commands/create.rs
@@ -4,22 +4,17 @@ use anyhow::Result;
 use tokio::fs;
 use yazi_config::popup::InputCfg;
 use yazi_proxy::{InputProxy, TabProxy, WATCHER};
-use yazi_shared::{
-	event::Cmd,
-	fs::{maybe_exists, ok_or_not_found, symlink_realpath, File, FilesOp, Url},
-};
+use yazi_shared::{event::Cmd, fs::{maybe_exists, ok_or_not_found, symlink_realpath, File, FilesOp, Url}};
 
 use crate::manager::Manager;
 
 pub struct Opt {
 	force: bool,
-	dir: bool,
+	dir:   bool,
 }
 
 impl From<Cmd> for Opt {
-	fn from(c: Cmd) -> Self {
-		Self { force: c.bool("force"), dir: c.bool("dir") }
-	}
+	fn from(c: Cmd) -> Self { Self { force: c.bool("force"), dir: c.bool("dir") } }
 }
 
 impl Manager {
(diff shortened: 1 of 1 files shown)


sxyazi:
Hey, thanks for the PR!

There's already a [`mkdir.yazi`](https://github.com/Sonico98/mkdir.yazi) plugin available, does it meet your needs?

Sonico98:
> Hey, thanks for the PR!
> 
> There's already a [`mkdir.yazi`](https://github.com/Sonico98/mkdir.yazi) plugin available, does it meet your needs?

Tbh I wrote the plugin due to my lack of rust knowledge and consider it to be a bit of a hack. I too believe that a `--dir` option should be available by default in yazi.

On a side note, I've been unsuccessful in properly highlighting the newly created directory in the same way the `create` command does, which makes the experience a bit worse.

sxyazi:
Fair enough, I'll review and merge it tomorrow

sxyazi (approved):
Thanks!

sxyazi:
Could you please also add the new option to the docs here https://yazi-rs.github.io/docs/configuration/keymap#manager.create? (Click the bottom link to edit)

abstrakct:
Documentation added! :+1:

Tyarel8:
> Nice. I'm thinking about using
> 
> ```toml
> [manager]
> 
> keymap = [
>   { on = "a", run = "create", desc = "Create a file" },
>   { on = "A", run = "create --dir", desc = "Create a directory" },
> ]
> ```
> 
> Any chance the pop-up window title could adapt? For now, it's `Create:` in both cases.
> 
> Maybe `Create file:` or `Create directory:` would be better descriptions? Just like `;` gives `Shell` and `:` gives `Shell (block):` or `s` gives `Search via fd:` and `S` gives `Search via rg`.
> 
> Edit: I've just realized that this wouldn't be a good idea for users that stick with the current default.
> 
> `Create:` makes sense considering you can create either a file or a directory by adding an extra `/` - `\` at the end.

`Create:` makes sense for the previous behavior because it can be used to create files or directories, but for the new option `create --dir` I think it would be better to have a different prompt so you know the difference like `Create dir:`.

PR #1512, opened 2024-08-18
perf: new `image_delay` option debounces image previews to avoid lag caused by terminal image decoding during fast scrolling

sxyazi:
I've been working on optimizing Yazi's image preview speed, and with multi-threading, preloading, and a built-in image decoder, it should already be the fastest among all terminal file managers.

However, I noticed some lag during fast scrolling and realized that at this point, the performance bottleneck isn't Yazi itself, but rather the terminal. When users scroll quickly, Yazi processes the images at a very high speed and sends them to the terminal, but the terminal can't keep up with processing them in time, which gives the impression of lag. In reality, it's not Yazi that's lagging, but the terminal.

This PR introduces a new `image_delay` configuration option to address this issue. When previewing images, it will wait at least `image_delay` milliseconds before starting to send the decoded image data to the terminal. 

This gives the terminal a breather and creates the perception that the file list is still scrolling smoothly.

## Before: 

https://github.com/user-attachments/assets/8cf40a08-95b5-44fd-a16b-a1efb70b25f6

## Now:

https://github.com/user-attachments/assets/b5dde838-b700-4fa3-97be-543abff0f4ab

### DIFF
--- a/yazi-config/preset/yazi.toml
+++ b/yazi-config/preset/yazi.toml
@@ -21,6 +21,7 @@ tab_size        = 2
 max_width       = 600
 max_height      = 900
 cache_dir       = ""
+image_delay     = 30
 image_filter    = "triangle"
 image_quality   = 75
 sixel_fraction  = 15
--- a/yazi-config/src/preview/preview.rs
+++ b/yazi-config/src/preview/preview.rs
@@ -15,6 +15,7 @@ pub struct Preview {
 
 	pub cache_dir: PathBuf,
 
+	pub image_delay:    u8,
 	pub image_filter:   String,
 	pub image_quality:  u8,
 	pub sixel_fraction: u8,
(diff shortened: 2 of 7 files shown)


leihaojun:
It seems that many terminals do not have this problem, I have tested foot,st-sx and kitty without this problem, only found this problem on konsole.

leihaojun:
I think this delay mechanism is a great mechanism for reducing resource usage. Is there any consideration of extending this mechanism to general tasks?

For example,, we can add two more parameters to `ya.manager_emit`, one is the `task_id` and one is the `delay`. If a task with the same `task_id` is triggered before the `delay` is over, the `delay` is reset to the `preset value` instead of adding a `new task`.

ya.manager_emit(cmd,{xxx,id=xxx,delay=xxx})

sxyazi:
The `ya.manager_emit` belongs to Yazi's event system, I guess you're referring to Yazi's concurrent task scheduling system since only tasks have task IDs. 

I plan to expose this through the plugin system in the future, which means there will be a way to create, destroy, and report task progress. This will be particularly useful for the built-in `extract` plugin. 

At that point, Yazi's progress can be automatically synced with the extraction progress from `7zip`. Users will be able to see the progress bar for each task in the task manager, and the overall progress bar at the bottom will include it as well.

PR #1547, opened 2024-08-24
fix: wrong `debug_assert` in `OrderedFloat` for NaN

yanCode:
When calling `new` function to create a `OrderedFloat`, it shouldn't allow `NAN` type to pass in, but the issue now only allows `NAN` , which does make much sense for the logic of `OrderedFloat`.

I guess this issue hasn't been detected before is possibly because `OrderedFloat` only works on some edge cases. But anyway, it's better to make the code more robust.


sxyazi (approved):
Nice catch, thanks for the patch!

PR #1565, opened 2024-08-28
fix: choose chafa if ueberzugpp does not exist

taoky:
Currently yazi-adapter always uses Überzug++ if terminal does not support other image protocols and desktop environment is X11 or Wayland. However, Überzug++ does not support some terminals (https://github.com/sxyazi/yazi/pull/1066), and unfortunately it also does not some Wayland compositors like GNOME mutter (some protocols not implemented in mutter maybe?).

In this case, yazi would show nothing in preview for images even if the fallback shall be "Chafa".

This PR tries to check if `ueberzugpp` exists in `PATH` -- and falls back to `chafa` if not. For Windows cfg it just skips `ueberzugpp`/X11/Wayland check as Überzug++ does not support Windows.

### DIFF
--- a/yazi-adapter/src/adapter.rs
+++ b/yazi-adapter/src/adapter.rs
@@ -1,4 +1,4 @@
-use std::{env, fmt::Display, path::Path, sync::Arc};
+use std::{env, fmt::Display, fs, path::Path, sync::Arc};
 
 use anyhow::Result;
 use ratatui::layout::Rect;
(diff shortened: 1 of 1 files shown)


sxyazi:
Thanks for the PR!

Does this mean that if users want to fall back to chafa, they have to uninstall ueberzugpp?

taoky:
> Does this mean that if users want to fall back to chafa, they have to uninstall ueberzugpp?

Yes. Unfortunately ueberzugpp correctly does not reliably report issues when current environment is unsupported -- it crashes, generates a coredump instead of exiting with a meaningful exit code:

```console
> ueberzug layer -o wayland
terminate called after throwing an instance of 'std::runtime_error'
  what():  Unable to calculate terminal sizes
fish: Job 1, 'ueberzug layer -o wayland' terminated by signal SIGABRT (Abort)
```

I don't think that generating a new coredump every time yuzi starts and checks adapter is a good idea.

sxyazi:
Thanks for the clarification, in this case, I'm afraid I can't merge the PR because users might have multiple terminal emulators installed, and `ueberzugpp` may not work only under a few of them (the ones that don't implement this `ioctl` system call are rare). Forcing Yazi to fall back to `chafa` by having to uninstall `ueberzugpp` seems a bit harsh, and it introduces a new uncertainty — if users accidentally install `ueberzugpp`, the original `chafa` functionality might be affected just because they have it installed.

I see two possible approaches:

- Request `ueberzugpp` to add a reliable way to detect whether the user's terminal supports it, maybe with a new flag like `ueberzugpp --detect`.
- Use existing environment variables as a workaround. As I mentioned, terminals that don't support this `ioctl` are rare, so in most cases, `ueberzugpp` will just work. But for some niche cases, users can force Yazi to use `chafa` by tricking it with `NVIM=1 NVIM_LOG_FILE=1 yazi`.

taoky:
> `ueberzugpp` may not work only under a few of them (the ones that don't implement this ioctl system call are rare)

`ueberzugpp layer -o wayland` also does not work with compositors other than sway, hyprland and wayfire. But afaik only VTE (stable branch) does not support both sixel and the `ioctl` to get terminal width & height, so yazi would probably choose other image protocols before `ueberzugpp` when using other modern terminal emulators, and only VTE-based terminals (gnome-terminal, tilix, etc.) would suffer from this issue.

I'm fine with both approaches.

sxyazi:
> also does not work with compositors other than sway, hyprland and wayfire

Do you think there's any way to detect them in Yazi? I think this would make more sense than checking if `ueberzugpp` is installed.

taoky:
> Do you think there's any way to detect them in Yazi? I think this would make more sense than checking if `ueberzugpp` is installed.

https://github.com/jstkdng/ueberzugpp/blob/e66353e1e0a6dc3238d1ff062d2928c5d848a626/src/canvas/wayland/config.cpp#L24-L42

Looks like ueberzugpp just simply checks three envs. I would submit a new PR for implementing this logic when deciding if using ueberzugpp wayland output.

PR #1566, opened 2024-08-28
fix: check compositor support status before using ueberzug wayland output

taoky:
This PR tries to check if current compositor is supported by Überzug++ before choosing `Adapter::Wayland`, as it's a cleaner approach than checking if `ueberzugpp` binary exists.

A bad news is that, if the terminal emulator is a wayland window, and the wayland session supports xwayland (in almost all cases), then it would still show nothing, as `DISPLAY` env exists, Überzug++ X11 output is used with errors in ueberzugpp log:

```log
[2024-08-29 01:21:27.923] [X11] [error] XCB: Window:no_extension, CreateWindow:no_minor, resource 3 sequence 32
[2024-08-29 01:21:27.923] [X11] [error] XCB: Drawable:no_extension, CreateGC:no_minor, resource 52428800 sequence 33
[2024-08-29 01:21:27.923] [X11] [error] XCB: Window:no_extension, ChangeProperty:no_minor, resource 52428800 sequence 34
[2024-08-29 01:21:27.923] [X11] [error] XCB: Window:no_extension, MapWindow:no_minor, resource 52428800 sequence 35
[2024-08-29 01:21:27.923] [X11] [error] XCB: Window:no_extension, SendEvent:no_minor, resource 52428800 sequence 36
```

This could be workarounded by:

```shell
DISPLAY= yazi
```

Related: #1565


yazi-adapter/src/adapter.rs
```diff
@@ -95,12 +95,23 @@ impl Adapter {
...
+			|| env_exists("WAYFIRE_SOCKET");
+
 		match env::var("XDG_SESSION_TYPE").unwrap_or_default().as_str() {
 			"x11" => return Self::X11,
-			"wayland" => return Self::Wayland,
+			"wayland" => {
+				if supports_ueberzug_wayland {
+					return Self::Wayland;
+				}
+			}
```
sxyazi, line 111:
```suggestion
			"wayland" if supports_ueberzug_wayland => return Self::Wayland,
```

sxyazi:
Could you please also add these 3 vars to the `yazi --debug` here?

https://github.com/sxyazi/yazi/blob/ddb8ce530f4faa1fadc72f529ecad99d0c3514e0/yazi-boot/src/actions/debug.rs#L29-L32

sxyazi (approved):
LGTM, thanks alot for the contribution! <3

Issue #1277, opened 2024-07-08
Show number of files for folders in size linemode


Ape:
### Please describe the problem you're trying to solve

I'd like to see the number of files in each folder when I am using the size linemode.

And no, the size of the folder on disk in bytes is not meaningful. I want to have the number of files inside.

### Would you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

It would look this:
```
empty_folder       0
small_folder       4
big_folder        53
regular_file   15.8K
big_file        3.2G
```

PoisonFrog:
I also prefer it like this, but I think the best would be if we had an option called "dirlinemode" or something like this, then the user would be able to choose if he wants to show the number of files inside the directory or the size of the folder.

sxyazi:
Done in https://github.com/sxyazi/yazi/pull/1591 🎉

PR #1576, opened 2024-08-31
Add entry count to directories

SolitudeSF:
Not sure how im supposed to treat the case when file is not a directory.

### DIFF
--- a/yazi-plugin/src/file/file.rs
+++ b/yazi-plugin/src/file/file.rs
@@ -42,6 +42,14 @@ impl File {
 				IconCache::Icon(cached) => Some(Icon::cast(lua, cached)).transpose(),
 			}
 		});
+
+		reg.add_method("count", |_, me, ()| {
+			if me.as_ref().cha.is_dir() {
+				return Ok(Some(std::fs::read_dir::<&std::path::Path>(me.as_ref().url.as_ref())?.count()));
+			} else {
+				return Ok(None);
+			}
+		});
 	}
 
 	pub fn install(lua: &Lua) -> mlua::Result<()> {


sxyazi:
Could you describe in detail what it's used for, when it is needed, and how you would use it?

SolitudeSF:
i needed it for the plugin that displays filecount on the directories, replicating ranger's behaviour

```lua
local save = ya.sync(function(st, counts)
	st.counts = counts
	ya.render()
end)

local function setup(st, opts)
	st.counts = {}

	Linemode:children_add(function(self)
    		if self._file.cha.is_dir and next(st.counts) then
        		local count = st.counts[tostring(self._file.url)]
			return ui.Line { ui.Span(tostring(count)) }
		else
    			return ui.Line("")
		end
	end, opts.order or 4000)
end

local function fetch(self)
	local counts = {}

	for _, file in ipairs(self.files) do
    		if file.cha.is_dir then counts[tostring(file.url)] = file:count() end
	end

	save(counts)

	return 3
end

return { setup = setup, fetch = fetch }
```

TobisLee:
It seems this PR is attempting to solve #1277? I've been looking for this feature for a while. After applying this commit, I found two issues:
1. When I navigate to / or /home, directories and files don't display correctly.
图片

2. When counting the number of files, hidden and unhidden files aren't distinguished, which could be confusing.
图片

In my opinion, if the item is a folder, it should show the item count, and if the item is a file, it should show the file size.

P.S.: I also tried to solve this with `fs::read_dir` (See https://github.com/TobisLee/yazi/commit/aa965199f5c2c5ec03d393db5e9fc83bc41de0e8), but I don't think it's the best approach since there's already a `Folder` struct, and `folder.files.len()` can get the file count without considering the hidden files. However, I couldn't find a way to use `files.len()` to implement this feature. :(

SolitudeSF:
works fine on my machine. unless you are using the plugin i provided, it shouldn't have any effect on your setup.

sxyazi:
I'll make a PR to re-implement it - we don't need any sync I/O, and also we need to avoid any I/O as much as possible. In this case, `history` already contains the complete directory list the user passed through. 

https://github.com/sxyazi/yazi/blob/fdd574997bf237a20dde31c9c7a28b3c47345cfc/yazi-core/src/tab/tab.rs#L23

SolitudeSF:
this was more of an rfc, since im not familiar with either rust or yazi's performance sensitive internals. as long as i can display a count and filesize.

PR #1580, opened 2024-08-31
feat: ensure that `Entity` and `Linemode` have consistent space padding

SolitudeSF:
For now the spacing of linemodes is a wild west. This change brings some uniformity - every linemode entry is followed by a space. This also means that every plugin linemode should only return text, without padding of spaces, just like builtin linemodes do.

Another path would be to add spacing parameters to children, so user can decided if certain linemodes should be padded with spaces or not.

### DIFF
--- a/yazi-plugin/preset/components/entity.lua
+++ b/yazi-plugin/preset/components/entity.lua
@@ -1,24 +1,27 @@
 Entity = {
 	_inc = 1000,
 	_children = {
-		{ "icon", id = 1, order = 1000 },
-		{ "prefix", id = 2, order = 2000 },
-		{ "highlights", id = 3, order = 3000 },
-		{ "found", id = 4, order = 4000 },
-		{ "symlink", id = 5, order = 5000 },
+		{ "space", id = 1, order = 1000 },
+		{ "icon", id = 2, order = 2000 },
+		{ "prefix", id = 3, order = 3000 },
+		{ "highlights", id = 4, order = 4000 },
+		{ "found", id = 5, order = 5000 },
+		{ "symlink", id = 6, order = 6000 },
 	},
 }
 
 function Entity:new(file) return setmetatable({ _file = file }, { __index = self }) end
+
+function Entity:space() return ui.Line(" ") end
 
 function Entity:icon()
 	local icon = self._file:icon()
 	if not icon then
 		return ui.Line("")
 	elseif self._file:is_hovered() then
-		return ui.Line(" " .. icon.text .. " ")
+		return ui.Line(icon.text .. " ")
 	else
-		return ui.Line(" " .. icon.text .. " "):style(icon.style)
+		return ui.Line(icon.text .. " "):style(icon.style)
 	end
 end
 
--- a/yazi-plugin/preset/components/linemode.lua
+++ b/yazi-plugin/preset/components/linemode.lua
@@ -14,14 +14,10 @@ function Linemode:solo()
 	end
 
 	if not self[mode] then
-		return ui.Line(" " .. mode .. " ")
+		return ui.Line(mode)
 	end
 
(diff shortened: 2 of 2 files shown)


sxyazi (approved):
Thanks

SolitudeSF:
now it looks even worse
image

sxyazi:
> whats the point of calling it a refactor if it does completely different thing that doesnt address the issue?

Well, people write bugs. Fixed in https://github.com/sxyazi/yazi/commit/fdd574997bf237a20dde31c9c7a28b3c47345cfc

- `require("git"):setup { order = 500 }`

screenshot-001989

- `require("git"):setup { order = 1500 }`:

screenshot-001990

Both look fine to me

SolitudeSF:
this is how this example configuration looks currently
```lua
Linemode:children_add("mtime", 1001)
Linemode:children_add(function(self) if self._file.cha.is_dir then return ui.Line("+1") end end, 999)
```
image

you either need to enforce spaces when rendering, or allow spaces to be added by configuration of `children_add`. your example only looks fine in this exact configuration you have.

sxyazi:
If you need a space, you shouldn't directly add `mtime` to `Linemode._children` - it's only meant to be exposed to users so they can easily override it to implement their own `mtime` behavior with something like `Linemode.mtime = function() end`. 

Instead, you should use a wrapper like `solo` as a child element, that's the reason why `sole` was initially introduced - `mtime` is an implementation of linemode, while `solo` is a renderable child element, that's the difference between them.

SolitudeSF:
if every element needs to add space manually to not blend together, why isnt it added automatically by rendering system? why do i need to slap arbitrary space on an arbitrary side that isnt relevant to the content?

sxyazi:
> if every element needs to add space manually to not blend together, why isnt it added automatically by rendering system? why do i need to slap arbitrary space on an arbitrary side that isnt relevant to the content?

Each element is independent and should have the ability to fully customize its own separator behavior to meet different user needs (whether to have a separator or not, using space, hyphen, or slash, or other styles such as color, etc.).

> and if user doesnt want there to be space, does every plugin need to expose space padding as configuration instead of it being feature of linemode?

Yes, this is similar to how the `git.yazi` plugin exposes the `order` option. The plugin itself should provide a default and reasonable configuration based on Yazi's presets, while also allowing advanced users to make more nuanced adjustments if needed.

> this is not composable

For more complex combinations, it should work like the current `mtime`, where the plugin exposes a linemode implementation (rather than a rendering function). Then users should be allowed to have something like:

```lua
Linemode:children_add(function()
  ui.Line { ui.Span(" "), require("git"):a(), ui.Span("-"):fg("gray"), require("git"):b():bold() }
end, 2000)
```

or a further level of abstraction.

SolitudeSF:
well, i was basing my plugin and expectations of yours. its not composable, space is hardcoded and the function that returns the actual content isnt exposed. user needs to manually reimplement the lambda you pass to `children_add` to get the customization he wants, even a simple one like custom separator. so im guessing its not how it should actually be done and proper api doesnt exist yet.

the fact the plugin itself should manage its own separator should itself be pretty ridiculous.

sxyazi:
> its not composable, padding space is hardcoded and the function that returns the actual content isnt exposed, so user needs to manually reimplement the lambda you pass to children_add to get the customization he wants, even a simple one like custom separator

> make more nuanced adjustments ***if needed***

I've already mentioned that it depends on the user's needs - if users haven't customized their requirements further, then `git.yazi` won't include it, and it also depends on whether the plugin author wants to support it. But from the perspective of the plugin system design, it's entirely possible.

> the fact the plugin itself should manage its own separator should itself be pretty ridiculous

If you are unhappy with this design, you are welcome to a full refund.

Issue #1548, opened 2024-08-24
Make the builtin `extract` and `archive` plugins handle `*.tar` formats better


xfzv:
### What system are you running Yazi on?

Linux X11

### What terminal are you running Yazi in?

kitty 0.35.2

### `yazi --debug` output

```Shell
Yazi
    Version: 0.3.1 (69c20d4 2024-08-24)
    Debug  : false
    OS     : linux-x86_64 (unix)

Ya
    Version: 0.3.1 (69c20d4 2024-08-24)

Emulator
    Emulator.via_env: ("xterm-kitty", "")
    Emulator.via_csi: Ok(Kitty)
    Emulator.detect : Kitty

Adapter
    Adapter.matches: Kitty

Desktop
    XDG_SESSION_TYPE: Some("tty")
    WAYLAND_DISPLAY : None
    DISPLAY         : Some(":0")

SSH
    shared.in_ssh_connection: false

WSL
    /proc/sys/fs/binfmt_misc/WSLInterop: false

Variables
    SHELL              : Some("/bin/zsh")
    EDITOR             : Some("nvim")
    VISUAL             : Some("nvim")
    YAZI_FILE_ONE      : None
    YAZI_CONFIG_HOME   : None

Text Opener
    default: Some(Opener { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block  : Some(Opener { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })

Multiplexers
    TMUX               : false
    tmux version       : No such file or directory (os error 2)
    ZELLIJ_SESSION_NAME: None
    Zellij version     : No such file or directory (os error 2)

Dependencies
    file             : 5.45
    ueberzugpp       : No such file or directory (os error 2)
    ffmpegthumbnailer: 2.2.2
    magick           : 7.1.1-25
    fzf              : 0.53.0
    fd               : 10.1.0
    rg               : 14.1.0
    chafa            : No such file or directory (os error 2)
    zoxide           : 0.9.4
    7z               : 16.02
    7zz              : No such file or directory (os error 2)
    jq               : 1.7.1


--------------------------------------------------
When reporting a bug, please also upload the `yazi.log` log file - only upload the most recent content by time.
You can find it in the "/home/xfzv/.local/state/yazi" directory.
```

### Did you try the latest nightly build to see if the problem got fixed?

[...]

PR #1583, opened 2024-09-01
feat: make the builtin `extract` plugin support compressed tarballs (`*.tar.gz`, `*.tar.bz2`, etc.)

DirkFi:
Close [#1548](https://github.com/sxyazi/yazi/issues/1548).

Description:
 * Add `spawn_tar` function and `is_tar_compatible` function
   - Create `spawn_tar` function to run command with `tar -xaf` to unzip file to tmp directory
   - Create `is_tar_compatible` function to tell if a zip file can be extracted with `tar` command
 * Add `os.excute()` line to make sure the tmp dir exists because `tar` command will not automatically create that dir, unlike `7z`
 * Refactor the logic in the `try()` function to if the zip file can be unzipped with tar and no password required, use `tar`, otherwise use `7z` instead

### DIFF
--- a/yazi-plugin/preset/plugins/archive.lua
+++ b/yazi-plugin/preset/plugins/archive.lua
@@ -53,6 +53,17 @@ function M:seek(units)
 	end
 end
 
+function M:spawn_tar(args)
+	local command = "tar"
+
+	local stdout = args[1] == "l" and Command.PIPED or Command.NULL
+	local child, code = Command(command):args(args):stdout(stdout):stderr(Command.PIPED):spawn()
+	if not child then
+		return nil, "Failed to spawn, error code: " .. tostring(code)
+	end
+	return child, nil
+end
+
 function M:spawn_7z(args)
 	local last_error = nil
 	local try = function(name)
--- a/yazi-plugin/preset/plugins/archive.lua
+++ b/yazi-plugin/preset/plugins/archive.lua
@@ -54,10 +54,8 @@ function M:seek(units)
 end
 
 function M:spawn_tar(args)
-	local command = "tar"
-
 	local stdout = args[1] == "l" and Command.PIPED or Command.NULL
-	local child, code = Command(command):args(args):stdout(stdout):stderr(Command.PIPED):spawn()
+	local child, code = Command("tar"):args(args):stdout(stdout):stderr(Command.PIPED):spawn()
 	if not child then
 		return nil, "Failed to spawn, error code: " .. tostring(code)
 	end
--- a/yazi-plugin/preset/plugins/archive.lua
+++ b/yazi-plugin/preset/plugins/archive.lua
@@ -54,7 +54,7 @@ function M:seek(units)
 end
 
 function M:spawn_tar(args)
-	local stdout = args[1] == "l" and Command.PIPED or Command.NULL
+	local stdout = args[1] and args[1]:sub(0, 2) == "-t" and Command.PIPED or Command.NULL
 	local child, code = Command("tar"):args(args):stdout(stdout):stderr(Command.PIPED):spawn()
 	if not child then
 		return nil, "Failed to spawn, error code: " .. tostring(code)
--- a/yazi-plugin/preset/plugins/archive.lua
+++ b/yazi-plugin/preset/plugins/archive.lua
@@ -53,15 +53,6 @@ function M:seek(units)
 	end
 end
 
-function M:spawn_tar(args)
-	local stdout = args[1] and args[1]:sub(0, 2) == "-t" and Command.PIPED or Command.NULL
-	local child, code = Command("tar"):args(args):stdout(stdout):stderr(Command.PIPED):spawn()
-	if not child then
-		return nil, "Failed to spawn, error code: " .. tostring(code)
-	end
-	return child, nil
-end
-
 function M:spawn_7z(args)
 	local last_error = nil
 	local try = function(name)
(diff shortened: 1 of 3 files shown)


yazi-plugin/preset/plugins/extract.lua
```diff
@@ -47,23 +57,45 @@ function M:try_with(url, pwd)
 		fail("Failed to determine a temporary directory for %s", url)
 	end
 
+	os.execute("mkdir -p " .. tostring(tmp))
```
sxyazi, line 60: Please add a new method `fs.create(type, url)` above and similar to the `fs.remove()` method here and use it instead to achieve cross-platform compatibility https://github.com/sxyazi/yazi/blob/640e497c358c5ef576643e646cdb73be67bf7c52/yazi-plugin/src/fs/fs.rs#L36-L54

- type: available values `"file"`, `"dir"`
- url: Url to operate

yazi-plugin/preset/plugins/extract.lua
```diff
@@ -47,23 +57,45 @@ function M:try_with(url, pwd)
...
+			fail("Failed to spawn `tar` command, error code %s", code)
+		end
 
-	local output, err = child:wait_with_output()
-	if output and output.status.code == 2 and archive:is_encrypted(output.stderr) then
-		fs.remove("dir_clean", tmp)
-		return true -- Needs retry
-	end
+		local output, err = child:wait_with_output()
+		if output and output.status.code == 2 and archive:is_encrypted(output.stderr) then
```
sxyazi, line 69: `archive:is_encrypted(output.stderr) `, Can `tar` be encrypted?

yazi-plugin/preset/plugins/extract.lua
```diff
@@ -47,23 +57,45 @@ function M:try_with(url, pwd)
...
 	end
 
+	os.execute("mkdir -p " .. tostring(tmp))
 	local archive = require("archive")
-	local child, code = archive:spawn_7z { "x", "-aou", "-p" .. pwd, "-o" .. tostring(tmp), tostring(url) }
-	if not child then
-		fail("Spawn `7z` and `7zz` both commands failed, error code %s", code)
-	end
+	if self:is_tar_compatible(tostring(url)) then
+		local child, code = archive:spawn_tar { "-xvf", tostring(url), "-C", tostring(tmp) }
```
sxyazi, line 63:
```suggestion
		local child, code = archive:spawn_tar { "-xf", tostring(url), "-C", tostring(tmp) }
```

Any reason for specifying the `-v` switch?

yazi-plugin/preset/plugins/extract.lua
```diff
@@ -36,6 +36,16 @@ function M:entry(args)
 	end
 end
 
+function M:is_tar_compatible(url)
+	local tar_extensions = { ".tar", ".tgz", ".tar.gz", ".tar.bz2" }
```
sxyazi, line 40: Are you sure `tar -xvf` also works for `.tar.gz` and `.tar.bz2` files?

yazi-plugin/preset/plugins/archive.lua
```diff
@@ -53,6 +53,17 @@ function M:seek(units)
 	end
 end
 
+function M:spawn_tar(args)
+	local command = "tar"
```
sxyazi, line 57:
```suggestion
```

sxyazi (changes requested):

yazi-plugin/preset/plugins/extract.lua
DirkFi: Tar does not have any built-in encryption like 7z, needs to be used with gpg or other tools.

yazi-plugin/preset/plugins/extract.lua
DirkFi: Ur right, it's an optional one for printing the output of filename, gonna delete this.

DirkFi:
BTW, Do I need to combine the other commits or I can have separate commit in the same PR?

veltza:
There are more [suffixes](https://en.wikipedia.org/wiki/Tar_(computing)#Suffixes_for_compressed_files) than ".tar", ".tgz", ".tar.gz", ".tar.bz2". And short forms should be supported as well.

sxyazi:
> BTW, Do I need to combine the other commits or I can have separate commit in the same PR?

You can make multiple commits in this PR, and I'll squash and merge them into one.

DirkFi:
> There are more [suffixes](https://en.wikipedia.org/wiki/Tar_(computing)#Suffixes_for_compressed_files) than ".tar", ".tgz", ".tar.gz", ".tar.bz2". And short forms should be supported as well.

Thanks! Definitely will go test and add this!

yazi-plugin/preset/plugins/extract.lua
DirkFi: Yeah, it did work. But think it's still necessary to change to `-a` to make it  `--auto-compress` to increase the compatibility.

veltza:
Is this going to fix the preview issues with tar files as well?

DirkFi:
> Is this going to fix the preview issues with tar files as well?

What issue specifically? If you're talking about preview does not show the final file of the `tar.gz` or other type of `tar`, no. It is not fixed in this PR.

veltza:
> What issue specifically? If you're talking about preview does not show the final file of the tar.gz or other type of tar, no. It is not fixed in this PR.

That's what I was referring to. And I hoped that this would have killed two birds with one stone. But that's okay because I think it will be easy to fix after this PR.

sxyazi:
I finally had time to test it today (sorry for the delay), I found that we can directly reuse 7zip to extract tar files, which eliminates the need for an additional dependency. Plus, 7zip is more cross-platform friendly (some older Windows systems can also use it)

sxyazi (approved):
Thanks!

Issue #1570, opened 2024-08-29
Bug: Yazi takes forever (30+ seconds) to load at the *new* Windows Terminal Preview 1.22.2362.0


MoaidHathot:
### What system are you running Yazi on?

Windows

### What terminal are you running Yazi in?

Windows Terminal Preview 1.22.2362.0

### `yazi --debug` output

```Shell
Yazi
    Version: 0.3.2 (ddb8ce5 2024-08-28)
    Debug  : false
    OS     : windows-x86_64 (windows)

Ya
    Version: 0.3.2 (ddb8ce5 2024-08-28)

Emulator
    Emulator.via_env: ("", "")
    Emulator.via_csi: Ok(Unknown([]))
    Emulator.detect : Unknown([])

Adapter
    Adapter.matches: Chafa

Desktop
    XDG_SESSION_TYPE: None
    WAYLAND_DISPLAY : None
    DISPLAY         : None

SSH
    shared.in_ssh_connection: false

WSL
    /proc/sys/fs/binfmt_misc/WSLInterop: false

Variables
    SHELL              : None
    EDITOR             : None
    VISUAL             : None
    YAZI_FILE_ONE      : Some("C:\\Program Files\\Git\\usr\\bin\\file.exe")
    YAZI_CONFIG_HOME   : None

Text Opener
    default: Some(Opener { run: "code %*", block: false, orphan: true, desc: "code", for_: None, spread: true })
    block  : Some(Opener { run: "code -w %*", block: true, orphan: false, desc: "code (block)", for_: None, spread: true })

Multiplexers
    TMUX               : false
    tmux version       : program not found
    ZELLIJ_SESSION_NAME: None
    Zellij version     : program not found

Dependencies
    file             : 5.45
    ueberzugpp       : program not found
    ffmpegthumbnailer: program not found
    magick           : program not found
    fzf              : 0.52.1
    fd               : 10.1.0
    rg               : 14.1.0
    chafa            : program not found
    zoxide           : program not found
    7z               : program not found
    7zz              : program not found
    jq               : 1.7.1
```

### Did you try the latest nightly build to see if the problem got fixed?

No

### Describe the bug

Yazi takes more than 30 seconds to load when used with the new Windows Terminal Preview 1.22.2362.0.
This issue doesn't reproduce with older Preview versions and with the non-preview version.

Logs:
```
  2024-08-29T21:26:15.499134Z  WARN yazi_adapter::emulator: [Adapter] No special environment variables detected
[...]

PR #1588, opened 2024-09-01
feat: support image preview in Windows Terminal

e82eric:
Hi!  The latest version of windows terminal supports Sixel (https://github.com/microsoft/terminal/releases/tag/v1.22.2362.0) https://github.com/microsoft/terminal/pull/17421.

I wanted to make an attempt at updating Yazi to add support.  This looked like it would work for both images and video.  This may cause a regression for older versions of windows terminal since it won't fallback to chafa.

terminal-sixel


sxyazi:
Yazi already supports Sixel through the standard [DA1 request](https://vt100.net/docs/vt510-rm/DA1.html). This means that as long as MS Terminal correctly sends the response indicating Sixel support, Yazi can recognize it without needing any additional detection.

But the DA1 response in Windows Terminal Preview 1.22.2362.0 is [currently broken](https://github.com/sxyazi/yazi/issues/1570#issuecomment-2319625368), so I can accept this PR as a temporary workaround until they fix it.

e82eric:
Ahh,  I was wondering about that exact part of the code,  it was taking 10s to load and I saw that it was waiting for that da1 and timing out.  I will try and pull down that windows terminal branch and give it a try.

sxyazi (approved):
Thanks!

sxyazi:
Hi @e82eric, could you send me an mp4 version of the PR gif demo? If possible, it would be great if you could make the window a bit smaller. 

I want to use it as the demo video for the upcoming version 0.3.3 release - Yazi is probably the first terminal file manager to support image preview in MS Terminal, especially when the standard DA1 isn't available now hah😄

e82eric:
yep,  working on the mp4 now.

Quick update on the MS fix,  I tried that branch without my changes,  and the first attempt still tried to use chafa, I commented out this if statement and it correctly detected Sixel https://github.com/sxyazi/yazi/blob/e940d81d2a0a100fee41c1a59af7e2a14b49077f/yazi-adapter/src/adapter.rs#L95.  Maybe when the MS fix is merged that can be updated to just check for Tmux?

By the way, your project is awesome!

sxyazi:
That's weird, this statement will only be entered if the `ZELLIJ_SESSION_NAME` env variable is detected:

https://github.com/sxyazi/yazi/blob/e940d81d2a0a100fee41c1a59af7e2a14b49077f/yazi-adapter/src/adapter.rs#L94-L96

Are you running Yazi in Zellij?

e82eric:
Not running it in Zellij,  will try to step through it to see what was happening.  I was thinking it removed Sixel from everything that wasn't Zellij.

How does this look?

https://github.com/user-attachments/assets/22900e7b-9193-409f-b6da-e0afba1afda1

sxyazi:
Looks great, thanks a lot!

e82eric:
Quick update.  it looks like it was just this line that needed to commented out.  I think it is removing everything that isn't iterm2.

https://github.com/sxyazi/yazi/blob/e940d81d2a0a100fee41c1a59af7e2a14b49077f/yazi-adapter/src/adapter.rs#L93

sxyazi:
Ah that makes sense, this is to ensure compatibility with third-party terminals that rely on the older version of ConPTY (like WezTerm and Mintty) for displaying images, so it still needs to be kept for now, and as time goes on and with OS updates, we can remove it in the future.

e82eric:
That makes sense,  side note, start time was back to 0 on the branch with the MS fix.

sxyazi:
Thanks for testing, I've edited the content of this PR and added some notes about it

PR #1593, opened 2024-09-02
fix: log only really need warning

TobisLee:
When I use `foot` terminal, I find these logs:
```txt
  2024-09-02T09:08:18.831969Z  WARN yazi_adapter::emulator: [Adapter] No special environment variables detected
    at yazi-adapter/src/emulator.rs:70

  2024-09-02T09:08:18.831983Z  WARN yazi_adapter::emulator: [Adapter] Unknown TERM_PROGRAM: 
    at yazi-adapter/src/emulator.rs:84
```
I know these logs are useful for debugging terminal issues, but I just want to delay the log time. I find them redundant when using the `foot` terminal, and they're a bit annoying since I open `yazi` thousands of times a day.

### DIFF
--- a/yazi-adapter/src/emulator.rs
+++ b/yazi-adapter/src/emulator.rs
@@ -58,6 +58,8 @@ impl Emulator {
 			return Self::Neovim;
 		}
 
+		let mut warnings = Vec::new();
+
 		let vars = [
 			("KITTY_WINDOW_ID", Self::Kitty),
 			("KONSOLE_VERSION", Self::Konsole),
@@ -70,7 +72,7 @@ impl Emulator {
 		];
 		match vars.into_iter().find(|v| env_exists(v.0)) {
 			Some(var) => return var.1,
-			None => warn!("[Adapter] No special environment variables detected"),
+			None => warnings.push("[Adapter] No special environment variables detected"),
 		}
 
 		let (term, program) = Self::via_env();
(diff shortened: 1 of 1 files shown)


yazi-adapter/src/emulator.rs
```diff
@@ -70,7 +72,7 @@ impl Emulator {
 		];
 		match vars.into_iter().find(|v| env_exists(v.0)) {
 			Some(var) => return var.1,
-			None => warn!("[Adapter] No special environment variables detected"),
+			None => warnings.push("[Adapter] No special environment variables detected"),
```
sxyazi, line 75: It's tricky to implement it this way, which means that if any of the subsequent conditions are met and an early return occurs, this log will never be printed.

sxyazi:
I think it would make more sense to add a new `level` option in the `yazi.toml`, if you don't want these logs just set:

```toml
[log]
level = "off"
```

TobisLee:
Log setting might not what I want, I don't want to ignore these warnings.

For example, contour terminal isn't in yazi emulators list, so when I open yazi in contour, it's reasonable and acceptable to print warnings in yazi.log. If I set the log level to `off`, then I won't see these warnings when opening yazi in contour.  On the other hand, since foot is supported, showing these logs seems unnecessary and might be misleading. When I first saw these logs while using foot, I thought I needed to adjust some settings to get Yazi working correctly in foot.

sxyazi:
> If I set the log level to `off`

No, you don't necessarily need to set it to "off"; you can also set it to "error". Just keep in mind that this log is only a warning level when the environment variable doesn't exist, and it's not an error.

I think it's perfectly fine as it is, if you feel the warning is too much, we can consider adjusting it to "debug" or "trace," but I just don't think it's necessary.

TobisLee:
> No, you don't necessarily need to set it to "off"; you can also set it to "error".

It's not just about the logging level I set; just the log settings don't quite fit my use case.

> I think it's perfectly fine as it is, if you feel the warning is too much, we can consider adjusting it to "debug" or "trace," but I just don't think it's necessary.

Fair enough, I also agree that changing the log level is not a good idea. I'll look for a more reasonable way to make it work comfortably for me.

Huge thanks for the explanation and help. I'll close this PR.

PR #1603, opened 2024-09-04
feat: add `is_absolute`, `has_root`, `starts_with`, `ends_with`, `strip_prefix` to `Url` Userdata

sxyazi:


sxyazi:
Need to think about the API a bit more. Reverted.

Issue #836, opened 2024-03-22
Yazi v0.3 TODO List


sxyazi:
### Please describe the problem you're trying to solve

a

### Would you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

a

AntoniosBarotsis:
For `ffmpegthumbnailer`, I was wondering what difficulties using `ffmpeg` instead could entail. I'm guessing that [this](https://github.com/sxyazi/yazi/blob/f2329a3b35f6f074e4029e19c5166e90872470d7/yazi-plugin/preset/plugins/video.lua#L37-L50) is the relevant code.

Wouldn't it be possible to use something like [this](https://trac.ffmpeg.org/wiki/Create%20a%20thumbnail%20image%20every%20X%20seconds%20of%20the%20video)? I am blissfully unaware of the complexities of anything ffmpeg related 😅

sxyazi:
> what difficulties using ffmpeg instead could entail

Some Windows users have [reported to me](https://github.com/sxyazi/yazi/issues/51#issuecomment-1874657020) that they cannot successfully build `ffmpegthumbnailer` on Windows. As a result, video previews are unavailable to Yazi's Windows users, and it is indeed in a state of lack of maintenance. 

Bud yesterday I tried `ffmpeg` and found that there's no easy way to generate thumbnails by percentage, only by precise time. Since `ffmpeg` is the only alternative to `ffmpegthumbnailer` that I know of (if there are others, please let me know), it might not happen until we find any good alternative solutions.

sxyazi:
All done except `ffmpegthumbnailer`, closing this todo list

AntoniosBarotsis:
@sxyazi if the percentage is the hold up then couldn't we get the video's duration with 1 ffmpeg call and then figure out exactly when you want to get the thumbnail? Would the 2 calls needed be problematic for performance? In that case maybe we could look into the sdk instead of spawning an external process. 

Not sure if I misunderstood what you meant by percentage. I'm guessing you meant something like "give me the frame at the 20% point of the video duration" but I could be completely wrong. 

I'm on windows myself so I'm trying to figure out if this is within my capabilities to implement 😅

PR #1616, opened 2024-09-06
Add fallback video thumbnailer for windows

e82eric:
Hi,  I had difficulty figuring out how to get ffmpegthumbnailer working on windows (both mysys and msvc eventually worked but required making some changes).

This adds a "fallback" thumbnailer that only implements the parameters needed for Yazi's thumbnails to work.  This should allow for thumbnails to work on windows with a base install of Yazi.

Totally understand if this pull request does not make sense for Yazi and you need to close it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "aes"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


sxyazi:
Sorry, I'm afraid I can't accept this PR. Bundling `ffmpeg` with Yazi seems a bit overkill to me, as it would increase build time and size. Besides, we already have plugins like https://github.com/Tyarel8/video-ffmpeg.yazi, which allows Windows users to use their existing `ffmpeg` installations directly. 

In the future, we can also consider switching Yazi's dependency from `ffmpegthumbnailer` to `ffmpeg`. In fact, I already thought about this in version 0.3, but gave it up because ffmpeg [doesn't support generating thumbnails by percentage](https://github.com/sxyazi/yazi/issues/836#issuecomment-2116679289). But after seeing `video-ffmpeg.yazi`, I got some inspiration, as we can always get the total length of the video first and then calculate it manually.

e82eric:
Totally understand, thank you for checking!

PR #1650, opened 2024-09-16
feat: change input title when using `create --dir`

Tyarel8:
When using `create` the title will rename the same (`Create:`), but when using the new `--dir` option, it will show `Create (dir):` to show that it is different in the input prompt.

### DIFF
--- a/yazi-config/preset/yazi.toml
+++ b/yazi-config/preset/yazi.toml
@@ -139,7 +139,7 @@ cd_origin = "top-center"
 cd_offset = [ 0, 2, 50, 3 ]
 
 # create
-create_title  = "Create:"
+create_title  = [ "Create:", "Create (dir):" ]
 create_origin = "top-center"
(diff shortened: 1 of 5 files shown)


yazi-config/preset/yazi.toml
```diff
@@ -139,7 +139,7 @@ cd_origin = "top-center"
 cd_offset = [ 0, 2, 50, 3 ]
 
 # create
-create_title  = "Create:"
+create_title  = [ "Create:", "Create (dir):" ]
```
sxyazi, line 142: This is a breaking change, which will cause a crash for users whose existing config contains `create_title = "Create:"`, as the type has changed, and Yazi can't parse it.

Please support both array and string formats, and print a warning when a string is detected to prompt users to switch to the new array format.

sxyazi (changes requested):

sxyazi (approved):
Thanks

PR #1680, opened 2024-09-25
feat!: make `backspace` command not close the input even when value is empty

XYenon:


yazi-config/preset/keymap.toml
```diff
@@ -232,10 +232,10 @@ keymap = [
 	{ on = "<End>",  run = "move 999",  desc = "Move to the EOL" },
 
 	# Delete
-	{ on = "<Backspace>", run = "backspace",	       desc = "Delete the character before the cursor" },
-	{ on = "<Delete>",    run = "backspace --under", desc = "Delete the character under the cursor" },
-	{ on = "<C-h>",       run = "backspace",         desc = "Delete the character before the cursor" },
-	{ on = "<C-d>",       run = "backspace --under", desc = "Delete the character under the cursor" },
+	{ on = "<Backspace>", run = "backspace --close",         desc = "Delete the character before the cursor" },
```
sxyazi, line 235: This is a breaking change that will alter the behavior for users with just `backspace --under` in their existing config. 

Also could you explain the necessity of adding it? "close when there's no more content" is meant to better mimic the behavior of the Vim command line prompt.

sxyazi (changes requested):

yazi-config/preset/keymap.toml
XYenon: When I want to delete all content of an `input`, I usually hold `backspace` for a while. And then I will to input something, but the `input` has already been closed, all my inputs will act as keymap.

XYenon:
I'm not always decide to delete all content when I start to delete them. Usually when I have deleted half of the content, I realized the rest of them is wrong too, so I'll keep `backspace` to delete all of them.

yazi-config/preset/keymap.toml
XYenon: What about change this option to `--no-close` to make it not break?

sxyazi:
I just got another report about this issue: https://github.com/sxyazi/yazi/issues/1681. It seems like enough people are bothered by this behavior, so I think we can just remove it since it's a breaking change anyway, and it's also hard to come up with a way (like warnings) to prompt users with the [existing `backspace --under`](https://github.com/search?q=%22backspace+--under%22&type=code) config to replace it

XYenon:
What about change this option to `--no-close` to make it not break?

sxyazi:
No, just remove it - if enough people are bothered by it, then its worth breaking. 

I'll also keep this PR open for a while to see if there are any differing opinions

sxyazi (approved):
Thanks!

PR #1684, opened 2024-09-25
fix: `magick` previewer for multi-layered image files with `-flatten` argument

v3natio:
As the title describes it, this is an easy fix in order for the previewer to work for `xcf` and other files which handle layers. It should work as normal for non-layered files.

### DIFF
--- a/yazi-plugin/preset/plugins/magick.lua
+++ b/yazi-plugin/preset/plugins/magick.lua
@@ -23,6 +23,7 @@ function M:preload()
 		"-density",
 		"200",
 		tostring(self.file.url),
+		"-flatten",
 		"-resize",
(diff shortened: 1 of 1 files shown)


sxyazi:
Could you please send me a file with layers for me to test?

v3natio:
Alright, so for example for the following `.xcf` file (uploading it as zip since github won't let me otherwise):
[test.zip](https://github.com/user-attachments/files/17137629/test.zip)

We get the following previews using the old command (no `-flatten` arg):
idk-0
idk-1
idk-2
idk-3

And we get a singular image using the arg:
idk

sxyazi (approved):
Tested and works great, thanks for the fix! 👍🏻

PR #1688, opened 2024-09-26
fix: enable texts of `search` related commnds to show in plural form when mutli files are selected.

yanCode:
## Issue Addressed

Previously, when pressing the `c` key, the text displayed was always in singular form, even when multiple files were selected:

```
Copy the file path
Copy the directory path
Copy the filename
```

## Solution

After this fix, when multiple files are selected and the `c` key is pressed, the nouns are converted to plural form:

```
Copy the file paths
Copy the directory paths
Copy the filenames
```

The original singular form is maintained if only one file is selected (or hovered).

## Implementation Details

1. Enabled the `desc` text field in `keymap.toml` to support formats like `wol{ves|f}` or `name{s}`, accommodating both regular and irregular plural forms.

2. The new system works as follows:
   a. Matches content inside `{}` and splits it.
   b. For plural form: Uses the first element of the split to replace the whole `{}`.
      Example: `wol{ves|f}` => `wolves`, `name{s}` => `names`
   c. For singular form: 
      - Uses the second element of the split if available.
      - If no second element (no `|`), removes the entire `{}`.
      Example: `name{s}` => `name`, `wol{ves|f}` => `wolf`

3. Updated the logic to call `selected_or_hovered` in advance  when the `c` key is pressed, ensuring accurate capture of selected file count in the `Which` window.

4. For the `Help` window, the default plural form is used to display help information.

## Additional Notes
I'd be much appericiated if you can provide me any feedback. To be honest, this PR is slightly more complicated than I initially though, But it helps me know lots more cool features of this project.

### DIFF
--- a/yazi-config/preset/keymap.toml
+++ b/yazi-config/preset/keymap.toml
@@ -89,10 +89,10 @@ keymap = [
 	{ on = [ "m", "n" ], run = "linemode none",        desc = "Linemode: none" },
 
 	# Copy
-	{ on = [ "c", "c" ], run = "copy path",             desc = "Copy the file path" },
-	{ on = [ "c", "d" ], run = "copy dirname",          desc = "Copy the directory path" },
-	{ on = [ "c", "f" ], run = "copy filename",         desc = "Copy the filename" },
-	{ on = [ "c", "n" ], run = "copy name_without_ext", desc = "Copy the filename without extension" },
+	{ on = [ "c", "c" ], run = "copy path",             desc = "Copy the file path{s}" },
+	{ on = [ "c", "d" ], run = "copy dirname",          desc = "Copy the directory path{s}" },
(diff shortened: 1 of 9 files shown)


sxyazi:
Sorry, but I'm afraid I can't accept this PR, it's too complex for me for this single feature, and every time you get the key description, it requires running a regex match, which is time-consuming. Plus, when there's a match, it needs to replace them, requiring new memory allocation. 

Passing `selected_or_hovered` to the `which` component and the config parsing also somewhat breaks the DDD domain boundaries, which is something I wanted to avoid in the design.

If we really want to fix this issue, the simplest way is to change the description to "Copy the file path(s)" as users usually know what they're doing - whether they've selected multiple files, so that should be enough.

yanCode:
i see rustfmt errors in git action, I am fixing it now.

yanCode:
no problem, let me close this PR.

PR #1690, opened 2024-09-26
feat: adapt for image preview in the Rio terminal

raphamorim:
If possible would like to add Rio to the terminal list 🙏 

Thank you for yazi :D 

Screenshot 2024-09-26 at 17 18 01


sxyazi:
Hey @raphamorim, I pushed a new commit https://github.com/sxyazi/yazi/pull/1690/commits/853f507b969fd31636d016f9d888250ec19de93d. 

Could you review it and check if the way of detecting Rio is correct? Also, are there any other conditions we could add to detect Rio too, like special env variables starting with `RIO_` or the `\x1b[>q` sequence? Let me know if there's anything I missed!

raphamorim:
> Could you review it and check if the way of detecting Rio is correct? Also, are there any other conditions we could add to detect Rio too, like special env variables starting with RIO_ or the \x1b[>q sequence? Let me know if there's anything I missed!

looks correct 👍 rio also injects TERM_PROGRAM as equal rio in the env by default

sxyazi:
Nice, merging now

raphamorim:
Thank you @sxyazi 🙏

PR #1718, opened 2024-10-02
feat: yazi-cli: support different Git hosting services

dm9pZCAq:
add support for different Git hosting services:
```sh
ya pack -a codeberg.org/owner/repo-name:plugin

# by default github.com will be used
ya pack -a owner/repo-name:plugin
```

This change is compatible with existing `package.toml`

### DIFF
--- a/yazi-cli/src/package/package.rs
+++ b/yazi-cli/src/package/package.rs
@@ -1,13 +1,33 @@
-use std::{borrow::Cow, io::BufWriter, path::PathBuf};
+use std::{borrow::Cow, io::BufWriter, path::PathBuf, sync::OnceLock};
 
-use anyhow::{Context, Result};
+use anyhow::{Error, Result};
 use md5::{Digest, Md5};
+use regex::Regex;
 use yazi_shared::Xdg;
 
+static PACKAGE_RAW_URL_RE: OnceLock<Regex> = OnceLock::new();
+static PACKAGE_SHORT_URL_RE: OnceLock<Regex> = OnceLock::new();
+
+#[inline]
+fn package_raw_url_re() -> &'static Regex {
+	PACKAGE_RAW_URL_RE.get_or_init(|| {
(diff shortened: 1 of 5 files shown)


yazi-cli/src/package/package.rs
```diff
@@ -55,8 +70,7 @@ impl Package {
 
 	#[inline]
 	pub(super) fn remote(&self) -> String {
-		// Support more Git hosting services in the future
-		format!("https://github.com/{}.git", self.repo)
+		format!("https://{}/{}/{}.git", self.host, self.owner, self.repo_name)
```
sxyazi, line 73: I'm not sure if using the three-level structure of `{host}/{owner}/{repo}` here is safe - this is also why I haven't supported it yet. Are there Git providers that don't use this format, for example `{host}/xxx/{owner}/{repo}`?

Another issue is that I think we should consider supporting the SSH protocol, as it would allow users to manage their private plugins.

dm9pZCAq: I think most of the popular ones uses three-level structure

here is some examples:
- git.sr.ht
- codeberg.org
- notabug.org
- gitlab.com
- **gitea**
- ...

> Another issue is that I think we should consider supporting the SSH protocol

I think this and support for not non three-level git repos can be done later in a separate PR

sxyazi: What are your thoughts on non-3-level structures and SSH? Do they keep backward compatibility with the format of this PR? 

My main concern is that once we support this PR's format, if we later need to change the format to support non-3-level and SSH, and their formats aren't compatible, it would be a breaking change. 

Users would need to switch to the new format when using `ya pack -a`, and some have integrated `ya pack` into their automation scripts. Plus, it's written into `package.toml`, which could require some level of migration work like [this](https://github.com/sxyazi/yazi/blob/8c5d8aa5893529cf73929b7955fb7182a0df8ac5/yazi-cli/src/package/parser.rs#L9-L34). 

So I prefer to fully discuss everything in one place before implementing as it's usually better than rolling out an incomplete feature.

dm9pZCAq:
@sxyazi, I added support for raw git urls and tests with examples

and it is backward compatible

PR #1777, opened 2024-10-13
feat: added the create function to the fs module

### LINKED ISSUE #1709: Expose file and directory creation to plugin developers as part of `fs`
### `yazi --debug` output

```Shell
N/A
```

There is already a function to remove files and directories inside the `fs` module, so having a function to create them would make it more convenient for plugin developers to create files and directories through Yazi instead of relying on shell commands.

Ideally, the function would be similar to the `create` file manager command that takes an `Url` as an input and a boolean argument to indicate whether a directory or file should be created, and maybe an additional argument for whether or not to overwrite the existing files, or rename the new files.

### Additional context

For context, I am trying to create a custom version of the `create` file manager command to suit my needs, and exposing an API to create files and directories would make this possible without any external dependencies and will work on all systems.

Alternatively, having the `create` file manager command take a given path to create would also work.

### Validations

hankertrix:
It is mostly based on the create_do function.

Closes #1709.


sxyazi:
Sorry, I'm afraid I can't accept this PR because the current implementation is too specific to a single feature and lacks broader usability.

Creating files/directories is a bit complex and requires some design at the API level. I haven't had the time to dive into it yet, but I do have some rough ideas:

- For directory creation, the default behavior should be non-recursive, with an option for recursive creation.
- There shouldn't be a dedicated API for creating files. Instead, files should be opened, returning a handle that can be used for chained operations like writing, flushing, etc.
- When opening a file, it should be possible to specify the mode (read/write, append, truncate, etc.). Rust's `OpenOptions` is a good reference: https://doc.rust-lang.org/std/fs/struct.OpenOptions.html.
- The file handle should interoperate with other APIs, such as opening a process via the [Command API](https://yazi-rs.github.io/docs/plugins/utils/#command) and redirecting the process's stdout to the file handle — we can tackle this part later.

Would you like to give this a try in this PR? I can provide some guidance if needed. Or feel free to close this PR, and I'll try to implement it after the 0.4 release, once I get some time.

hankertrix:
Ah I see, no worries. I can try to tackle the first 3 points you mentioned, but it'll take me a lot more time, which I currently don't have much of until about 2 months later.

I'll convert this pull request to a draft.

hankertrix:
Closing this pull request for now, I’ll create another pull request when I implement the API.

PR #1783, opened 2024-10-14
feat: prevent paths from being treated as options in `file` command

gaesa:


sxyazi:
Thanks for the PR! Could you explain a bit what issue it solves? i.e. in what cases and with which filenames would they be treated as options? Also, does it work on non-Unix platforms like Windows?

sxyazi:
> especially those starting with a `-`

Sorry, I still don't quite understand the issue. These file paths are always absolute paths, meaning a filename like just `-foo` could never exist. Instead, it would at least be something like `/-foo`. Could you share which file paths you encountered that couldn't be processed?

sxyazi:
I see your point. My thoughts and concerns are that if something isn't broken, it's better not to touch it. 

Yet since `file` is an external dependency, its behavior might differ across versions or systems - I've noticed some of these differences over time, for example, the version installed via [Scoop doesn't support Unicode filenames](https://yazi-rs.github.io/docs/installation#windows), and [older versions of `file` use `x-` while newer ones don't](https://github.com/sxyazi/yazi/discussions/1663#discussioncomment-10705031).

So, I have some doubts about whether the support for `--` will be consistent as well. That said, I think I'm okay with accepting this PR because we'll never know the answer without trying, I'll go ahead and merge it now.

sxyazi (approved):
Thanks!

PR #1787, opened 2024-10-15
Change naming convention for folders to append index after the name

raj921:
### Summary
This pull request modifies the `unique_name` function in the `path.rs` file to change the naming convention for folders. Instead of inserting an index before the last dot in the folder name, the index is now appended at the end of the folder name. This change aims to improve the clarity and usability of folder names when duplicates are created.

### Changes Made
- Updated the logic in the `unique_name` function to check if the path is a directory.
- For directories, the index is now appended at the end of the name (e.g., `a.b.c` becomes `a.b.c_1`).
- Maintained the existing behavior for file names.

### Testing
- The changes have been tested to ensure that both file and folder naming conventions work as expected.


raj921:
no problem

PR #1789, opened 2024-10-16
feat: confirm component customization

GrzegorzKozub:
While preserving the original design if we don't provide any config, this change let's us customize the confirm component further by:

- hiding the separators and the scrollbar
- changing the button text
- changing the border, title, content, list and button styles

I'm not sure how to add the new config to [the schema](https://yazi-rs.github.io/schemas/theme.json).


sxyazi (changes requested):
Thank you for the PR. Here are a few change requests:

- The "no" button shouldn't use the same style as the "yes" button (i.e., `reversed = true`). This reduces distinction between the two, making it hard to identify the current selection and potentially introducing safety risks.
- There should be a way to style the "no" and "yes" buttons separately, rather than applying the same style to both.
- Please merge the two labels into a single option, like this: `["  [Y]es  ", "  (N)o  "]`.
- Please remove `show_scrollbar` and `show_separators`. Scrollbars only appear when needed (i.e., when the content is long enough to require scrolling), so disabling them isn't very meaningful. As for `show_separators`, it reduces clarity between the content and buttons and isn't visually appealing — this will also make it harder to distinguish once `ya.confirm()` API is introduced since it will allow arbitrary styles and alignment of its content.

GrzegorzKozub:
Hopefully all your requests are now satisfied @sxyazi. This brought us back to the correct default:

image

While still allowing us to customize to fit our themes:

```toml
[confirm]
button_labels = ["Yes", "No"]
border = { fg = "darkgray" }
title = { fg = "darkgray" }
content = { fg = "gray" }
list = { fg = "reset" }
button_yes = { fg = "gray" }
button_no = { fg = "darkgray" }
```

image

sxyazi (approved):
All looks good to me now, thank you so much for contributing this feature @GrzegorzKozub!

GrzegorzKozub:
Thank you!

PR #1801, opened 2024-10-17
feat: prioritize paths that need to be processed first during bulk renaming

### LINKED ISSUE #1766: Analyze the renaming order of bulk-renamed files to avoid conflicts
### `yazi --debug` output

```Shell
Yazi
    Version: 0.3.3 (termux 2024-09-05)
    Debug  : false
    OS     : android-aarch64 (unix)

Ya
    Version: 0.3.3 (termux 2024-09-05)

Emulator
    Emulator.via_env: ("xterm-256color", "")
    Emulator.via_csi: Ok(Unknown([]))
    Emulator.detect : Unknown([])

Adapter
    Adapter.matches: X11

Desktop
    DISPLAY                    : Some(":1")

SSH

WSL
    WSL: false

Variables
    SHELL              : Some("/data/data/com.termux/files/usr/bin/fish")
    EDITOR             : Some("/data/data/com.termux/files/usr/bin/nvim")
    VISUAL             : Some("/data/data/com.termux/files/usr/bin/nvim")
    YAZI_FILE_ONE      : None
    YAZI_CONFIG_HOME   : None

Text Opener
    default: Some(Opener { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block  : Some(Opener { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })

Multiplexers
    TMUX               : false
    ZELLIJ_SESSION_NAME: None

Dependencies
    file             : 5.45
    ffmpegthumbnailer: No such file or directory (os error 2)
    magick           : No such file or directory (os error 2)
    fzf              : 0.55
    fd               : 10.2.0
    rg               : 14.1.1
    zoxide           : 0.9.6
    7z               : 17.05
    jq               : 1.7.1


--------------------------------------------------
When reporting a bug, please also upload the `yazi.log` log file - only upload the most recent content by time.
You can find it in the "/data/data/com.termux/files/home/.local/state/yazi" directory.
```

when trying to bulk rename files yazi skips renaming if desnination file already exists, even if you were going to rename the destinatoin too

this does not work as expected even though there is no conflict:
```
a -> b
b -> c
```

this makes doing something like incrementing numbers in files in bulk harder to do
```
01 -> 02 ×
02 -> 03 ×
03 -> 04 ×
04 -> 05 ×
05 -> 06 ×
06 -> 07 ×
07 -> 08 ×
08 -> 09 ×
09 -> 10 ✓
```

use temp/intermediate files if destination exists and is scheduled to be renamed:
```
a -> b.XXXXX
b -> c
b.XXXXX -> b
```

or reshuffle rename execution:
```
b -> c
a -> b
```

(i recommend using temp files since it also takes care of swapping files)

### Additional context

_No response_

### Validations

yw1ee:
Resolve #1766 . Sort bulk-renamed files in a similar way to topological sort.

### DIFF
--- a/yazi-core/src/manager/commands/bulk_rename.rs
+++ b/yazi-core/src/manager/commands/bulk_rename.rs
@@ -142,9 +142,15 @@ impl Manager {
 			});
 
 			if has_no_incomes.is_empty() {
-				todo!("Consider cycle")
+				// Remaining rename set has cycle, so we cannot sort, just return them all
+				let mut remain = todos.drain().collect::<Vec<_>>();
+				remain.sort();
+				sorted.reverse();
+				sorted.extend(remain);
+				return sorted;
 			}
 
+			has_no_incomes.sort();
 			for old in has_no_incomes {
 				income_map.remove(&old);
 				let Some(new) = todos.remove(&old) else { unreachable!("") };
@@ -189,8 +195,14 @@ mod tests {
 
 		#[rustfmt::skip]
 		cmp(
-			&[("1", "2"), ("2", "1")],
-			&[("2", "1"), ("1", "2")]
+			&[("2", "1"), ("1", "2")],
+			&[("1", "2"), ("2", "1")]
+		);
+
+		#[rustfmt::skip]
+		cmp(
+			&[("3", "2"), ("2", "1"), ("1", "3"), ("a", "b"), ("b", "c")],
+			&[("b", "c"), ("a", "b"), ("1", "3"), ("2", "1"), ("3", "2")]
 		);
 	}
 }
--- a/yazi-core/src/manager/commands/bulk_rename.rs
+++ b/yazi-core/src/manager/commands/bulk_rename.rs
@@ -1,12 +1,24 @@
-use std::{borrow::Cow, collections::HashMap, ffi::{OsStr, OsString}, io::{BufWriter, Write, stderr}, path::PathBuf};
+use std::{
+	borrow::Cow,
+	collections::HashMap,
+	ffi::{OsStr, OsString},
+	io::{stderr, BufWriter, Write},
+	path::PathBuf,
+};
 
-use anyhow::{Result, anyhow};
+use anyhow::{anyhow, Result};
 use scopeguard::defer;
-use tokio::{fs::{self, OpenOptions}, io::{AsyncReadExt, AsyncWriteExt, stdin}};
+use tokio::{
+	fs::{self, OpenOptions},
+	io::{stdin, AsyncReadExt, AsyncWriteExt},
+};
 use yazi_config::{OPEN, PREVIEW};
 use yazi_dds::Pubsub;
-use yazi_proxy::{AppProxy, HIDER, TasksProxy, WATCHER};
-use yazi_shared::{fs::{File, FilesOp, Url, max_common_root, maybe_exists, paths_to_same_file}, terminal_clear};
+use yazi_proxy::{AppProxy, TasksProxy, HIDER, WATCHER};
+use yazi_shared::{
+	fs::{max_common_root, maybe_exists, paths_to_same_file, File, FilesOp, Url},
+	terminal_clear,
+};
 
 use crate::manager::Manager;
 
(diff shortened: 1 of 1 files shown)


sxyazi:
Thanks for the PR! Could you please add tests to the `sort()` function?

yw1ee:
Sure! I add a very simple unittest for sort.

sxyazi:
Thanks for the change! I added more tests and found that the last two tests are failing. Could you please fix them? 

Specifically, the second test is producing unstable sorting results. It could be:

```
&[("3", "4"), ("2", "3"), ("1", "3")]
```

or 

```
&[("3", "4"), ("1", "3"), ("2", "3")]
```

This might be related to the `HashMap` being used, which has an unpredictable order. Switching it to a `BTreeMap` or `IndexMap` might solve the issue.

As for the last test, it hits `todo!("Consider cycle")`. Normally, `a->b` and `b->a` shouldn't happen, but since the data comes from user input, it's an uncontrollable facto, so we need to handle this case to avoid crashing the program here.

yw1ee:
I also had some concerns while writing the code, so thank you for creating tests for that. I have two questions.
- input validation
  - I think the input of second test case is invalid. 1 -> 3 and 2 -> 3 cannot do in one bulk rename. In the sort function, is it enough to just return the order without validation? I think it will be catched at `maybe_exist` and not be renamed, so is it ok?
- how to handle error
  - I think it is sub-problem of first question. As you mentioned, we can handle the error when cyclic rename detected. But I'm not sure about how to handle this invalidate input and what is proper behavior of `sort` function.

So, could I ask you, the author of this project, to clearly define the roles and responsibilities of the sort function? If it doesn't really matter, I'll work on it at my own discretion.

sxyazi:
I haven't fully understood the implementation of `sort`, but I *think* we should try to keep the user's input order and only prioritize certain items on top of that, like:

```
a -> b
b -> b
b -> c
```

In this case, we have `a -> b` and `b -> b` appearing twice, which means the data itself is invalid. 

However, we don't need to take responsibility for invalid data; that's the user's responsibility, and the program should only handle valid data, as we already know that `b -> c` can be prioritized, so the sorted result should be:

```
b -> c
a -> b
b -> b
```

Invalid data remain unchanged and are passed to the upper layer for handling, which in this case is `maybe_exist`.

yw1ee:
So do you mean that we will rename a valid subset of the user's input even if the user's full input set is invalid (duplicate target name or cyclic rename)? For that purpose, the role of sort function is just prioritize some rename that could be executed without any concerns, right? Please let me know if I have misunderstood anything.

sxyazi:
Yes, the sorting function should only elevate those valid renaming items that have no side effects and can be executed early to the front of the execution queue, assisting the user as much as possible (rather than taking over the user's work and responsibilities) in doing all valid operations, while leaving invalid operations for higher-level resolution — here, that means throwing errors to the user for manual intervention.

yw1ee:
One thing I would like to suggest is that the sort function returns a separate rename set containing a cycle. And in the above function, if there is a cycle, how about showing the list to the user on stderr to indicate that the cycle exists and just return (or user interaction to do only valid things or not?).
```rust
let (valid, cycle) = Self::sort(old, new);
if !cycle.is_empty() {
    writeln!(stderr, "cyclic rename detected");
    for (old, new) in cycle {
        wrintln!(stderr, "{old} -> {new}");
    }
    return Ok(()); // or user interaction to do or not
}
```

sxyazi:
Sorry - no. I don't really see the practical significance of "cycle detection" for users, as it's more of an internal implementation detail. The program just needs to ensure that it safely handles the valid data provided by the user, while leaving the invalid data for the user to handle manually. Introducing this detection seems to discard the entire bulk rename operation, rather than just the invalid parts

yw1ee:
I was mistaken. I thought that if there is a cycle, the rename will be performed and all the files will be messed up, but it will not be performed because it is caught in `maybe_exist`.

I make result consistent and just return cyclic rename entries instead of panic. Thanks for your comment.

sxyazi:
Thanks! I tested the latest changes, and it seems the ordering issue still exists. For example, this test will fail:

```rust
#[rustfmt::skip]
cmp(
    &[("b", "b_"), ("a", "a_"), ("c", "c_")],
    &[("b", "b_"), ("a", "a_"), ("c", "c_")],
);
```

yw1ee:
I miss reserving user input order. I fix it!

sxyazi:
Hey @yw1ee! 

I did some refactoring to simplify the code, could you please do a final review? If everything looks good I'll merge it 🚀

yazi-core/src/manager/commands/bulk_rename.rs
```diff
@@ -117,4 +117,90 @@ impl Manager {
...
+				incomes.get_mut(&n).map(|b| *b = true);
+				(o, n)
+			})
+			.collect();
+
+		let mut sorted = Vec::with_capacity(old.len());
+		while !todos.is_empty() {
+			// Paths that are non-incomes and don't need to be prioritized in this round
+			let mut outcomes: Vec<_> = incomes.iter().filter(|(_, &b)| !b).map(|(&p, _)| p).collect();
+			outcomes.sort_unstable_by(|a, b| orders[b].cmp(&orders[a]));
```
yw1ee, line 137: Is there any reason to use `sort_unstable_by` instead of `sort_by`? for in-place?

yw1ee (commented):
I have just one simple question. LGTM 👍

yazi-core/src/manager/commands/bulk_rename.rs
sxyazi: Yeah, stable needs n/2 auxiliary memory to sort n elements in worst case

sxyazi:
Merging now. 

Thank you so much for contributing this awesome feature @yw1ee, really appreciate it! 🎉

sxyazi (approved):

PR #1837, opened 2024-10-26
refactor: add tests for `Pattern::match_path`

sxyazi:


boydaihungst:
This PR breaks some previewer and preloader, fetcher rules. Maybe the theme > [filetype] as well.
Before we can use `{ name = "/mnt/*", run = "noop" }` to disable preview and preloader for all files under `/mnt` folder, no matter how deep it's. 
But for now we can only use `{ name = "/mnt/**/*", run = "noop" }`. 
So maybe you should add a breaking change when v0.4 is released.

sxyazi:
I'm surprised this would break some existing rules - if you want to match directories deeper than one level, you should always use `**`, and the doc has stated this: https://yazi-rs.github.io/docs/configuration/theme#icon

This PR fixes a bug where `*` and `**` were behaving the same way — `*` should only be used for single-level depth, while `**` should be used for multi-level depth (as in your case).

boydaihungst:
That's true. Thanks for explaining. 😉

PR #1841, opened 2024-10-26
Add Command.which(cmd) function to lua interface

shurizzle:
I added the which function to the global Command table, which simply resolves the path for the command passed as an argument and returns nil if it can't find it.

An example of usage is:
```lua
local bin = Command.which('yazi')
if not bin then
  error('yazi not found')
end
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3398,6 +3398,7 @@ dependencies = [
  "tracing",
  "unicode-width 0.2.0",
  "uzers",
+ "which",
  "yazi-adapter",
  "yazi-boot",
  "yazi-config",
--- a/yazi-plugin/Cargo.toml
+++ b/yazi-plugin/Cargo.toml
@@ -39,6 +39,7 @@ tokio-stream  = { workspace = true }
 tokio-util    = { workspace = true }
 tracing       = { workspace = true }
 unicode-width = { workspace = true }
+which = "6.0.3"
 yazi-prebuild = "0.1.2"
 
 [target."cfg(unix)".dependencies]
--- a/yazi-plugin/src/process/command.rs
+++ b/yazi-plugin/src/process/command.rs
@@ -23,12 +23,17 @@ impl Command {
 			Ok(Self { inner })
 		})?;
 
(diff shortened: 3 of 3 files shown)


sxyazi:
Thanks for the PR! I have a few questions: 

- Are you sure the behavior of the `which` crate is consistent with `Command`? If they aren't consistent, there could be situations where `which` finds an executable but `Command` can't run it, or vice versa. 
- What was your X problem — why do you need to check if the executable exists first? Can't that be determined at runtime? `Command` returns a status code if the execution fails, indicating the executable doesn't exist. 
- Yazi's `Command` is meant to mirror [Rust's `Command`](https://doc.rust-lang.org/std/process/struct.Command.html) as closely as possible, so I don't think it should be added to `Command`, maybe to `ya` instead, but we already have `ya.which`. Do you have any better name suggestions?

shurizzle:
First of all, thank you for expressing your reasonable concerns; you give me the impression that you want to keep the project healthy and well-organized.

- `Command` uses `execvp` or `posix_spawn` directly, so theoretically, they could be inconsistent with each other.
- I simply wanted to create a fallback system using `which` since it has a simpler interface for that purpose.
- I thought about it and didn’t include it in `ya` precisely because `ya.which` already exists, so I opted for `Command.which`. A second option could be `ya.exepath`.

sxyazi:
Thanks for the explanation. Since it doesn't necessarily align with the `Command` behavior and the `which` crate just looks for executables in the file system, let's move it to `fs`: `fs.executable() -> (Url, err)`. 

If found, return the file URL and set err to nil; otherwise, return URL as nil and provide the corresponding err.

shurizzle:
One thing isn't clear to me: what error should I return if I simply can't find the executable?

sxyazi:
You can try creating a new error structure mapping and representing the `Error` from the `which` crate (https://docs.rs/which/latest/which/enum.Error.html). 

If you're not sure how to do it, just leave a `// TODO: xxx` :)

sxyazi:
I'm going to close this because it still needs some work and hasn't seen any activity recently. @shurizzle  thanks for the effort!

PR #1874, opened 2024-10-31
refactor: destructure tuples to enhance readability

Integral-Tech:


yazi-adapter/src/dimension.rs
```diff
@@ -12,9 +12,9 @@ impl Dimension {
 		}
 
 		if size.rows == 0 || size.columns == 0 {
-			if let Ok(s) = crossterm::terminal::size() {
-				size.columns = s.0;
-				size.rows = s.1;
+			if let Ok((columns, rows)) = crossterm::terminal::size() {
```
sxyazi, line 15:
```suggestion
			if let Ok((cols, rows)) = crossterm::terminal::size() {
```

yazi-adapter/src/image.rs
```diff
@@ -73,13 +73,13 @@ impl Image {
 			.unwrap_or((PREVIEW.max_width, PREVIEW.max_height))
 	}
 
-	pub(super) fn pixel_area(size: (u32, u32), rect: Rect) -> Rect {
+	pub(super) fn pixel_area((width, height): (u32, u32), rect: Rect) -> Rect {
```
sxyazi, line 76: Please keep the original code here as `size` is a field term in the `yazi-adapter` scope

yazi-adapter/src/kgp.rs
```diff
@@ -343,7 +343,7 @@ impl Kgp {
 	}
 
 	async fn encode(img: DynamicImage) -> Result<Vec<u8>> {
-		fn output(raw: &[u8], format: u8, size: (u32, u32)) -> Result<Vec<u8>> {
+		fn output(raw: &[u8], format: u8, (width, height): (u32, u32)) -> Result<Vec<u8>> {
```
sxyazi, line 346: Also here

yazi-config/src/manager/mouse.rs
```diff
@@ -45,7 +45,7 @@ impl From<MouseEvents> for Vec<String> {
 			(MouseEvents::MOVE, "move"),
 			(MouseEvents::DRAG, "drag"),
 		];
-		events.into_iter().filter(|v| value.contains(v.0)).map(|v| v.1.to_owned()).collect()
+		events.into_iter().filter(|(e, _)| value.contains(*e)).map(|(_, e)| e.to_owned()).collect()
```
sxyazi, line 48: I don't like this style because it has two `e`s at the same time, which makes it less recognizable compared to the previous `v.0` and `v.1`.

yazi-fm/src/signals.rs
```diff
@@ -92,19 +92,19 @@ impl Signals {
 				if let Some(t) = &mut term {
 					select! {
 						biased;
-						Some(mut s) = rx.recv() => {
-							term = term.filter(|_| s.0);
-							s.1.take().map(|cb| cb.send(()));
+						Some((s, mut cb)) = rx.recv() => {
+							term = term.filter(|_| s);
+							cb.take().map(|cb| cb.send(()));
```
sxyazi, line 97: I used `s` here because if we use `cb`, it will clash with the `cb` arg in `map(|cb|`, which lowers the recognizability

yazi-plugin/src/bindings/position.rs
```diff
@@ -11,7 +11,7 @@ impl Deref for Position {
 }
 
 impl From<Position> for yazi_config::popup::Position {
-	fn from(value: Position) -> Self { value.0 }
+	fn from(Position(value): Position) -> Self { value }
```
sxyazi, line 14: I rarely see Rust traits written like this, and it repeats the type twice.

yazi-plugin/src/elements/line.rs
```diff
@@ -54,11 +54,11 @@ impl TryFrom<Value<'_>> for Line {
 			Value::Table(tb) => return Self::try_from(tb),
 			Value::String(s) => s.to_string_lossy().into_owned().into(),
 			Value::UserData(ud) => {
-				if let Ok(span) = ud.take::<Span>() {
-					span.0.into()
-				} else if let Ok(mut line) = ud.take::<Line>() {
-					line.0.spans.iter_mut().for_each(|s| s.style = line.0.style.patch(s.style));
-					line.0
+				if let Ok(Span(span)) = ud.take::<Span>() {
```
sxyazi, line 57: Also this, it repeats the type twice, I think this is unnecessary

sxyazi: Maybe we can remove `::<Span>` but rather keep `Ok(Span(` in this case

yazi-config/src/manager/mouse.rs
Integral-Tech: How about this one?
```rs 
events.into_iter().filter(|(a, _)| value.contains(*a)).map(|(_, b)| b.to_owned()).collect()
```

sxyazi: IMO this doesn't make it clearer, since when I see `a` and `b`, I still have to think about what corresponds to `v.0` and `v.1`, so why not just use `v.0` and `v.1` directly?

Integral-Tech:
@sxyazi All suggestions fixed :)

yazi-fm/src/signals.rs
```diff
@@ -92,19 +92,19 @@ impl Signals {
 				if let Some(t) = &mut term {
 					select! {
 						biased;
-						Some(mut s) = rx.recv() => {
-							term = term.filter(|_| s.0);
-							s.1.take().map(|cb| cb.send(()));
+						Some((status, mut callback)) = rx.recv() => {
```
sxyazi, line 95: nit

```suggestion
						Some((state, mut callback)) = rx.recv() => {
```

yazi-plugin/src/elements/gauge.rs
```diff
@@ -43,8 +43,8 @@ impl UserData for Gauge {
 			Ok(ud)
 		});
 
-		methods.add_function_mut("label", |_, (ud, label): (AnyUserData, Span)| {
-			ud.borrow_mut::<Self>()?.label = Some(label.0);
+		methods.add_function_mut("label", |_, (ud, Span(label)): (AnyUserData, Span)| {
```
sxyazi, line 46: Please keep the original code here since I want to avoid the `Span` twice

sxyazi (changes requested):
Thanks, it looks great overall! 

I have two small change requests:

Integral-Tech:
> Thanks, it looks great overall!
> 
> I have two small change requests:

Change requests fixed :)

sxyazi (approved):
LGTM! These refactors are stunning I can't lie :D

PR #1879, opened 2024-11-02
Use XDG spec for goto navigation.

### LINKED ISSUE #1878: Use XDG paths for Goto.

Current Goto definitions are as follows: `{ on = [ "g", "d" ],       run = "cd ~/Downloads",   desc = "Goto ~/Downloads" },`. Which is.. something, I guess. Why not use much more portable approach of `cd $XDG_DOWNLOAD_DIR`? Same goes for `~/.config`.

Use XDG variables.

### Additional context

I tested them, they work as expected.

### Validations

- [X] The [latest nightly build](https://yazi-rs.github.io/docs/installation/#official-binaries) doesn't already have this feature

voronind-com:
Closes #1878
More info about XDG [here](https://specifications.freedesktop.org/basedir-spec/latest/).

### DIFF
--- a/yazi-config/preset/keymap.toml
+++ b/yazi-config/preset/keymap.toml
@@ -121,10 +121,10 @@ keymap = [
 	{ on = [ ",", "r" ], run = "sort random --reverse=no",                      desc = "Sort randomly" },
 
 	# Goto
-	{ on = [ "g", "h" ],       run = "cd ~",             desc = "Go home" },
-	{ on = [ "g", "c" ],       run = "cd ~/.config",     desc = "Goto ~/.config" },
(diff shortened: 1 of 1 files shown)


sxyazi:
Sorry, I'm afraid I can't accept the PR, because this breaks on macOS and Windows, they don't have an `XDG_CONFIG_HOME` and `XDG_DOWNLOAD_DIR` at all, but they generally have the `~/.config` and `~/Downloads` folder

sxyazi:
Don't get me wrong - I can accept platform-specific features as long as they don't cause trouble for users on other platforms, and there needs to be a better way to implement it than this PR, maybe as a built-in plugin

PR #1887, opened 2024-11-04
refactor: use for loop to print debug info

Integral-Tech:


yazi-boot/src/actions/debug.rs
```diff
@@ -79,17 +82,22 @@ impl Actions {
...
+			("fzf", "--version"),
+			("fd", "--version"),
+			("rg", "--version"),
+			("chafa", "--version"),
+			("zoxide", "--version"),
+			("7z", "i"),
+			("7zz", "i"),
+			("jq", "--version"),
+		] {
+			writeln!(s, "    {:17}: {}", cmd, Self::process_output(cmd, arg))?;
```
sxyazi, line 99: This doesn't seem to make maintenance easier, since now I have to manually figure out which one is the longest and calculate its length. 

The previous code was WYSIWYG — if it looked aligned in the editor, the output would be aligned too, and I could freely fill in the missing spaces. 

What's worse is that for things like `file`, which need special logic, I still have to handle them the old way — manually filling in the spaces. This creates two different styles.

sxyazi (changes requested):

yazi-boot/src/actions/debug.rs
```diff
@@ -79,17 +89,26 @@ impl Actions {
 			"    file             : {}",
```
sxyazi, line 89: How can I align this `file` if there's a need to add in more deps?

Integral-Tech: @sxyazi Solved :)

PR #1891, opened 2024-11-07
feat!: deprecate `--sync` option for the `plugin` command

sxyazi:


dawsers:
How does this work for plugins that have entry points that are sync and others that are async?

In the past, you could call certain functions of a plugin with `--sync`, leaving the others async. It seems now you need to choose whether the plugin works completely in synchronous mode or not?

This could be a problem for plugins that work mostly in synchronous mode but may have certain functionality that needs to call a shell or some function that needs to work in asynchronous mode.

For example:

```
    { on = "<Tab>", run = "plugin --sync dual-pane --args=next_pane",  desc = "Dual-pane: switch to the other pane" },
    { on = [ "b", "o" ], run = "plugin dual-pane --args='shell --interactive --block \"%c\"'", desc = "Dual-pane: open with command, blocking" },
    { on = [ "b", "O" ], run = "plugin dual-pane --args='shell --interactive \"%c\"'", desc = "Dual-pane: open with command, background" },
```

sxyazi:
Ah, that's really unfortunate, I missed considering both sync and async plugins...

Is it possible to turn your plugin into a fully async one? In an async plugin, you can run any sync code using `ya.sync()`, like this:

```lua
local sync_job1 = ya.sync(function()
  -- do some sync job
end)

local sync_job2 = ya.sync(function()
  -- do some sync job
end)

local sync_job3 = ya.sync(function()
  -- do some sync job
end)

local entry = function(_, job)
  if job.args[1] == "sync_job1" then
    sync_job1()
  elseif job.args[1] == "sync_job2" then
    sync_job2()
  elseif job.args[1] == "sync_job3" then
    sync_job3()
  else
    -- do other async job
  end
end

return { entry }
```

I was thinking that this limitation will be lifted in the future with the support for multi-file plugins. 

At that point, each file will be able to have its own execution environment (`@sync` tag), and users will be able to start different Lua entry files using `.` like this:

```sh
plugin my-plugin       # ~/.config/yazi/plugins/my-plugin.yazi/init.lua
plugin my-plugin.sync  # ~/.config/yazi/plugins/my-plugin.yazi/sync.lua
```

And the `sync.lua` file have the `@sync` tag:

```lua
--- @sync entry
return { entry = function() end }
```

But this feature hasn't been implemented yet, so for now, the only way to solve this is through async plugins - sorry about that!

dawsers:
No problem, I just wanted to make sure I wasn't missing something. Thanks!

Issue #1141, opened 2024-06-08
Add alignment settings for preview


gaesa:
### Please describe the problem you're trying to solve

Currently, the `yazi.toml` configuration file does not provide options to control the alignment of images within the preview section. This limitation restricts users from customizing how images are positioned within the designated preview area, leading to less flexibility in layout management. Plugin developers may often need to manually calculate and adjust image positions, which can be time-consuming and error-prone. This is a common demand in many applications.

### Would you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

Add a new section `[preview.alignment]` within the existing `[preview]` settings in `yazi.toml`. This section would include two options: `horizontal` and `vertical`, allowing users to specify the alignment of images. 

The horizontal option should support values `"left"`, `"center"`, and `"right"`, while the vertical option should support values `"top"`, `"center"`, and `"bottom"`. There are 9 combinations in total, all are valid. By default, `horizontal` should be set to `"left"` and `vertical` should be set to `"top"` to maintain backward compatibility.

The new alignment settings will serve as the default preview rule. Plugins can still use the existing `[plugin.prepend_previewers]` configuration to override the default alignment behavior.

### Additional context

Introducing these alignment settings would significantly enhance user experience by providing greater control over image placement, making their workflow more convenient and efficient, resulting in a more versatile and user-friendly preview experience. The feature would support nine valid alignment combinations, ensuring comprehensive options for users to choose from. The most commonly used combinations are:

1. `horizontal = "left", vertical = "top"`
2. `horizontal = "center", vertical = "top"`
3. `horizontal = "center", vertical = "center"`

Additionally, plugin developers will benefit from this feature by avoiding repetitive calculations for image positioning. This built-in functionality reduces the complexity of plugin development and maintenance, allowing developers to focus on other critical features.

sxyazi:
Hi there.

I don't think there's a simple way to achieve this because Yazi supports a mixed mode preview of text and images (like what the https://github.com/Sonico98/exifaudio.yazi plugin does). In this case, we can't predict the relative position of the image and text — whether the image is above the text, the text is above the image, or how much space the text takes up to leave room for the image. This information is only known by the plugin itself.

PS: Currently, only community plugins support mixed preview, but I plan to add these text details (such as image dimensions, video length, etc.) to Yazi's built-in previewer in the future.

sxyazi:
Closing with https://github.com/sxyazi/yazi/pull/1897

PR #1897, opened 2024-11-10
feat: support customizable alignment for image display

gaesa:
This is an attempt to provide a possible solution to #1141.

- Adds configuration options in `yazi.toml` for setting preferred image alignment, which plugin developers can optionally respect.
- Updates plugin defaults to use `PREVIEW.alignment` for aligned image display.
- Ensures backward compatibility for existing plugins.

It appears to work well overall, though some aspects remain unclear:
1. Does the current use of `serde` align with usual practices?
2. Are the new Lua API function consistent with current design philosophy?

### DIFF
--- a/yazi-adapter/src/adapter.rs
+++ b/yazi-adapter/src/adapter.rs
@@ -5,7 +5,7 @@ use ratatui::layout::Rect;
 use tracing::warn;
 use yazi_shared::env_exists;
 
-use crate::{Brand, Emulator, SHOWN, TMUX, WSL, drivers};
+use crate::{Brand, Emulator, Offset, SHOWN, TMUX, WSL, drivers};
 
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
 pub enum Adapter {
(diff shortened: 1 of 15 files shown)


sxyazi:
Thanks for the PR! 

Could you please just change the key parts and pull out the refactoring into a new PR? That would make the review a bit easier.

sxyazi:
I've decided not to build it in, or at least not put it under `[preview]`, because that implies it applies to all previewers while it usually only makes sense for image previewers and this PR also only adds support for image alignment. 

Changing alignment for text previewers (like [`code`](https://github.com/sxyazi/yazi/blob/main/yazi-plugin/preset/plugins/code.lua), [`archive`](https://github.com/sxyazi/yazi/blob/main/yazi-plugin/preset/plugins/archive.lua), [`folder`](https://github.com/sxyazi/yazi/blob/main/yazi-plugin/preset/plugins/folder.lua)) would feel kinda odd, and it's still unclear how it should behave for previews that mix text and images together. Instead, I think this should be handled at the previewer level and be configurable per-previewer - since that PR, Yazi has added support for [passing params to previewers](https://github.com/sxyazi/yazi/pull/1966). 

PR #1900, opened 2024-11-11
refactor(adapter): extract `image_area` from `image_show`

gaesa:
Exposing `image_area` separately allows future Lua integration to handle image placement more flexibly, without needing to invoke full rendering. This change promotes modularity and simplifies obtaining image dimensions alone.

This refactor provides the foundation for feature PR #1897, preparing for easier integration.

### DIFF
--- a/yazi-adapter/src/adapter.rs
+++ b/yazi-adapter/src/adapter.rs
@@ -5,7 +5,7 @@ use ratatui::layout::Rect;
 use tracing::warn;
 use yazi_shared::env_exists;
 
-use super::{Iip, Kgp, KgpOld};
+use super::{Iip, Image, Kgp, KgpOld};
 use crate::{Chafa, Emulator, SHOWN, Sixel, TMUX, Ueberzug, WSL};
 
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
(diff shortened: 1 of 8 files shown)


sxyazi:
Why not implement the feature directly in `image_show()` so that it supports alignment, instead of requiring an extra step to get `image_size()` first? 

To me, this is a significant performance loss, as now two I/O reads are necessary. Plus, for large images that haven't been downscaled, an additional `downscale()` operation is needed.

sxyazi:
> If modifying the `image_show` function signature is acceptable, I could consider submitting a new PR with these changes. I’m definitely open to other potential approaches if they better suit the codebase

Sure, I can accept changing the function signature as long as it benefits performance.

> As long as the original `image_show` performs downscaling, any `image_area` extracted from it will already have this downscaling applied.

If I understand correctly, from another PR, [`image_show_aligned()` API](https://github.com/sxyazi/yazi/pull/1897/files#diff-60898fc6da94b85ebcc1f941945449f4de080f41f0a8b8260247bbeb929af052) calls both `image_area()` and `image_show()`, and both of these functions ultimately call `downscale()`. The issues here are: 

1. The image gets fully decoded twice, even if it doesn't need to be downscaled
  https://github.com/sxyazi/yazi/blob/c477f68058ca3d546aca0c1533cc582e270cc3ef/yazi-adapter/src/image.rs#L44-L45
2. Even when `image_area()` has already downscaled the image, `image_show()` still uses the original image without utilizing the already downscaled result. This leads to images that need downscaling getting processed twice.

PR #1925, opened 2024-11-20
fix: show "Empty" instead of "0/1" in empty directories

Integral-Tech:


yazi-plugin/preset/components/status.lua
```diff
@@ -124,9 +124,10 @@ function Status:position()
 	local length = #self._current.files
 
 	local style = self:style()
+	local display_text = length > 0 and string.format(" %2d/%-2d ", cursor + 1, length) or " Empty "
 	return ui.Line {
 		ui.Span(THEME.status.separator_open):fg(style.bg):bg(THEME.status.separator_style.fg),
-		ui.Span(string.format(" %2d/%-2d ", cursor + 1, length)):style(style),
+		ui.Span(display_text):style(style),
```
sxyazi, line 130:
```suggestion
		ui.Span(string.format(" %2d/%-2d ", math.min(cursor + 1, length), length)):style(style),
```

sxyazi (changes requested):
Please apply this change request:

Issue #1864, opened 2024-10-30
extracting `tar.gz` archive doesn't preserve permissions


xfzv:
### What system are you running Yazi on?

Linux X11

### What terminal are you running Yazi in?

kitty 0.36.4

### `yazi --debug` output

```Shell
Yazi
    Version: 0.3.3 (c668723 2024-10-30)
    Debug  : false
    OS     : linux-x86_64 (unix)

Ya
    Version: 0.3.3 (c668723 2024-10-30)

Emulator
    Emulator.via_env: ("xterm-kitty", "")
    Emulator.via_csi: Ok(Kitty)
    Emulator.detect : Kitty

Adapter
    Adapter.matches: Kgp

Desktop
    XDG_SESSION_TYPE           : Some("tty")
    WAYLAND_DISPLAY            : None
    DISPLAY                    : Some(":0")
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Variables
    SHELL              : Some("/bin/zsh")
    EDITOR             : Some("nvim")
    VISUAL             : Some("nvim")
    YAZI_FILE_ONE      : None
    YAZI_CONFIG_HOME   : None

Text Opener
    default     : Some(Opener { run: "${EDITOR} \"$@\"", block: true, orphan: false, desc: "Neovim", for_: None, spread: true })
    block-create: Some(Opener { run: "${EDITOR} \"$@\"", block: true, orphan: false, desc: "Neovim", for_: None, spread: true })
    block-rename: Some(Opener { run: "${EDITOR} \"$@\"", block: true, orphan: false, desc: "Neovim", for_: None, spread: true })

Multiplexers
    TMUX               : false
    tmux version       : No such file or directory (os error 2)
    tmux build flags   : enable-sixel=Unknown
    ZELLIJ_SESSION_NAME: None
    Zellij version     : No such file or directory (os error 2)

Dependencies
    file             : 5.45
    ueberzugpp       : No such file or directory (os error 2)
    ffmpegthumbnailer: 2.2.2
    magick           : 7.1.1-38
    fzf              : 0.54.3
    fd               : 10.2.0
    rg               : 14.1.1
    chafa            : No such file or directory (os error 2)
    zoxide           : 0.9.6
    7z               : 17.05
    7zz              : No such file or directory (os error 2)
    jq               : 1.7.1


[...]

PR #1926, opened 2024-11-20
feat: add more supported archive formats to the preset config

Integral-Tech:
References:
- https://documentation.help/7-Zip-18.0/formats.htm
- https://www.digipres.org/formats/mime-types
- https://mimetype.io
- https://github.com/file/file


Integral-Tech:
@sxyazi This PR has been completed :) Would you like to review it?

sxyazi:
Thanks for the PR!

I noticed you referenced the supported formats of `7zip`, but `7zip` has several different versions. On Arch Linux, [Yazi uses `p7zip`](https://yazi-rs.github.io/docs/installation#arch-linux), which is a fork of `7zip`, while on macOS, it [uses `7zip`](https://yazi-rs.github.io/docs/installation#homebrew) (`sevenzip`), I'm not sure about Windows. 

The supported formats often vary between them. For default configuration, we need to consider the formats supported across all platforms, which means focusing on the smallest common subset - supported by both `7zip` and `p7zip` on Linux, macOS and Windows. 

This PR includes many formats I've never seen before, which makes testing them much more difficult. I think we can limit it to common formats that have been verified to work. This would also simplify the rules, especially considering that `7zip` is just a temporary solution during the transition. Eventually, I'll implement it myself (see https://github.com/sxyazi/yazi/issues/1864#issuecomment-2447902303), but it's unlikely to support all the formats `7zip` does, which would also help reduce confusion once we switch to another tool.

Integral-Tech:
@sxyazi I just removed `application/zlib` which is not supported by `p7zip`. 

Additionally, `.dmg` & `.DMG` is added as not all `.dmg` files have MIME type `application/apple-diskimage`. Take case of [Icalingua++](https://github.com/Icalingua-plus-plus/Icalingua-plus-plus/releases/download/v2.12.24/Icalingua++-2.12.24.AppImage), it has `application/zlib` MIME type.

sxyazi (approved):
LGTM, thank you so much!

PR #1935, opened 2024-11-22
refactor: make error messages more user-friendly

Integral-Tech:
- "Spawn a command/process" is not straightforward for non-technical users. Using "Run a command/process" or "Execute a command/process" in error messages would be better.


sxyazi (approved):
Thanks!

PR #1941, opened 2024-11-23
perf: use rayon to parallelize directory operations

Integral-Tech:
- Currently, all directory operations including size calculating, removing and etc, are single-threaded per task. Use `rayon` to parallelize these operations.


sxyazi:
Closing as we discussed last time on Telegram, thanks for your effort @Integral-Tech!

PR #1955, opened 2024-11-26
update: change percent text Bot to End

leyiang:
In the status bar, the percentage text displays "Bot," which initially confused me. After moving the cursor around, I realized it stands for "bottom." It might be clearer to change it to "End," which would make the meaning more obvious to users.

image

### DIFF
--- a/yazi-plugin/preset/components/status.lua
+++ b/yazi-plugin/preset/components/status.lua
@@ -108,7 +108,7 @@ function Status:percent()
 	if percent == 0 then
 		percent = " Top "
 	elseif percent == 100 then
-		percent = " Bot "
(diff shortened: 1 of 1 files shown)


sxyazi:
Sorry, I'm afraid I can't accept this PR. The use of "Bot" (short for bottom) is intended to pair with "Top", as in `bottom / top`

And using "End" sounds a bit strange, like `end / top` — it doesn't quite follow the usual English convention. 

Also, "Bot" is borrowed from Vim/Neovim - they also use "Bot" as the short form of "bottom". Since Yazi is a Vim-like file manager, I don't think this should be an issue.

PR #1956, opened 2024-11-26
update: exit input instaed of change mode for empty input

leyiang:
Sometimes, I accidentally open an input window and want to exit it. I currently have to press Esc twice: once to return to normal mode and again to exit. But if the input field is empty, I believe it should be possible to exit directly.

### DIFF
--- a/yazi-core/src/input/commands/escape.rs
+++ b/yazi-core/src/input/commands/escape.rs
@@ -25,6 +25,11 @@ impl Input {
 				snap.op = InputOp::None;
 			}
 			InputMode::Insert => {
+				if snap.value.trim().is_empty() {
+					self.close(false);
+					return;
+				}
+
 				snap.mode = InputMode::Normal;
 				self.move_(-1);
 


boydaihungst:
Why don't use close?
https://yazi-rs.github.io/docs/configuration/keymap#input.close

leyiang:
I used to use the Esc key to close windows. I tried adding the "close" command to Esc, but it didn’t work, and I still have to press Esc twice. (The C-c keymap works, but I prefer not to use a separate key.) Feel free to close this PR if this isn’t a natural habit for most users.

boydaihungst:
Like you, I am just a regular user and have no authority over this repository. 

Set this in keymap.toml > `[input]` section. 
`{ on = "<Esc>", run = "close", desc = "Go back the cancel input" }`

leyiang:
Yes, you can map the Esc key to close the input, but this will disable its default functionality. 
My changes ensure that the input is closed when empty, without affecting other Esc functionalities.
I'm okay with pressing Esc twice to close the input. I appreciate the Vim-like behavior, but my changes aim to use a single Esc press to close the input when it is empty.

sxyazi:
Sorry, but I'm afraid I can't accept this PR. 

The main issue is that "using an empty character to determine whether to close the input" is a tricky behavior. I've worked on this in the past — for example, when the input is empty, pressing the `Delete` key to close the input seemed reasonable at first, but in practice, it caused confusion for many users, see https://github.com/sxyazi/yazi/pull/1680.

In the case of your PR, it's just swapping `Delete` with the `Escape` key, but the underlying issue is the same. I think a better approach would be to maintain consistent behavior across the board, but allow users to configure certain less common inputs (like the find input) to directly exit to visual mode by once Escape press.

PR #1957, opened 2024-11-27
update: add find --clear option

leyiang:
Add new flag for find command. run `find --clear` will clear the previous search input.
I usually use `find` command to search folder and enter, with  `find --clear` I can have this:

```toml
[[manager.prepend_keymap]]
on   = "l"
run  = ["plugin smart-enter", "find --clear"]
```
open folder and also clear search, then don't need to manually type "/" + "Enter" to clear search.


sxyazi:
`escape --find` is exactly what you're looking for, https://yazi-rs.github.io/docs/configuration/keymap#manager.escape

Closing as already exists

Issue #1913, opened 2024-11-16
`cd -` to jump to previous directory


Alex23rodriguez:
### `yazi --debug` output

```Shell
Yazi
    Version: 0.3.3 (Homebrew 2024-09-04)
    Debug  : false
    OS     : macos-aarch64 (unix)

Ya
    Version: 0.3.3 (Homebrew 2024-09-04)

Emulator
    Emulator.via_env: ("xterm-256color", "WezTerm")
    Emulator.via_csi: Ok(WezTerm)
    Emulator.detect : WezTerm

Adapter
    Adapter.matches: Iterm2

Desktop
    XDG_SESSION_TYPE           : None
    WAYLAND_DISPLAY            : None
    DISPLAY                    : None
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Variables
    SHELL              : Some("/bin/zsh")
    EDITOR             : Some("vim")
    VISUAL             : None
    YAZI_FILE_ONE      : None
    YAZI_CONFIG_HOME   : None

Text Opener
    default: Some(Opener { run: "nvim \"$@\"", block: true, orphan: false, desc: "nvim", for_: None, spread: true })
    block  : Some(Opener { run: "nvim \"$@\"", block: true, orphan: false, desc: "nvim", for_: None, spread: true })

Multiplexers
    TMUX               : false
    tmux version       : No such file or directory (os error 2)
    ZELLIJ_SESSION_NAME: None
    Zellij version     : No such file or directory (os error 2)

Dependencies
    file             : 5.41
    ueberzugpp       : No such file or directory (os error 2)
    ffmpegthumbnailer: No such file or directory (os error 2)
    magick           : No such file or directory (os error 2)
    fzf              : 0.55.0
    fd               : 10.2.0
    rg               : 14.1.0
    chafa            : No such file or directory (os error 2)
    zoxide           : No such file or directory (os error 2)
    7z               : No such file or directory (os error 2)
    7zz              : No such file or directory (os error 2)
    jq               : 1.6


--------------------------------------------------
When reporting a bug, please also upload the `yazi.log` log file - only upload the most recent content by time.
You can find it in the "/Users/alejandro.rodriguez/.local/state/yazi" directory.
```

### Please describe the problem you're trying to solve

`cd -` to jump to previous directory does not work

[...]

PR #1958, opened 2024-11-27
feat: support cd - (cd last)

leyiang:
Support `cd -`, go to last visit dir.
I konw in #1913 you metioned there's `back` and `forward`, but it's logic is not exactly the same with `cd -`, "back" shift from backstack but won't add current cwd into it. 
https://github.com/sxyazi/yazi/blob/af565fbc40b9d443669db2ab46ee491be323c6fd/yazi-config/preset/keymap.toml#L42-L43
`cd -` can toggle between two places, I know we can simulate this with tabs. But I personally want the `cd -` feature, feel free to close or merge this PR.

### DIFF
--- a/yazi-core/src/tab/commands/cd.rs
+++ b/yazi-core/src/tab/commands/cd.rs
@@ -26,7 +26,7 @@ impl From<Cmd> for Opt {
 
 impl From<Url> for Opt {
 	fn from(mut target: Url) -> Self {
-		if target.is_regular() {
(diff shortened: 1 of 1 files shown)


sxyazi (changes requested):
Please create a new dedicated command called `flip` for it, because `cd` is not just a command, but also an important API that will be used in the plugin system. 

Supporting `cd -` would affect cases where the directory name itself is `-` (I know this might not be very common, but I want to keep the behavior of `cd` as consistent as possible).

Issue #1947, opened 2024-11-25
Let `ya` support `.nu` files


zooeywm:
### `yazi --debug` output

```Shell
❯ yazi --debug

Yazi
    Version: 0.3.3 (1277b017 2024-11-24)
    Debug  : false
    Triple : x86_64-unknown-linux-gnu (linux-x86_64)
    Rustc  : 1.82.0 (f6e511ee 2024-10-15)

Ya
    Version: 0.3.3 (1277b017 2024-11-24)

Emulator
    Emulator.via_env: ("xterm-256color", "WezTerm")
    Emulator.via_csi: Ok(WezTerm)
    Emulator.detect : WezTerm

Adapter
    Adapter.matches: Iip

Desktop
    XDG_SESSION_TYPE           : Some("wayland")
    WAYLAND_DISPLAY            : Some("wayland-1")
    DISPLAY                    : Some(":0")
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: Some("12f9a0d0b93f691d4d9923716557154d74777b0a_1732499023_1841097409")
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Variables
    SHELL           : Some("/usr/bin/zsh")
    EDITOR          : Some("lvim")
    VISUAL          : Some("lvim")
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
    FZF_DEFAULT_OPTS: Some("--ansi --height=100% --reverse --cycle")

Text Opener
    default     : Some(Opener { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block-create: Some(Opener { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block-rename: Some(Opener { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })

Multiplexers
    TMUX               : false
    tmux version       : No such file or directory (os error 2)
    tmux build flags   : enable-sixel=Unknown
    ZELLIJ_SESSION_NAME: None
    Zellij version     : 0.41.2

Dependencies
    file          : 5.45
    ueberzugpp    : No such file or directory (os error 2)
[...]

PR #1973, opened 2024-11-30
feat: support `assets` installation for the `ya pack` subcommand

zooeywm:
feature for #1947 , haven tested based on [my sudo.yazi fork](https://github.com/zooeywm/sudo.yazi/tree/main)

```zsh
❯ ./ya pack --add=zooeywm/sudo

  Upgrading package `sudo.yazi`

remote: Enumerating objects: 5, done.
remote: Counting objects: 100% (5/5), done.
remote: Compressing objects: 100% (1/1), done.
remote: Total 3 (delta 2), reused 3 (delta 2), pack-reused 0 (from 0)
Unpacking objects: 100% (3/3), 544 bytes | 544.00 KiB/s, done.
From https://github.com/zooeywm/sudo.yazi
   4f446d3..8a3c803  main       -> origin/main
Previous HEAD position was 4f446d3 chore(ya): `ya pack` -a support assets folder contains miscellaneous files
HEAD is now at 8a3c803 chore(yazi): https://github.com/sxyazi/yazi/pull/1966

❯ tree ~/.config/yazi/plugins/sudo.yazi
/home/zooeywm/.config/yazi/plugins/sudo.yazi
├── assets
│   └── fs.nu
├── DO_NOT_MODIFY_ANYTHING_IN_THIS_DIRECTORY
├── init.lua
├── LICENSE
└── README.md
```

### DIFF
--- a/yazi-cli/src/package/deploy.rs
+++ b/yazi-cli/src/package/deploy.rs
@@ -1,34 +1,46 @@
 use anyhow::{Context, Result, bail};
-use tokio::fs;
-use yazi_shared::{Xdg, fs::{copy_dir_all, maybe_exists, must_exists}};
+use tokio::fs::{self, remove_dir_all};
+use yazi_shared::{Xdg, fs::{copy_dir_all, maybe_exists, must_exists, remove_dir_clean}};
 
-use super::Package;
+use super::{DiffAction, Package};
+use crate::package::Git;
 
 const TRACKER: &str = "DO_NOT_MODIFY_ANYTHING_IN_THIS_DIRECTORY";
 
 impl Package {
 	pub(super) async fn deploy(&mut self) -> Result<()> {
 		let Some(name) = self.name().map(ToOwned::to_owned) else { bail!("Invalid package url") };
-		let from = self.local().join(&self.child);
+		let origin = self.local().join(&self.child);
 
 		self.header("Deploying package `{name}`")?;
-		self.is_flavor = maybe_exists(&from.join("flavor.toml")).await;
-		let to = if self.is_flavor {
+		self.is_flavor = maybe_exists(&origin.join("flavor.toml")).await;
+		let dest = if self.is_flavor {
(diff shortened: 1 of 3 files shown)


boydaihungst:
Is there an option to use a different folder name instead of `assets`? For example: `ya pack -a --assets=folder_x`.
Some plugin repositories already have an assets folder, which is used for image previews.

sxyazi:
@boydaihungst Specifying the assets directory name when adding a plugin means you'll still need to specify it when uninstalling, otherwise Yazi won't know if the user has used a custom assets name, which can mess with automation a bit.

Also, this doesn't seem to affect existing plugins - it's just that directories that weren't copied before will now be copied during installation. 

On another note, it's not recommended to put preview images alongside the code because they're large binaries, which can slow down the cloning process when installing plugins.

The recommended approach is to drag the image into the markdown file to upload it, and GitHub will automatically create a permanent link for it, like this https://github.com/yazi-rs/plugins/blob/a53d5440481f0f9a2160ded47d343bd22ffbc1fb/chmod.yazi/README.md?plain=1#L5 (not an image but a video though).

boydaihungst:
@sxyazi I see. How about a file like `.gitignore` for yazi to ignore thoe files?

sxyazi:
Thanks for the PR @zooeywm! 

I tested it and found that there seems to be an issue with file status synchronization, where files in the assets folder are never deleted, even though the remote plugin's assets directory no longer contains those files.

zooeywm:
Hi @sxyazi , I've just tried to solve the synchronize between config files status and local state files in commit 5ef2e81.

By write rev to the TRACK file, check if the recorded rev is the same as new rev, if not, the `DiffAction` with be make with `Git` module and processed by `deploy()`. I have also use the `remove_dir_clean()` to remove empty folders.

Thanks for review and hope that works!

sxyazi (approved):
Thanks!

Issue #1272, opened 2024-07-08
Show mount points


Ape:
### Please describe the problem you're trying to solve

I'd like to see if a directory is a mount point.

### Would you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

It could look something like this:

```
bin
boot (vfat on /dev/sdb1)
dev (devtmpfs)
etc
home (btrfs on /dev/sdc1)
lib
lib64
mnt
opt
proc (proc)
root
run (tmpfs)
sbin
srv
sys (sysfs)
tmp (tmpfs)
usr
var
```

sxyazi:
Does this mean it only works on Unix systems? I'm not sure if Windows has the concept of mount points.

I think it should be implemented as a plugin rather than a builtin feature. Recently, I've been enhancing the API for extending UI components with plugins, https://github.com/sxyazi/yazi/pull/1257. This will make it easier to implement it as a UI plugin & fetcher.

Ape:
Yeah, I guess this would be Unix only. I'd be happy to have the feature as a plugin.

Ape:
I made a proof-of-concept plugin for this feature: https://github.com/Ape/mount-points.yazi

However, it is currently quite unusable because of performance implications. The plugin refreshes the mount point information constantly. Ideally it would do it in a task and only when needed.

AnirudhG07:
@Ape there's a plugin [archivemount.yazi](https://github.com/AnirudhG07/archivemount.yazi), it can mount archives(as the name suggests) as well tell if a direc is a mountpoint or not.
Check if out. If you do not find it satisfactory, we can update the UI part for that plugin, as per requirements suggest. Thanks!

Ape:
> @Ape there's a plugin [archivemount.yazi](https://github.com/AnirudhG07/archivemount.yazi), it can mount archives(as the name suggests) as well tell if a direc is a mountpoint or not. Check if out. If you do not find it satisfactory, we can update the UI part for that plugin, as per requirements suggest. Thanks!

That plugin doesn't seem to currently have the feature I'm looking for. Check `mount-points.yazi` for reference.

ovwxxwvo:
@Ape @sxyazi hey, mate ! how about this. I thing this will be a little bit better.
![20241118_120238](https://github.com/user-attachments/assets/af80aacf-2542-46f2-a0bb-feb545c5b339)
```
function Status:name()
	local h = self._tab.current.hovered
[...]

Issue #927, opened 2024-04-18
built in mount menu


4rtemis-4rrow:
### Please describe the problem you're trying to solve

a way to easily manage drives

### Would you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

simply add a menu used to manage different drives/partitions (mount, unmount, CD to), would also be preferable if it showed all mounted partitions in an easily accessible path, though that depends on if yazi was designed from the start to be obscured from the file system or not

sxyazi:
Hi, it's already included in the [Feature Request](https://github.com/sxyazi/yazi/issues/51#issuecomment-1757470132), but I want to implement it as a plugin, just like https://github.com/SL-RU/ranger_udisk_menu, instead of embedding it, because there are cross-platform issues - it only works for Unix users.

Ideally, built-in commands should satisfy all platforms, and implementing it as a plugin will provide greater flexibility, allowing users to configure them in Lua. If you or someone interested in implementing it, please let me know, and we can further explore it, closing as duplicate.

4rtemis-4rrow:
I mean, I can design great ncurses based UIs, and I could write the entire mounter myself, issue is, idk lua, so there's that

if we could write it in python, and then the init.lua just runs that, then yes, I could do it

sxyazi:
Yup this is possible, by implementing it through Python and then using [`shell --block` command](https://yazi-rs.github.io/docs/configuration/keymap/#manager.shell) to call it as an external TUI program, just requires introducing Python as a separate dependency, the integration of `fzf` and `zoxide` is also achieved in this way - they are external TUI apps.

4rtemis-4rrow:
alright, great, I'll write it today, create a repo for it, and send you the link once it's done

would also be useful for people who want to use such a program, but don't want to use yazi, so there's that

c02y:
There is a bash script called [bashmount](https://github.com/jamielinux/bashmount/), I used it in ranger or just in terminal, I tried it in yazi, it works but it doesn't display input.

sxyazi:
Made a plugin for this, https://github.com/yazi-rs/plugins/tree/main/mount.yazi

PR #1977, opened 2024-11-30
feat: Mount manager

SL-RU:
Hi, I think mount points management is important part of file manager and because of that I want to implement that feature. This pull request is the starting point. It was mentioned here #1272  #927 . I think it cannot be fully done using plugin system and must be part of this project.

Currently there is only data from `/proc/mounts` in the menu, but later it must contain data same as lsblk (including not mounted yet volumes) and from fstab (not mounted yet volumes). 

Functionality now is limited: only change direcory to mountpoint. In plans: mount, unmount, poweroff and show info about partition.

Keymap:

In file manager key `M` opens mount manager.

In mount manager up and down scrolls; `l` or `<Enters>` changes directory to selected. Mountpoint list is updated only when opens or arrow keys pressed, later maybe makes sense to update on timer.

Only UNIX systems are supported now for demonstration purposes. But later is not a problem to add mac and win.

Please give some feedback, can it be merged in the future? Do you think it is required feature in your project?

image

### DIFF
--- a/yazi-config/preset/keymap.toml
+++ b/yazi-config/preset/keymap.toml
@@ -195,6 +195,9 @@ keymap = [
 	{ on = "<Up>",   run = "arrow -1", desc = "Move cursor up" },
 	{ on = "<Down>", run = "arrow 1",  desc = "Move cursor down" },
 
+	{ on = "<Enter>", run = "mountpoint_cd", desc = "Change directory to selected mountpoint" },
+	{ on = "l", run = "mountpoint_cd",  desc = "Change directory to selected mountpoint" },
+
 	# Help
 	{ on = "~",    run = "help", desc = "Open help" },
 	{ on = "<F1>", run = "help", desc = "Open help" },
--- a/yazi-config/preset/keymap.toml
+++ b/yazi-config/preset/keymap.toml
@@ -151,6 +151,9 @@ keymap = [
 	# Tasks
 	{ on = "w", run = "tasks_show", desc = "Show task manager" },
 
+        # Mount manager
+	{ on = "M", run = "mount_show", desc = "Show mount manager" },
+
 	# Help
 	{ on = "~",    run = "help", desc = "Open help" },
 	{ on = "<F1>", run = "help", desc = "Open help" },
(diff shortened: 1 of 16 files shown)


sxyazi:
Thank you so much for your PR! 

This is a feature I've always wanted to support but just didn't have the time to complete, so I'm really happy to see someone take it up, especially since you're the original author of [`ranger_udisk_menu`](https://github.com/SL-RU/ranger_udisk_menu), which was my original inspiration. I really appreciate your outstanding contributions to the entire file manager community with such a great plugin!

I think it's best to offer this as a plugin rather than a built-in feature, and expose the API for getting mount information to Lua. This way, it can also benefit other plugins. 

If many users need it, we can offer it as a built-in plugin, but initially, it's better to maintain it as a separate repository. So that we can get quick issue fixes independent of Yazi's release cycle, just like with the [`git.yazi` plugin](https://github.com/yazi-rs/plugins/tree/main/git.yazi). 

I'm more than happy to explore any blockers in implementing this via the plugin system, and if necessary, I can provide any missing plugin APIs — just let me know!

SL-RU:
Thank you for such an answer! I get your point and agreed that this functionality is possible to implement with extended API! 

I'll create list of required API and we'll decide how to implement it better.

SL-RU:
I understand your perspective and agree that leveraging an extended API is a viable approach. However, I believe that integrating a mount manager directly into the Rust codebase is essential for achieving a robust and seamless user experience, especially given the limitations of the current Lua API in delivering a sophisticated user interface. I couldn't imagine situation when mount manager is not required for user.

My suggestion is to implement the core mount manager functionality in Rust, ensuring multi-platform support for essential features such as viewing mount points and unmounted drives, navigating to mount points, and performing mount and unmount operations. This would provide a solid foundation for users across different systems.

Subsequently, we can develop a Lua API to allow for further customization and extension through functional plugins. This hybrid approach would combine the stability and performance of Rust with the flexibility of Lua, catering to both basic and advanced user needs.

I look forward to collaborating and contributing on this and exploring how we can best implement these features to enhance the project.

SL-RU:
@sxyazi hi, I've made plugin :smile: 

It is in early stage, I'll debug it a little bit and make it public. It is LUA plugin with Rust application. Rust app based on linux sysfs, and udisk2 via dbus.

https://github.com/SL-RU/mount.yazi
https://github.com/SL-RU/mmt

screencast:
image

SL-RU:
@sxyazi can you give some feedback please?

sxyazi:
Sorry for the delay. I did some research on supporting this across 3 platforms (Linux, macOS, Windows). Since I'm not familiar with Linux and Windows — those aren't my work systems, it took me some time.

I took a look at the `mount.yazi` plugin, and it's really cool! I really appreciate your effort. Is it possible to move some of the `mmt` code into Yazi as a Lua API? We could start by just listing devices.

If you're not familiar with the Lua part, that's totally fine! You can just create a simple function for me, and I'll port it to Lua.

In the meantime, I'll try to implement the device listing for macOS, as it's my work system and should be lot easier for me. Once we have that API, we can start working on the UI in the plugin. If you're not familiar with Yazi's UI plugin API, no worries — just leave it to me, and I'll provide any help needed to implement it.

As for Windows, it looks like it will require calling some Windows-specific APIs. If you're familiar with those, please try to implement a function for getting the device list there as well. If not, that's fine too — we can leave it for now and wait until someone familiar with Windows platforms is available to handle it.

SL-RU:
Thank you for the reply, I'll add windows support and api for Linux. 

It requires dbus and udisk for mounting, but these libraries are not required to be installed on the machine, only few rust crates.

sxyazi:
Superseded by https://github.com/sxyazi/yazi/pull/2199

PR #1991, opened 2024-12-05
fix: add missing rar mime type

hankertrix:
The newest version of file, file 5.46, now classifies `rar` archives as `application/vnd.rar` instead of `application/rar`.

There is also a [bug](https://bugs.astron.com/view.php?id=571) in this newest version of file that classifies some `zip` archives as `application/octet-stream`, but hopefully that will get fixed before Yazi 0.4 goes live so we don't have to classify a data stream as an archive.

### DIFF
--- a/yazi-config/preset/theme-base.toml
+++ b/yazi-config/preset/theme-base.toml
@@ -205,7 +205,7 @@ rules = [
 
 	# Archives
 	{ mime = "application/{,g}zip", fg = "red" },
-	{ mime = "application/{tar,bzip*,7z-compressed,xz,rar}", fg = "red" },
+	{ mime = "application/{tar,bzip*,7z-compressed,xz,rar,vnd.rar}", fg = "red" },
 
 	# Documents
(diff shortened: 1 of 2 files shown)


sxyazi:
Thanks for the PR, this is really a headache. 

I've also been bothered by the differences between versions of `file(1)` in the past, https://github.com/sxyazi/yazi/pull/1927. Unfortunately, this fix won't work for users who have already copied the Yazi config, because their configs still use `rar` instead of `vnd.rar`. 

I think it would be best to change the built-in [`mime` plugin](https://github.com/sxyazi/yazi/blob/main/yazi-plugin/preset/plugins/mime.lua) to replace the `vnd.rar` returned by the newer version of `file(1)` with `rar`. 

I'll look into this later to see if it's feasible — specifically, whether `rar` is being used for files other than RAR files. If so, replacing `vnd.rar` with `rar` might cause potential conflicts.

hankertrix:
Ah I see, `file(1)` inconsistencies are really a pain.

I’m glad replacing `vnd.rar` with `rar` works.

PR #2006, opened 2024-12-07
fix: respect `PREVIEW.image_quality` for the default PDF previewer

gaesa:
See also: https://man.archlinux.org/man/pdftoppm.1#quality


sxyazi (approved):
Thank you!

PR #2007, opened 2024-12-07
fix: avoid upscaling for the video previewer

gaesa:
See also:
- https://trac.ffmpeg.org/wiki/Scaling
- https://superuser.com/questions/566998/how-can-i-fit-a-video-to-a-certain-size-but-dont-upscale-it-with-ffmpeg

Additionally, ensuring even dimensions (using `-2`) may not be necessary, as these frames are only being extracted as images without any further encoding.


sxyazi (approved):
Nice catch, thanks!

PR #2012, opened 2024-12-08
refactor: replace `format!` with `concat!` for string literals

Integral-Tech:
- Although `format!` macro is very powerful, it introduces overhead of memory allocation on the heap compared to `&str`. Therefore, when `format!` macro is unnecessary, we should avoid using it.

### DIFF
--- a/yazi-dds/src/body/hey.rs
+++ b/yazi-dds/src/body/hey.rs
@@ -15,7 +15,7 @@ pub struct BodyHey {
 impl BodyHey {
 	#[inline]
 	pub fn owned(peers: HashMap<u64, Peer>) -> Body<'static> {
-		Self { peers, version: BodyHi::version() }.into()
+		Self { peers, version: BodyHi::version().to_string() }.into()
 	}
 }
 
--- a/yazi-dds/src/body/hey.rs
+++ b/yazi-dds/src/body/hey.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::{borrow::Cow, collections::HashMap};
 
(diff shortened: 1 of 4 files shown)


yazi-dds/src/body/hi.rs
```diff
@@ -24,7 +24,9 @@ impl<'a> BodyHi<'a> {
 	}
 
 	#[inline]
-	pub fn version() -> String { format!("{} {}", env!("CARGO_PKG_VERSION"), env!("VERGEN_GIT_SHA")) }
+	pub fn version() -> String {
```
sxyazi, line 27:
```suggestion
	pub fn version() -> &'static str {
```

Integral-Tech:
@sxyazi Done :)

yazi-dds/src/body/hey.rs
```diff
@@ -15,7 +15,7 @@ pub struct BodyHey {
 impl BodyHey {
 	#[inline]
 	pub fn owned(peers: HashMap<u64, Peer>) -> Body<'static> {
-		Self { peers, version: BodyHi::version() }.into()
+		Self { peers, version: BodyHi::version().to_string() }.into()
```
sxyazi, line 18: Please also change the `version`'s type to `&'static str` or `Cow<'static, str>`

Integral-Tech: @sxyazi Changing type of `version` to `&'static str` gives rise to this error:
```
error[E0521]: borrowed data escapes outside of associated function
```

sxyazi (approved):
Thanks!

Issue #2026, opened 2024-12-11
Suppress deprecation warnings


netmute:
### `yazi --debug` output

```Shell
Yazi
    Version: 0.4.1 (Homebrew 2024-12-10)
    Debug  : false
    Triple : aarch64-apple-darwin (macos-aarch64)
    Rustc  : 1.83.0 (90b35a62 2024-11-26)

Ya
    Version: 0.4.1 (Homebrew 2024-12-10)

Emulator
    Brand.from_env      : Some(Kitty)
    Emulator.detect     : Emulator { kind: Left(Kitty), light: true, cell_size: Some((15, 27)) }
    Emulator.detect_full: Ok(Emulator { kind: Left(Kitty), light: true, cell_size: Some((15, 27)) })

Adapter
    Adapter.matches: Kgp

Desktop
    XDG_SESSION_TYPE           : None
    WAYLAND_DISPLAY            : None
    DISPLAY                    : None
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

NVIM
    NVIM          : false
    Neovim version: 0.10.2

Variables
    SHELL           : Some("/opt/homebrew/bin/fish")
    EDITOR          : Some("nvim")
    VISUAL          : None
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
    FZF_DEFAULT_OPTS: None

Text Opener
    default     : Some(Opener { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block-create: Some(Opener { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block-rename: Some(Opener { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })

Multiplexers
    TMUX               : 0
    tmux version       : No such file or directory (os error 2)
    tmux build flags   : enable-sixel=Unknown
    ZELLIJ_SESSION_NAME: None
    Zellij version     : No such file or directory (os error 2)

Dependencies
    file          : 5.41
    ueberzugpp    : No such file or directory (os error 2)
    ffmpeg/ffprobe: 7.1 / 7.1
[...]

PR #2027, opened 2024-12-11
feat: new option to suppress 0.4 deprecation warnings

sxyazi:
An implementation of the feature request: https://github.com/sxyazi/yazi/issues/2026

Closes https://github.com/sxyazi/yazi/issues/2026

```toml
# yazi.toml
[manager]
_v4_suppress_deprecation_warnings = true
```

### DIFF
--- a/yazi-config/src/manager/manager.rs
+++ b/yazi-config/src/manager/manager.rs
@@ -26,6 +26,10 @@ pub struct Manager {
 	pub scrolloff:    u8,
 	pub mouse_events: MouseEvents,
 	pub title_format: String,
+
+	// TODO: remove this in Yazi 0.4.2
+	#[serde(default)]
+	pub _v4_suppress_deprecation_warnings: bool,
 }
 
 impl FromStr for Manager {
--- a/yazi-plugin/src/bindings/cha.rs
+++ b/yazi-plugin/src/bindings/cha.rs
@@ -18,10 +18,9 @@ impl<T: Into<yazi_fs::Cha>> From<T> for Cha {
 	fn from(cha: T) -> Self { Self(cha.into()) }
 }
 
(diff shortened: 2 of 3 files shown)


uncenter:
Why is this version specific? Wouldn't it make sense to have it as a permanent field?

sxyazi:
You mean using `suppress_deprecation_warnings` instead of `_v4_suppress_deprecation_warnings`?

The issue with this is that if a user enabled it in v0.4 and forgot to remove it, since it's not version-specific, they won't even see the deprecation warnings for the v0.5 API when they upgrade to v0.5. 

As a result, they might run into issues when the v0.5 API compatibility layer is completely removed, without realizing it.

uncenter:
This seems like a hack though. Maybe "suppress_deprecation_warnings" has a value like `"0.4"` where it targets warnings in that range? Just something more dynamic. I'd hate to see this be removed and then requested again and brought back for _v5_... etc.

sxyazi:
My plan is to remove this option when removing the 0.4 compatibility layer, since it feels weird to still have a `suppress_deprecation_warnings` if all the compatibility code (warnings) is gone as there's nothing left to "suppress". 

If we're going to make it permanent, we'd need to reconsider its name and position, placing it under `[manager]` feels odd.

Like the `[headsup]` and `disable_exec_warn` option [introduced in 0.3](https://github.com/sxyazi/yazi/pull/1045/files#diff-5ab1e81c585ce8f7ecd1931c48a9374a1b14e74748989d986c0c030e13358110), which were eventually removed along with the compatibility code though.

Overall, I think making it permanent is a bit of an overkill, especially since it's just a temporary workaround for the transition.

uncenter:
Speaking as the person who edits the schemas adding all these temporary config options is not a good solution 😅 
Maybe add an option in the TUI itself that toggles an internal state property, so outside of the config? It can be like this, for a specific version, and cleared by Yazi by later versions.

PR #2062, opened 2024-12-17
feat: expose file and directory creation APIs

### LINKED ISSUE #1709: Expose file and directory creation to plugin developers as part of `fs`

hankertrix:
Follow up to #1777.

I need some guidance on the last point regarding the file handler interoperating with other APIs. The code currently doesn't compile as I'm not really sure what to return or how to design this return value of the open function.

Closes #1709.

### DIFF
--- a/yazi-plugin/src/fs/fs.rs
+++ b/yazi-plugin/src/fs/fs.rs
@@ -1,9 +1,40 @@
 use globset::GlobBuilder;
-use mlua::{ExternalError, ExternalResult, Function, IntoLua, IntoLuaMulti, Lua, MetaMethod, Table, Value};
+use mlua::{
+	ExternalError, ExternalResult, Function, IntoLua, IntoLuaMulti, Lua, MetaMethod, Table, Value,
+};
 use tokio::fs;
 use yazi_fs::remove_dir_clean;
 
-use crate::{Error, bindings::{Cast, Cha}, file::File, url::{Url, UrlRef}};
+use crate::{
+	Error,
+	bindings::{Cast, Cha},
+	file::File,
+	url::{Url, UrlRef},
+};
+
+const READ: u8 = 1;
+const WRITE: u8 = 2;
+const APPEND: u8 = 4;
+const TRUNCATE: u8 = 8;
+const CREATE: u8 = 16;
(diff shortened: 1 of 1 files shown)


hankertrix:
@sxyazi What do you think of the bit flags parameter as the options for the `open` function? Would you rather use a Lua table?

Also, the return value of the open function currently just returns a Yazi `File` object, as I am not sure how to design a separate structure to interop with the Command plugin API.

sxyazi:
Thanks for the PR! 

> On second thought, would the `open` function be considered unnecessary? Lua already has an `io.open` function to create files which supports everything from [OpenOptions](https://doc.rust-lang.org/std/fs/struct.OpenOptions.html) except truncate mode.

Yeah, Lua already provides `io.open()`, so reimplementing it in Yazi doesn't seem to be a high priority, how do you feel about separating the `fs.open()` part from this PR into a standalone one? That way, we can quickly merge `fs.create_dir()`, while I still have some details I want to discuss about `fs.open()`, I need to think more about the API design. 

Or, if `io.open()` already meets your needs and `fs.open()` is no longer valuable to you, you can just discard it. I'll revisit how to add it when I have some time.

hankertrix:
> Yeah, Lua already provides `io.open()`, so reimplementing it in Yazi doesn't seem to be a high priority, how do you feel about separating the `fs.open()` part from this PR into a standalone one?

Oh yeah, that would be great. I'll do that.

> I need to think more about the API design.

Sure thing, no worries!

PR #2068, opened 2024-12-19
feat: expose directory creation as part of the `fs` module

hankertrix:
Split from #2062.

This implements the `fs.create_dir` function for plugin developers to use.


sxyazi (approved):
Thank you so much!

hankertrix:
You’re welcome!

sxyazi:
Hi, on second thought, I renamed it to `fs.create()`, so it's consistent with `fs.remove()`, and it also allows us to expand with other behaviors later, like `fs.create("file", ...)` for creating files, not just directories. Hope that's fine with you!

hankertrix:
Yeap, that's fine with me!

Having one function to create multiple things is really nice!

PR #2077, opened 2024-12-21
feat: add `cbr` and `cbz` as valid archive extensions

AminurAlam:
https://en.wikipedia.org/wiki/Comic_book_archive


sxyazi (approved):
Thanks

PR #2089, opened 2024-12-23
feat: hide specified files and dirs

### LINKED ISSUE #694: Hide specified files or directories

It's nice to have control over file/folder visibility, e.g. folders like `Desktop`, `go` and `steam` in my home directory are an eyesore for me, I would like to hide them, and some people may want to treat their `.config` as visible folder****

There should be an option in `yazi.toml` where the user can add rules for how to treat certain files and folders. It probably should resides under the `[manager]` category alongside with `show_hidden`. 

**Which is a more intuitive way to represent this option?**
1) An array of patterns with a specification how to treat them
```toml
[manager]
visibility_rules = [
    { pattern = "~/.config", visible = true,  }
    { pattern = "~/Desktop", visible = false, }
    { pattern = "~/go"     , visible = false, }
]
```
2) Two arrays - one for visible entries, one for hidden entries
```toml
[manager]
treat_as_visible = [ "~/.config", ]
treat_as_hidden = [ "~/Desktop", "~/go", ]
```

_No response_

### Additional context

Requested by @5ouma, https://github.com/sxyazi/yazi/issues/51#issuecomment-1725011140

TobisLee:
Try to fix #694.

Works with the given regex string:
https://github.com/TobisLee/yazi/blob/2fdd7b6f554dd4e15648ed8f611bd83cedd511c8/yazi-shared/src/url/urn.rs#L22
but failed with hidden_rule config:
https://github.com/TobisLee/yazi/blob/2fdd7b6f554dd4e15648ed8f611bd83cedd511c8/yazi-shared/src/url/urn.rs#L23

Cargo shows
```txt
error: cyclic package dependency: package `yazi-config v0.4.3 (/home/tlss/repos/rust_project/yazi/yazi-config)` depends on itself. Cycle:
package `yazi-config v0.4.3 (/home/tlss/repos/rust_project/yazi/yazi-config)`
    ... which satisfies path dependency `yazi-config` (locked to 0.4.3) of package `yazi-shared v0.4.3 (/home/tlss/repos/rust_project/yazi/yazi-shared)`
    ... which satisfies path dependency `yazi-shared` (locked to 0.4.3) of package `yazi-fs v0.4.3 (/home/tlss/repos/rust_project/yazi/yazi-fs)`
    ... which satisfies path dependency `yazi-fs` (locked to 0.4.3) of package `yazi-config v0.4.3 (/home/tlss/repos/rust_project/yazi/yazi-config)`
    ... which satisfies path dependency `yazi-config` (locked to 0.4.3) of package `yazi-adapter v0.4.3 (/home/tlss/repos/rust_project/yazi/yazi-adapter)`
```

### DIFF
--- a/yazi-config/src/mgr/mgr.rs
+++ b/yazi-config/src/mgr/mgr.rs
@@ -26,6 +26,7 @@ pub struct Mgr {
 	pub scrolloff:    u8,
 	pub mouse_events: MouseEvents,
 	pub title_format: String,
+	pub ignore_rule:  String,
 }
 
 impl FromStr for Mgr {
--- a/yazi-core/src/tab/folder.rs
+++ b/yazi-core/src/tab/folder.rs
@@ -2,7 +2,7 @@ use std::mem;
 
 use yazi_config::{LAYOUT, MGR};
 use yazi_dds::Pubsub;
-use yazi_fs::{File, Files, FilesOp, FolderStage, Step, cha::Cha};
+use yazi_fs::{File, Files, FilesOp, FolderStage, Step, cha::Cha, Filter};
 use yazi_proxy::MgrProxy;
 use yazi_shared::{Id, url::{Url, Urn, UrnBuf}};
 
@@ -24,7 +24,7 @@ impl Default for Folder {
 		Self {
 			url:    Default::default(),
 			cha:    Default::default(),
-			files:  Files::new(MGR.show_hidden),
+			files:  Files::new(MGR.show_hidden, Filter::new(MGR.ignore_rule.as_str(), yazi_fs::FilterCase::Smart).ok()),
 			stage:  Default::default(),
 			offset: Default::default(),
 			cursor: Default::default(),
--- a/yazi-fs/src/files.rs
+++ b/yazi-fs/src/files.rs
@@ -18,6 +18,7 @@ pub struct Files {
 
 	sorter:      FilesSorter,
 	filter:      Option<Filter>,
+	ignore_filter: Option<Filter>,
 	show_hidden: bool,
 }
 
@@ -28,7 +29,7 @@ impl Deref for Files {
 }
 
 impl Files {
-	pub fn new(show_hidden: bool) -> Self { Self { show_hidden, ..Default::default() } }
+	pub fn new(show_hidden: bool, ignore_filter: Option<Filter>) -> Self { Self { show_hidden, ignore_filter, ..Default::default() } }
 
 	pub async fn from_dir(dir: &Url) -> std::io::Result<UnboundedReceiver<File>> {
 		let mut it = fs::read_dir(dir).await?;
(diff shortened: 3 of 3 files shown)


Laughing-q:
@sxyazi Hi! Thanks for your awesome work on this awesome file manager! Wanted to push this feature a little bit more here haha, recently I'm switching my primary file manager from `lf` to `yazi`, and as a python developer I really want to hide the annoying `__pycache__` folders and files. 
I used to use `set hiddenfiles .*:__pycache__:*.pyc` to hide it in `lf` manager. It'd be perfect if `yazi` has similar option to do so! Thanks!
jsSpPZr59Z

sxyazi:
@TobisLee This feature should not be implemented in `yazi-shared/src/url/urn.rs` because `yazi-shared` is a fairly low-level crate that provides shared functions and should not contain any business-related code. 

You need to implement it in `yazi-fs/src/files.rs` and add it as a field in `Files`, similar to the existing `filter: Option<Filter>`:

https://github.com/sxyazi/yazi/blob/00e8adc3decc370a7e14caaeae3676361549fceb/yazi-fs/src/files.rs#L20

 Then, pass it from the outside, just like what the `set_filter` method does:

https://github.com/sxyazi/yazi/blob/00e8adc3decc370a7e14caaeae3676361549fceb/yazi-fs/src/files.rs#L360

TobisLee:
Hey, @sxyazi , I implement `ignore_filter` in `files.rs` as you said and test with my environment. I leave some comments about my confusion, hope to hear your thoughts. If there anything else I can do, please let me know.

TobisLee:
Just the comments in review? I can see them on my timeline. Do I need some settings? 
图片

TobisLee:
I submit the review, the comments should be seen now.

TobisLee:
Hey, @sxyazi , can you review this code plz?

boydaihungst:
It would be awesome if we could use a custom Lua function to check whether the file should be hidden or not, instead of hard-coding it with a regex or pattern. Something like: 
```lua
files.should_hide(function(file)
  return not has_tag(file)
end)
```
:face_holding_back_tears:

TobisLee:
Hey @sxyazi . If this PR has problem or can be improved, please let me know. Otherwise can it be merged asap?

sxyazi:
Sorry for the delay, I will review it in the next few days. Can you rebase it to main?

TobisLee:
Okay

sxyazi:
Thank you for the update! I have some change requests:

Please move `hidden_rule` into its own dedicated section:

```toml
# yazi.toml
[files]
excludes = [
  { url = [".{git,DS_Store}", "__pycache__", "*.pyc"], in = "/code/**" },
  { url = ".DS_Store", in = "*" }
]
```

Here, `url` is a glob expression that supports both a string (a single glob) and an array (multiple globs), and `in` specifies the directory where the expression applies, which is also a glob.

When parsing the configuration from TOML, store `url` uniformly as a `Vec` and precompile it into regular expressions — you can compile it into globs using [`globset`](https://docs.rs/globset/latest/globset/), and then [obtain the corresponding regular expression](https://docs.rs/globset/latest/globset/struct.Glob.html#method.regex).

When applying the rules to directories, match the directory path against the first excludes rule, and compile the resulting `Vec<Regex>` into a [`RegexSet`](https://docs.rs/regex/latest/regex/struct.RegexSet.html), using the `RegexSet` to match all files for better performance.

TobisLee:
All right, I will try this. Is there some reference code in yazi?

sxyazi:
If you mean config parsing, all the code is in `yazi/yazi-config/src`

sxyazi:
I'm going to close this because it still needs some work and hasn't seen any activity recently. @TobisLee  thanks for the effort!

PR #2136, opened 2025-01-01
feat: add more Ogg formats

hankertrix:
The `ogx` file extension is a container that can contain any Ogg format, so I'm not sure if it should be classified as a video, as it seems to be more like an archive, but for Ogg formats.

Source:
https://wiki.xiph.org/index.php/MIME_Types_and_File_Extensions


sxyazi:
Hi, thanks for the PR! 

The icons for Yazi come from https://github.com/nvim-tree/nvim-web-devicons and are automatically generated through https://github.com/sxyazi/yazi/blob/main/scripts/icons/generate.lua script.

So it's best to submit this icon upstream so it will be easier to update.

hankertrix:
Oh okay, will do!

hankertrix:
Closing in favour of the upstream pull request https://github.com/nvim-tree/nvim-web-devicons/pull/538.

Issue #2147, opened 2025-01-02
Delete/uninstall plugin


og900aero:
### `yazi --debug` output

```Shell
Yazi
    Version: 0.4.2 (5cfcab3 2024-12-20)
    Debug  : false
    Triple : x86_64-unknown-linux-gnu (linux-x86_64)
    Rustc  : 1.83.0 (90b35a62 2024-11-26)

Ya
    Version: 0.4.2 (5cfcab3 2024-12-20)

Emulator
    Brand.from_env      : None
    Emulator.detect     : Emulator { kind: Right(Unknown { kgp: false, sixel: false }), light: false, cell_size: None }
    Emulator.detect_full: Ok(Emulator { kind: Right(Unknown { kgp: false, sixel: false }), light: false, cell_size: None })

Adapter
    Adapter.matches: X11

Desktop
    XDG_SESSION_TYPE           : Some("tty")
    WAYLAND_DISPLAY            : None
    DISPLAY                    : Some(":0")
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

NVIM
    NVIM          : false
    Neovim version: No such file or directory (os error 2)

Variables
    SHELL           : Some("/bin/bash")
    EDITOR          : Some("micro")
    VISUAL          : Some("micro")
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
    FZF_DEFAULT_OPTS: Some("--exact --color=fg:#f8f8f2,hl:#bd93f9 --color=fg+:#f8f8f2,bg+:#44475a,hl+:#bd93f9 --color=info:#ffb86c,prompt:#50fa7b,pointer:#ff79c6 --color=marker:#ff79c6,spinner:#ffb86c,header:#6272a4")

Text Opener
    default     : Some(Opener { run: "$EDITOR \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block-create: Some(Opener { run: "$EDITOR \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block-rename: Some(Opener { run: "$EDITOR \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })

Multiplexers
    TMUX               : 0
[...]

PR #2181, opened 2025-01-09
feat: new `ya pack -d` subcommand to delete packages

MrAsler:
Closes https://github.com/sxyazi/yazi/issues/2147

### Details about the implementation

To add the `delete` command, I followed the same pattern that existed in the `src/args.rs`, `src/main.rs` and `src/package/*` directories.

I tried to follow the existing code style, and kept refactorings to a minimum - I created two helper methods - `find_dep_in_package` and `deployed_directory`. If there are any naming issues or you would like the code to be structured differently, feel free to say!

#### Some code is still not deleted
As things are, this will not delete the package code that is present in `Xdg::local()` directory, as I don't understand why the `state_dir()` of a dependency is using an hash to identify the package (`format!("{:x}", XxHash3_128::oneshot(self.remote().as_bytes()))`) instead of just using the package name. 
I'm assuming it is desired to also remove the content of this `state_dir()`, but I'll `await` for the answer 🕵️ 

#### Can the field `Dependency.use_` be used the dependency ID?
The function `Package::add(..)` was detecting if a package existing with the following code:
```rust
if self.plugins.iter().any(|d| d.repo == dep.repo && d.child == dep.child) {
	bail!("Plugin `{name}` already exists in package.toml");
}
```
However, given that the objective is to see if a package already exists in `package.toml` and each package keeps the `use` value there, I'm under the impression that this can be simplified to using `use_`. 
If this `use` field is internal and should not be used, should there be a way of giving an ID to a package, or should we keep using the `parent` and `child` names?

### DIFF
--- a/yazi-cli/src/args.rs
+++ b/yazi-cli/src/args.rs
@@ -57,6 +57,9 @@ pub(super) struct CommandPack {
 	/// Add a package.
 	#[arg(short = 'a', long)]
 	pub(super) add:     Option<String>,
+	/// Delete a package.
+	#[arg(short = 'd', long)]
+	pub(super) delete:  Option<String>,
 	/// Install all packages.
 	#[arg(short = 'i', long)]
 	pub(super) install: bool,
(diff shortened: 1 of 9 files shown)


sxyazi (approved):
Thank you!

mslalith:
I was just looking for this. When is the next release?

PR #2185, opened 2025-01-09
feat: new `--no-cwd-file` option for the `close` command

igxlin:
It will be helpful to forward `close` params to `quit`. Then users can use commands like `close --cwd-file`, `close --no-cwd-file`, to control the quit behavior indirectly.

### DIFF
--- a/yazi-core/src/manager/commands/close.rs
+++ b/yazi-core/src/manager/commands/close.rs
@@ -1,12 +1,26 @@
 use yazi_shared::event::CmdCow;
 
-use crate::{manager::Manager, tasks::Tasks};
+use crate::{manager::Manager, tasks::Tasks, manager::commands::quit};
+
+#[derive(Default)]
+struct Opt {
+	no_cwd_file: bool,
+}
+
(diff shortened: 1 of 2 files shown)


sxyazi (changes requested):
Please create a new `Opt` for the `close` command like `quit` does:

https://github.com/sxyazi/yazi/blob/1e0c5ff61c23cb904bb3e2f45297fc34326fbb60/yazi-core/src/manager/commands/quit.rs#L12-L14

to make it safer. Otherwise, when new parameters are added to `quit`, there could be unintended parameter passthrough.

igxlin:
Updated: https://github.com/sxyazi/yazi/pull/2185/commits/f3b5dcf125d6a37b343549678637a9993a496601 

@sxyazi please help review.

I made `quit::Opt` as public in order to pass arguments. Looks like all `Opt` of other commands are private ... Please let me know if I should do it by other ways. Thanks.

sxyazi (approved):
Thank you!

PR #2186, opened 2025-01-09
fix: incorrect monorepo sub-plugin path resolution

lpnh:
When attempting to install a plugin with a format like `yazi-rs/plugins:mime-ext`, an error message "No such file or directory" is returned.

The error occurs because the constructed local path to the plugin source is missing the `.yazi` extension. This results in incorrect paths such as `mime-ext/init.lua` instead of the correct `mime-ext.yazi/init.lua`.

This patch addresses this issue by ensuring the `.yazi` extension is correctly included when constructing the local path for these plugins.

Thanks for your amazing work!


sxyazi:
Oh right, forgot to test sub-plugins, thanks for the patch!

sxyazi (approved):
Thank you!

PR #2188, opened 2025-01-10
feat: arm64 Snap package

jnsgruk:
This PR makes a few small adjustments to the snap package.

Firstly, it uses packages from the Ubuntu archive (or Snap store) for `yazi`'s dependencies such as `fzf`, `zoxide`, `jq`, etc.

The one exception is `magick`, where in the Ubuntu package `magick` binary is not present (I don't know why!), so we build from source. This has the advantage over fetching the prebuilt binary in that it'll work across architectures.

Also re-enables `arm64` (see [post](https://forum.snapcraft.io/t/help-needed-cant-build-arm64-snap-with-github-actions/41356)).

Tested this on my `amd64` laptop - as far as I can tell it all works (including the image previews 😎 )

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -11,7 +11,7 @@ confinement: classic
 
 platforms:
   amd64:
-  # arm64:
+  arm64:
 
 apps:
   yazi:
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -23,34 +23,31 @@ parts:
   yazi:
     plugin: rust
     source: https://github.com/sxyazi/yazi.git
-    build-packages:
-      - wget
+    stage-snaps:
+      - jq
+    stage-packages:
+      - 7zip
+      - fd-find
+      - fzf
+      - ripgrep
+      - zoxide
     override-build: |
       craftctl default
       craftctl set version=$(git describe --tags --abbrev=0)
+    organize:
+      # Ubuntu's `fd` package installs a binary named `fdfind`. Rename it in the snap.
+      usr/bin/fdfind: usr/bin/fd
 
-      wget ripgrep-14.1.1-aarch64-unknown-linux-gnu.tar.gz
-      tar fxz ripgrep-*.tar.gz
-      mv ripgrep-*/rg $CRAFT_PART_INSTALL/bin/
-
-      wget https://github.com/sharkdp/fd/releases/download/v10.2.0/fd-v10.2.0-aarch64-unknown-linux-gnu.tar.gz
-      tar fxz fd-*.tar.gz
-      mv fd-*/fd $CRAFT_PART_INSTALL/bin/
-
-      wget https://github.com/jqlang/jq/releases/download/jq-1.7.1/jq-linux-amd64
-      mv jq-linux-amd64 $CRAFT_PART_INSTALL/bin/jq
-
-      wget https://imagemagick.org/archive/binaries/magick
-      mv magick $CRAFT_PART_INSTALL/bin/magick
-
-      wget https://github.com/junegunn/fzf/releases/download/v0.56.2/fzf-0.56.2-linux_amd64.tar.gz
-      tar fxz fzf-*.tar.gz
-      mv fzf $CRAFT_PART_INSTALL/bin/
-
-      wget https://www.7-zip.org/a/7z2408-linux-x64.tar.xz
-      tar fxz 7z*.tar.xz
-      mv 7zz 7zzs $CRAFT_PART_INSTALL/bin/
-
-      wget https://github.com/ajeetdsouza/zoxide/releases/download/v0.9.6/zoxide-0.9.6-x86_64-unknown-linux-musl.tar.gz
-      tar fxz zoxide-*.tar.gz
-      mv zoxide $CRAFT_PART_INSTALL/bin/
+  # The Ubuntu `imagemagick` package does not ship a `magick` binary,
+  # which `yazi` looks for. Building the package oursleves yields the
+  # expected binary without much overhead, and works across platforms.
+  magick:
+    plugin: autotools
+    source: https://imagemagick.org/archive/ImageMagick.tar.gz
+    source-type: tar
+    stage-packages:
+      - libgomp1
+    autotools-configure-parameters:
+      - --prefix=/usr
+    build-attributes:
+      - enable-patchelf
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -31,6 +31,8 @@ parts:
       - fd-find
       - ffmpeg
       - fzf
+      - libglu1-mesa
+      - libglut3.12
       - poppler-utils
       - ripgrep
       - wl-clipboard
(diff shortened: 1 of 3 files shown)


jnsgruk:
image

This is a screenshot of the snap running on an arm64 azure instance :)

sxyazi:
Awesome, thanks for the PR!

I have some questions and change requests:

- The packages from the Ubuntu archive (fzf, zoxide, jq, etc.) don't require separate builds right - they're already executable binaries?
- Is it possible to use the official binaries for ImageMagick across different platforms to avoid any building?
- Please add the other [missing dependencies](https://yazi-rs.github.io/docs/installation): ffmpeg and poppler.
- Please enable this in the CI here https://github.com/sxyazi/yazi/blob/b495f78dfe7641f2047f1684561c842f20a8518b/.github/workflows/draft.yml#L147-L171

gapplef:
> Firstly, it uses packages from the Ubuntu archive (or Snap store) for yazi's dependencies such as `fzf`, `zoxide`, `jq`, etc.

A small reminder here: these dependencies are bundled for a reason.
Some packages in the Ubuntu archive are too old to function correctly, especially with older LSTs. This is the reason I initially built the snap package.
Related: https://github.com/sxyazi/yazi/discussions/487,  https://github.com/sxyazi/yazi/discussions/502

jnsgruk:
> The packages from the Ubuntu archive (fzf, zoxide, jq, etc.) don't require separate builds right - they're already executable binaries?

Correct, they're pre-built for each of the architectures, no need to build them.

> Is it possible to use the official binaries for ImageMagick across different platforms to avoid any building?

As far as I can tell, the upstream only provide binaries for x86_64, and pulling pre-built binaries from somewhere else seems a little unsafe? Is there a reason to avoid building ImageMagick? (In my testing it's fairly quick...)

> Some packages in the Ubuntu archive are too old to function correctly

Do you know which versions are required? Looks like 24.04 (which is `core24`) ships with `fd` `9.0.0` (Dec 23) and `zoxide 0.93` (latest). The linked issue about `fd` doesn't seem to be an issue with the version of `fd` in `core24`? (I'm happy to change, it just seems sensible to use the distro packages wherever possible to avoid complicating the snapcraft yaml)

jnsgruk:
I've updated the Github Action, though to use the `snapcraft remote-build` feature (to get multi-arch building) you'll need to add a secret to the repo (see [here](https://github.com/sxyazi/yazi/pull/2188/files#diff-94311e04376f1d0b288644e9f82aa4e98f60e027aa07247bab80330965d45150R168))

You can do this by running `snapcraft remote-build` locally from within this repo, and it'll create `~/.local/share/snapcraft/launchpad-credentials` which you can copy into the secret `LP_TOKEN`

jnsgruk:
(Looks like I have an issue to fix with 7z - need to step away for now but will come back to it!)

jnsgruk:
Okay! This seems to be working properly now. I've tested this as much as I can - as far as I can tell all of the PDF preview, image preview, JSON processing etc. is all working as it should.

There is a bit of a workaround in the latest commit which I can circle back to once the bug is fixed in snapcraft.

sxyazi:
That's so fast, thanks for the update! 

I'll add the missing `LP_TOKEN` after merging the PR.

snap/snapcraft.yaml
```diff
@@ -11,46 +11,64 @@ confinement: classic
...
-      - wget
+    stage-snaps:
+      - jq
+    stage-packages:
+      - 7zip-standalone
+      - chafa
+      - fd-find
+      - ffmpeg
+      - fzf
+      - libpoppler-cpp0t64
```
sxyazi, line 34: The name `libpoppler-cpp0t64` sounds like a dynamic library - does it provide the `pdftoppm` binary?

jnsgruk: Oops, good catch. I've swapped this for `poppler-utils`, which will indeed pull in this library, but also provide the `pdftoppm` binary explicitly.

jnsgruk:
Okay, think I've done tinkering again. Last commit addresses all the linter warnings that snapcraft was showing. This is essentially just cleaning up files that weren't used, and adding libraries that could have been missing.

The snap now builds with no warnings, and all still seems to work great, but with less bloat :)

snap/snapcraft.yaml
```diff
@@ -11,46 +11,153 @@ confinement: classic
...
+      - -usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libsoftokn3.so
+      - -usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libsphinxad.so.3.0.0
+      - -usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libssl3.so
+      - -usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libtheora.so.0.3.10
+      - -usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libxcb-dri2.so.0.0.0
+      - -usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libxcb-glx.so.0.0.0
+      - -usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libxcb-present.so.0.0.0
+      - -usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libxcb-sync.so.1.0.0
+      - -usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libxshmfence.so.1.0.0
+      - -usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libzvbi-chains.so.0.0.0
```
sxyazi, line 115: How should the file list be updated in the future? Is there a more automated way, like using something similar to `rm $(snapcraft unused-files)` instead?

jnsgruk: Not at the moment, but after going through this I've got a conversation with them about how we make this a little better.

That said, the linter does make it very clear at the end of the build process with messages that list all of the unused files -- so it's mostly a little bit of copy and paste each time new dependencies are added.

I might be able to shrink this list slightly by using package [slices](https://documentation.ubuntu.com/rockcraft/en/latest/explanation/chisel/) in the future, but that'd be a longer term project to have those slices created

jnsgruk:
I may be able to reduce the list (and the size of the snap) by using content snaps. Will investigate and post back here.

jnsgruk:
> I may be able to reduce the list (and the size of the snap) by using content snaps. Will investigate and post back here.

Okay, sadly not in this case. I think this is where we're at for now. If that changes, I can come back and make some adjustments 🙂

snap/snapcraft.yaml
sxyazi: Gotcha!

sxyazi (approved):
Everything looks great now, thank you so much for your help @jnsgruk!

sxyazi:
CI seems failed in https://github.com/sxyazi/yazi/actions/runs/12768983031, please have a look

jnsgruk:
> CI seems failed in https://github.com/sxyazi/yazi/actions/runs/12768983031, please have a look

Ooops, see https://github.com/sxyazi/yazi/pull/2200 🙂

PR #2203, opened 2025-01-14
ci(snap): don't use a shallow clone when checking out the repo for snap build

jnsgruk:
Resolves an error in the latest snap build run in CI. Disables shallow cloning when pulling the code.

### DIFF
--- a/.github/workflows/draft.yml
+++ b/.github/workflows/draft.yml
@@ -153,6 +153,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
       - name: Setup sccache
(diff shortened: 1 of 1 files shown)


sxyazi (approved):
Thank you!

jnsgruk:
I saw that, are you able to retry the job? Looks like an odd failure. There is a fallback behaviour for snapcraft we can try if that doesn't work.

sxyazi:
Since it still fails to build unfortunately, I've disabled it in https://github.com/sxyazi/yazi/commit/6c32e4cd02cbf952bb0635a0bbdc461ee97c3be7 for now. 

I think it should be enough if users can build it successfully, maybe we can investigate it further in the future.

jnsgruk:
No worries, I'll do some more investigation locally and see if I can come up with a PR to fix it. Not clear to me why the build is failing yet.

PR #2214, opened 2025-01-17
ci: use GitHub Linux arm64 runners for the snap build

jnsgruk:
I've taken a look at the snap build failures on arm64. It looks like snapcraft behaves subtly differently on amd64 and arm64 with respect to automatically patching ELF binaries. This PR addresses that - it turns out we don't need the manual patch for things to function. Tested locally on my machine, and everything seems to work just fine.

In addition, Github [announced](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/) yesterday that arm64 runners are now available to public repos - so this PR also simplifies the snap build pipeline to use those - which simplifies both the pipeline and (hopefully) the debugging of any issues.

### DIFF
--- a/.github/workflows/draft.yml
+++ b/.github/workflows/draft.yml
@@ -145,43 +145,27 @@ jobs:
           path: yazi-${{ matrix.target }}.zip
 
   build-snap:
-    if: false # Can't make CI pass, disable for now
     strategy:
       matrix:
-        arch:
-          - amd64
-          - arm64
-    runs-on: ubuntu-latest
+        include:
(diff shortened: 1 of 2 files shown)


sxyazi:
Thanks, the code looks much cleaner now, nice work @jnsgruk !

TIL GitHub has opened up Linux arm64 runners, that's awesome!

jnsgruk:
I did just notice a small error, will correct with a patch soon (afk today)

Issue #2215, opened 2025-01-17
Nix flake build failed


sxyazi:
In https://github.com/sxyazi/yazi/pull/2199, the minimum Rust version was bumped to 1.83.0, and the [Nix build failed](https://github.com/sxyazi/yazi/actions/runs/12827417425/job/35769327749) due to a Rust version mismatch. 

To fix it, I ran `nix flake update` to update the `flake.lock` to the latest, but the latest flake.lock seems to involve infinite recursion, causing the [build to still fail](https://github.com/sxyazi/yazi/actions/runs/12827602862/job/35769909699). 

I'm not sure how to fix this since I don't use Nix myself, so any help would be greatly appreciated!

sxyazi:
Tagging @XYenon and @eljamm maybe can take a look if you have some time :)

PR #2216, opened 2025-01-18
fix(nix): rust overlay infinite recursion

eljamm:
The `rust-overlay` project strongly recommend not overlaying the global Rust platform since that can cause a dependency loop, creating an infinite recursion (see https://github.com/oxalica/rust-overlay/issues/198#issuecomment-2566755261).

Instead, a custom `rustPlatform` can be passed to the derivation, as per the instructions in the [Nixpkgs manual](https://nixos.org/manual/nixpkgs/stable/#using-rust-nightly-in-a-derivation-with-buildrustpackage).

Fixes: https://github.com/sxyazi/yazi/issues/2215

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -21,21 +21,12 @@
       let
         pkgs = import nixpkgs {
           inherit system;
-          overlays = [
-            rust-overlay.overlays.default
-            (
-              final: prev:
-              let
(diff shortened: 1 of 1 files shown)


sxyazi (approved):
Thank you so much for the fix @eljamm !

PR #2217, opened 2025-01-18
ci(snap): correct arm64 runner OS

jnsgruk:
Correct a minor error I introduced in refactoring the snap workflow. Now this actually uses an arm runner 🤦‍♂️ 

Sorry about the noise!

### DIFF
--- a/.github/workflows/draft.yml
+++ b/.github/workflows/draft.yml
@@ -150,7 +150,7 @@ jobs:
         include:
           - os: ubuntu-24.04
             arch: amd64
-          - os: ubuntu-24.04
+          - os: ubuntu-24.04-arm
             arch: arm64
     runs-on: ${{ matrix.os }}
     steps:


sxyazi:
No worries, you're doing well, thanks for the quick fix!

.github/workflows/draft.yml
```diff
@@ -150,7 +150,7 @@ jobs:
         include:
           - os: ubuntu-24.04
             arch: amd64
-          - os: ubuntu-24.04
+          - os: ubuntu-24.04-arm
```
sxyazi, line 153: Is it possible to use `latest` instead of a fixed version here for more flexibility?

```suggestion
          - os: ubuntu-latest-arm
```

jnsgruk: I don't believe so - the docs seem to imply it's just `ubuntu-22.04-arm` or `ubuntu-24.04-arm`

sxyazi: It looks like tomorrow (January 17th, 2025) `ubuntu-latest` will be updated to `ubuntu-24.04`, https://github.com/actions/runner-images/issues/10636 - let's wait a day and see 😄

jnsgruk: Yeh of course your call. I've always avoided the `*latest` images to avoid unwanted breaking changes - though we've probably got about 2 years until `ubuntu-latest` is `26.04`

MrG9090 (approved):

sxyazi (approved):
Merging now, thank you!

MrG9090 (approved):

PR #2226, opened 2025-01-19
fix: ignore completions artifacts

euamotubaina:
Update `.gitignore` to not track the built completions files.

Added general pattern to ignore all `completions` directories in the first level of any `yazi-*` directory.


sxyazi (approved):
Nice improvement, thank you!

PR #2234, opened 2025-01-22
fix: use `u32` for parsing Linux partition blocks

lpnh:
Now it matches the `#blocks` size of the `/proc/partitions` table, for example:

```
major minor  #blocks  name

   8        0  234431064 sda
   8       16  488386584 sdb
   8       17  488385536 sdb1
 259        0  468851544 nvme0n1
 259        1    1048576 nvme0n1p1
 259        2    8388608 nvme0n1p2
 259        3  459412480 nvme0n1p3
   8       32   61587456 sdc
```


sxyazi (approved):
Ah never thought blocks would exceed u16... 

Great found, thanks for the fix!

Issue #2272, opened 2025-01-31
Crash when searching with fd


Tyarel8:
### What system are you running Yazi on?

Windows

### What terminal are you running Yazi in?

wezterm 20241205-083533-6f375e29

### `yazi --debug` output

```Shell
Yazi
    Version: 0.4.3 (b6cb1fa 2025-01-29)
    Debug  : false
    Triple : x86_64-pc-windows-msvc (windows-x86_64)
    Rustc  : 1.84.0 (9fc6b431 2025-01-07)

Ya
    Version: 0.4.3 (b6cb1fa 2025-01-29)

Emulator
    TERM                : Some("xterm-256color")
    TERM_PROGRAM        : Some("WezTerm")
    TERM_PROGRAM_VERSION: Some("20241205-083533-6f375e29")
    Brand.from_env      : Some(WezTerm)
    Emulator.detect     : Emulator { kind: Left(WezTerm), light: false, cell_size: None }
    Emulator.detect_full: Ok(Emulator { kind: Right(Unknown { kgp: false, sixel: false }), light: false, cell_size: None })

Adapter
    Adapter.matches: Iip

Desktop
    XDG_SESSION_TYPE           : None
    WAYLAND_DISPLAY            : None
    DISPLAY                    : None
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Neovim
    NVIM          : false
    Neovim version: 0.10.4

Variables
    SHELL           : Some("nu")
    EDITOR          : Some("hx")
    VISUAL          : None
    YAZI_FILE_ONE   : Some("C:\\Users\\<user>\\scoop\\apps\\git\\2.47.1.2\\usr\\bin\\file.exe")
    YAZI_CONFIG_HOME: Some("C:\\Users\\<user>\\Sync\\config\\yazi")
    YAZI_ZOXIDE_OPTS: None
[...]

PR #2286, opened 2025-02-04
fix: enable unwinding on Windows release builds

sxyazi:
Try to fix https://github.com/sxyazi/yazi/issues/2272

See https://github.com/mlua-rs/mlua/issues/520


sxyazi:
Sad, setting a build profile for a specified target is still not supported, https://github.com/rust-lang/cargo/issues/4897

PR #2294, opened 2025-02-05
feat: new `arrow top` and `arrow bot` commands to jump to the top and bottom

sxyazi:
This PR deprecates `arrow -99999999` and `arrow 99999999`, replacing them with `arrow top` and `arrow bot`, respectively. `-99999999` and `99999999` are still available but will trigger a warning.

This change is in preparation for an upcoming navigation wraparound feature — when the cursor reaches the top or bottom of the list, it will automatically jump to the opposite end.  

Some users requested this feature, and I initially [provided a plugin](https://yazi-rs.github.io/docs/tips#navigation-wraparound) as a solution. However, I believe it's worth integrating into the core. 

The issue is that when users enable this feature, pressing `gg` or `G` would cause the cursor to behave erratically since `arrow 99999999` would make it loop through the file list multiple times, so `top` and `bot` (bottom) are introduced.

### DIFF
--- a/yazi-config/preset/keymap-default.toml
+++ b/yazi-config/preset/keymap-default.toml
@@ -29,8 +29,8 @@ keymap = [
 	{ on = "<PageUp>",     run = "arrow -100%", desc = "Move cursor up one page" },
 	{ on = "<PageDown>",   run = "arrow 100%",  desc = "Move cursor down one page" },
 
-	{ on = [ "g", "g" ], run = "arrow -99999999", desc = "Move cursor to the top" },
-	{ on = "G",          run = "arrow 99999999",  desc = "Move cursor to the bottom" },
+	{ on = [ "g", "g" ], run = "arrow top", desc = "Move cursor to the top" },
+	{ on = "G",          run = "arrow bot",  desc = "Move cursor to the bottom" },
 
 	# Navigation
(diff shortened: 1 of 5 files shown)


supertrianguloid:
[This tip](https://yazi-rs.github.io/docs/tips/#navigation-wraparound) probably needs to get updated. Is there a recommendation for how to get this to work while the native wraparound is still on the way? (I don't know Lua).

supertrianguloid:
> > [This tip](https://yazi-rs.github.io/docs/tips/#navigation-wraparound) probably needs to get updated. Is there a recommendation for how to get this to work while the native wraparound is still on the way? (I don't know Lua).
> 
> That tip works fine and doesn't need any fixing.
> 
> If I must guess, you're referring to the warning that appears when using that tip, which was fixed in [52a5ab5#diff-84d365f5e3d97756e98c1133e3137398046fcddf708b807ddb2d0d6343926767](https://github.com/sxyazi/yazi/commit/52a5ab59e943447c0d6d4e04b7afffc1a4ab23f2#diff-84d365f5e3d97756e98c1133e3137398046fcddf708b807ddb2d0d6343926767) – the warning threshold has been increased from 999 to 999999.

Yes that was it! Thanks!

Issue #2251, opened 2025-01-27
Switch to Calendar Versioning


uncenter:
### `yazi --debug` output

```Shell
N/A
```

### Please describe the problem you're trying to solve

I think Calendar Versioning (CalVer)](https://calver.org/) would make a lot of sense for a project like Yazi. With Yazi making breaking changes most releases, Semantic Versioning (SemVer) doesn't really make much sense. For example, see Helix, where they release every ~4 months using CalVer: https://github.com/helix-editor/helix/releases - so the versions look like 24.07 and 25.01.

### Would you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

Release the next version using Calendar Versioning.

### Additional context

It could look like (assuming it was released today): 2025.01.27 (YYYY.MM.DD), 2025.01 (YYYY.MM), 25.01 (YY.MM, what Helix uses as mentioned above), 25.01.27 (YY.MM.DD), etc. If a format without the day as the third number (e.g. YY.MM) is used, another digit can be added for subsequent versions in the same month. So versioning might go like 25.01 (one release in January), 25.01.1 (another January release), 25.02 (now a release in February), 25.02.1 (another February release), 25.02.3 (yet another February release), 26.4 (a release in April of the next year), etc...

### Checklist

- [x] I have searched the existing issues/discussions
- [x] The [latest nightly build](https://yazi-rs.github.io/docs/installation/#binaries) doesn't already have this feature

sxyazi:
+1 for this. 

I read about Calendar Versioning and think Yazi fits well with the Ubuntu example it describes:

> Even a simple operating system involves many, many parts, making it difficult to communicate much meaning with an arbitrary number

Yazi consists of several functional submodules (plugin system, `ya` CLI, flavors), and any breaking change within one of them is considered a breaking change for the whole Yazi, even if it only affects that one submodule. Current semantic versioning doesn't describe this situation very well.

I like the idea of pinpointing version numbers down to the day, like `25.01.27`, and for nightly builds, using the build date. This way, plugins can depend on a feature exclusive to nightly versions by specifying the minimum required Yazi version in their metadata. Before running a plugin, Yazi can check for this requirement and notify the user in a more user-friendly way if they need to upgrade.

[...]

PR #2301, opened 2025-02-07
chore: bump version to 25.2.7

sxyazi:


boydaihungst:
Seem like Video preview + preloader doesn't work anymore.

tifandotme:
may I ask why the big jump in versioning?

sxyazi:
Yazi switched to the [Calendar Versioning](https://calver.org/) in https://github.com/sxyazi/yazi/issues/2251

tifandotme:
@sxyazi oh didnt know that, I guess calver would be a nightmare for node packages

Issue #2312, opened 2025-02-10
[Feature Request] Enhance status line configuration


PFiS1737:
### `yazi --debug` output

```Shell
Yazi
    Version: 25.2.7 (Arch Linux 2025-02-09)
    Debug  : false
    Triple : x86_64-unknown-linux-gnu (linux-x86_64)
    Rustc  : 1.84.1 (e71f9a9a 2025-01-27)

Ya
    Version: 25.2.7 (Arch Linux 2025-02-09)

Emulator
    TERM                : Some("xterm-kitty")
    TERM_PROGRAM        : None
    TERM_PROGRAM_VERSION: None
    Brand.from_env      : Some(Kitty)
    Emulator.detect     : Emulator { kind: Left(Kitty), light: false, cell_size: Some((14, 31)) }
    Emulator.detect_full: Ok(Emulator { kind: Left(Kitty), light: false, cell_size: Some((14, 31)) })

Adapter
    Adapter.matches: Kgp

Desktop
    XDG_SESSION_TYPE           : Some("wayland")
    WAYLAND_DISPLAY            : Some("wayland-0")
    DISPLAY                    : Some(":0")
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Neovim
    NVIM          : false
    Neovim version: 0.10.4

Variables
    SHELL           : Some("/bin/fish")
    EDITOR          : Some("nvim")
    VISUAL          : None
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
[...]

PR #2313, opened 2025-02-10
feat: enhance status line configuration.

PFiS1737:
Close: #2312

## Description

- Add configurations for status line separator and format
  - `separator_left  = [ "", "" ]`
  - `separator_right = [ "", "" ]`
  - `format_percent  = [ "Top", "%2d%%", "Bot" ]`
  - `format_position = "%2d/%-2d"`
  - `format_mode     = "%.3s"`

## Checklist:

- [ ] make another PR to update the docs


sxyazi:
Actually, `separator_left` and `separator_right` don't seem to make the `theme.toml` that much more complicated. 

Do you have a screenshot showing what it looks like when different values are set for them?

PFiS1737:
Certainly



### Example

sxyazi:
That looks great, this will be a nice improvement! 

Could you make this PR only include the changes for `separator_left` and `separator_right`?

PFiS1737:
done

yazi-config/preset/theme-dark.toml
```diff
@@ -76,8 +76,8 @@ unset_alt  = { fg = "red", bg = "gray" }
 # : Status bar {{{
 
 [status]
-separator_open  = ""
-separator_close = ""
+separator_left  = [ "", "" ]
```
sxyazi, line 79: Please use a dict instead to make it more readable:

```suggestion
separator_left  = { open = "", close = "" }
```

PFiS1737:
I sincerely request you to reconsider whether the formatting-related features can be merged into the main branch. They are not complex and provide convenience for those who want customization. For others, they can simply remain unchanged.

If possible, I will submit another PR. Those changes don’t seem to be breaking.

PFiS1737:
docs update: yazi-rs/yazi-rs.github.io#203

sxyazi:
All look good to me now, and thanks for updating the documentation!

Sorry, I can't accept formatting-related fields for the following reasons:

1. `%2d%%` is unclear — without looking at the Lua source code, it's hard to understand the specific data represented by `%d`.
2. `%2d%%` is a Lua-specific syntax, and to configure it, the user would need some knowledge of Lua. This somewhat goes against the idea of keeping TOML and Lua separate — TOML, as an easy-to-use front-end, is meant to provide users with a simplified configuration method.

sxyazi (approved):
Thank you!

PFiS1737:
Thanks for your reply, but I’d still like to push for this.

> `%2d%%` is a Lua-specific syntax

[In fact](https://www.lua.org/pil/20.html#:~:text=Or%2C%20better%20yet%2C%20see%20a%20C%20manual%2C%20as%20Lua%20calls%20the%20standard%20C%20libraries%20to%20do%20the%20hard%20work%20here), this is the standard C `printf` syntax (though it does still introduce a connection between TOML and another language, but they are not complex and easy to understand). 

> `%2d%%` is unclear

We can clarify the meaning of placeholders and provide a short explanation of this syntax in the documentation.

Additionally, for users who don’t need these formatting options, just keeping these settings at their default values, and the current appearance remains unchanged.

I understand that you have broader considerations. If this update is ultimately unacceptable, I won’t push further on this and will look for alternative solutions.

PR #2337, opened 2025-02-14
feat: respect the user's `wrap` setting for the built-in JSON previewer

boydaihungst:
Json previewer doesn't wrap even if user set `  wrap = "yes"` in yazi.toml. This PR fix it.

Before:
image

After:
image


sxyazi (approved):
Awesome, thank you!

PR #2349, opened 2025-02-16
fix: incorrect kebab-case for the notify level enum

hankertrix:
For context, refer to:

https://github.com/yazi-rs/yazi-rs.github.io/pull/207

Closes https://github.com/yazi-rs/yazi-rs.github.io/pull/207


sxyazi:
Looks like this is a bug introduced in nightly.

The doc is correct - it should be `error` instead of `Error`, I'll push a new commit to fix it. Thanks for spotting this!

sxyazi (approved):
Thank you!

hankertrix:
You're welcome!

PR #2383, opened 2025-02-22
feat: add ripgrep-all support for search

paperbenni:
ripgrep-all supports all options which are currently used for ripgrep. 
In the future, this could get spun out into  external/rga.rs, but as of now it would be a complete copy of rg.rs with even more duplicated code in the search function. 
I do not know if making use of features which ripgrep-all does not have is at all on any roadmap, so I went for the simpler solution.

### DIFF
--- a/yazi-plugin/src/external/rg.rs
+++ b/yazi-plugin/src/external/rg.rs
@@ -7,13 +7,15 @@ use yazi_shared::url::Url;
 
 pub struct RgOpt {
 	pub cwd:     Url,
+	pub rga:     bool,
 	pub hidden:  bool,
 	pub subject: String,
 	pub args:    Vec<String>,
 }
 
 pub fn rg(opt: RgOpt) -> Result<UnboundedReceiver<File>> {
-	let mut child = Command::new("rg")
+	let commandname = if opt.rga { "rga" } else { "rg" };
+	let mut child = Command::new(commandname)
 		.args(["--color=never", "--files-with-matches", "--smart-case"])
 		.arg(if opt.hidden { "--hidden" } else { "--no-hidden" })
 		.args(opt.args)
--- a/yazi-proxy/src/options/search.rs
+++ b/yazi-proxy/src/options/search.rs
@@ -35,9 +35,9 @@ impl TryFrom<CmdCow> for SearchOpt {
 pub enum SearchOptVia {
 	// TODO: remove `None` in the future
 	None,
-	Rga,
 	Rg,
 	Fd,
+	Rga,
 }
 
 impl From<&str> for SearchOptVia {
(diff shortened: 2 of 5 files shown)


sxyazi:
> ripgrep-all supports all options which are currently used for ripgrep.

I haven't used `ripgrep-all` myself, it's an enhanced version of `ripgrep`, meaning that `ripgrep-all` includes all the features available in `ripgrep` and has roughly the same performance. 

If that's the case, wouldn't it make more sense to switch to `ripgrep-all` so that users can search through a wider range of file types?

paperbenni:
Ripgrep-all is a wrapper for ripgrep. 

It does have the same search performance, but because it can search through more files, it is more expensive to run. Parsing large PDF files can take lots of resources, which users might not be expecting. Rga does have a cache, and considering what it can do I find its performance truly incredible, but by definition it's more expensive. 

I have yet to come across a rg option which rga does not support, but it does not explicitly claim being a superset of ripgrep.

I would not suggest using it as a default, but it would definitely be fine for yazi to treat it the same as ripgrep

yazi-plugin/src/external/rg.rs
```diff
@@ -7,13 +7,15 @@ use yazi_shared::url::Url;
```
sxyazi, line 7: Please copy it as a new `rga.rs`

yazi-proxy/src/options/search.rs
```diff
@@ -35,6 +35,7 @@ impl TryFrom<CmdCow> for SearchOpt {
 pub enum SearchOptVia {
 	// TODO: remove `None` in the future
 	None,
+	Rga,
 	Rg,
 	Fd,
```
sxyazi, line 40:
```suggestion
	Rg,
	Fd,
	Rga,
```

yazi-proxy/src/options/search.rs
```diff
@@ -53,6 +55,7 @@ impl Display for SearchOptVia {
 	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
 		f.write_str(match self {
 			Self::Rg => "rg",
+			Self::Rga => "rga",
 			Self::Fd => "fd",
```
sxyazi, line 59:
```suggestion
			Self::Rg => "rg",
			Self::Fd => "fd",
			Self::Rga => "rga",
```

sxyazi (changes requested):

yazi-plugin/src/external/rg.rs
paperbenni: Did that :)

sxyazi (approved):
Thank you!

Issue #1986, opened 2024-12-04
Theme support for the spotter


GrzegorzKozub:
### `yazi --debug` output

```Shell
Yazi
    Version: 0.3.3 (20dc055 2024-12-02)
    Debug  : false
    Triple : x86_64-unknown-linux-gnu (linux-x86_64)
    Rustc  : 1.83.0 (90b35a62 2024-11-26)

Ya
    Version: 0.3.3 (20dc055 2024-12-02)

Emulator
    Brand.from_env      : Some(Kitty)
    Emulator.detect     : Emulator { kind: Left(Kitty), light: false }
    Emulator.detect_full: Ok(Emulator { kind: Left(Kitty), light: false })

Adapter
    Adapter.matches: Kgp

Desktop
    XDG_SESSION_TYPE           : Some("wayland")
    WAYLAND_DISPLAY            : Some("wayland-0")
    DISPLAY                    : Some(":0")
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Variables
    SHELL           : Some("/bin/zsh")
    EDITOR          : Some("nvim")
    VISUAL          : Some("nvim")
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
[...]

PR #2391, opened 2025-02-23
feat: new `tbl_col` and `tbl_cell` in theme system for spotter table styling

sxyazi:
Follow up to https://github.com/sxyazi/yazi/pull/2389

---

An implementation of the feature request: https://github.com/sxyazi/yazi/issues/1986

Closes https://github.com/sxyazi/yazi/issues/1986

### DIFF
--- a/scripts/validate-form/main.js
+++ b/scripts/validate-form/main.js
@@ -1,4 +1,6 @@
 const LABEL_NAME = "needs info"
+const RE_VERSION = /Yazi\s+Version\s*:\s\d+\.\d+\.\d+\s\(/gm
+const RE_DEPENDENCIES = /Dependencies\s+[/a-z]+\s*:\s/gm
 const RE_CHECKLIST = /#{3}\s+Checklist\s+(?:^-\s+\[x]\s+.+?\n){2}/gm
 
 function bugReportBody(creator, content, hash) {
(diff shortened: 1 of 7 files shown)


uncenter:
These fields names aren't clear to what they actually affect. Using your diagram:

image

`tbl_col` is really *inactive* table cells, and `tbl_cell` is really *active* table cells. Something like `tbl_col` -> `tbl_cell` and `tbl_cell_hovered` (or `tbl_cell_selected`, there isn't any consistency with this kind of naming so far).

uncenter:
The green "header" text on the left above each list of fields also doesn't seem themable?

sxyazi:
> `tbl_col` is really *inactive* table cells, and `tbl_cell` is really *active* table cells.

Didn't quite understand, could you draw a diagram?

uncenter:
Yeah totally.

CleanShot 2025-02-26 at 17 40 17

Boxes in green (and yellow) are what I think are supposed to be cells of the spotter table. They are the boxes you can move up and down in with the arrow keys. The yellow box aroundn the yellow background, at the top, is the currently selected/hovered table cell. If I move the arrow keys up/down, the selected (yellow box) moves. I think all the green boxes - not hovered table cells - should be configured by `tbl_cell`, and the yellow box - hovered table cell - should be configured by `tbl_cell_hovered`/`tbl_cell_selected`.

The red "cells" are what I would also think of as cells, but are not configurable. The green text above each set of red "cells" are also cells presumably, also not configurable.

Ideally I would think of this kind of configuration.

```toml
[spot]
border = { fg = "{{ accent.hex }}" }
title  = { fg = "{{ accent.hex }}" }

[table]
# Red boxes (fields/keys of the table like 'created:' and 'mimetype:').
key_cell = { fg = "white" }
# Green boxes (values of the table like 'video/mp4' or '0:48').
value_cell = { fg = "blue" }
# Yellow box/text (current selected value of the table, the created date value in my diagram).
value_cell_selected = { fg = "yellow", reversed = true }
# Not on diagram, but green text above each set of red boxes (headers like 'video', 'base', and 'plugins').
header_cell = { fg = "green" }
```

uncenter:
@sxyazi any thoughts on my suggestions above?

sxyazi:
`ui.Table` is a Lua binding of [Ratatui's Table](https://docs.rs/ratatui/latest/ratatui/widgets/struct.Table.html), based on the existing API provided by Ratatui, the "selected" in "value_cell_selected" is meaningless, because you always and can only [set a style for a selected cell](https://docs.rs/ratatui/latest/ratatui/widgets/struct.Table.html#method.cell_highlight_style):

> Set the style of the selected cell

Ratatui also lacks an API for styling "cells" as a group - you can only operate on one (and the only) cell, which is the intersection of the current row and column.

If you want to style multiple "cells", you would need to treat them as a whole row or column, and the existing `col_style` is designed to keep consistent with Ratatui's terminology:

https://github.com/sxyazi/yazi/blob/3dc6d82b980be9b974f8123190f8b8f1aaec8ed1/yazi-plugin/preset/plugins/file.lua#L22-L28

Issue #2386, opened 2025-02-23
Permission numbers for Unix, align ui.Span, position of selection marker


csswright:
### `yazi --debug` output

```Shell
N/A
```

### Please describe the problem you're trying to solve

Feature requests (I don't want to spam as separate small feature requests and they are all related, not sure what the dev prefers). [Picture is worth a thousand words](https://0x0.st/8Tfl.png), compares the line mode between above (the desired line mode in nnn file manager) and below, my modified line mode attempt in Yazi:

* Support for [permission numbers for Unix](https://doc.rust-lang.org/beta/std/os/unix/fs/trait.PermissionsExt.html#tymethod.mode). ([perm()](https://yazi-rs.github.io/docs/plugins/types/#shared.cha) is much wordy and IMO less readable than e.g. `755`.

* Allow allowing `ui.Span`--I have line mode left-aligned but its `uiSpan` component for size is variable width so aligning improves readability. 

* Move selection marker position to area between line mode and filename. IMO, line mode only makes sense to the left since its length is much more predictable than the filename (having it on the right affects readability if you have e.g. a short filename and a long filename both in view. With line mode on the left in between filenames and where the selection marker is currently, it's hard to track whether whether the currently hovered file is part of selection compared to if selection marker can be positioned to the left of the filename column. I've tried extending the cursor highlight to the line mode so that it reaches the selection marker, but it's ugly and effects readability of the subdued line mode. 

* Unrelated: provide `yazi --config <path>` as a quick way to switch to yazi defaults (something done often when debugging, thinking about workflows, and determining feature requests). 

### Would you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

See above.

### Additional context

My line mode:

    function Linemode:my_metadata()
      local time = math.floor(self._file.cha.mtime or 0)
      local modified_time

      if time ~= 0 and ya.time() - time <= 300 then
        modified_time = ui.Span(os.date("%Y-%m-%d %H:%M", time)):reverse()
      else
        modified_time = ui.Span(os.date("%Y-%m-%d %H:%M", time))
      end

      local line = ui.Line({
        modified_time,
        "  ",
        self._file.cha:perm(),
        "  ",
        ui.Span(tostring(ya.readable_size(self._file.cha.len))),
      })

      return line:style(ui.Style():fg("#B4B4B4")) -- less white
    end

    local current_render = Current.redraw
    function Current:redraw()
      local t = current_render(self)
[...]

PR #2394, opened 2025-02-24
feat: expose `mode` field in `Cha`

sxyazi:
An implementation of the feature request: https://github.com/sxyazi/yazi/issues/2386

Closes https://github.com/sxyazi/yazi/issues/2386


csswright:
`mode()` just reports the raw st_mode bits which is not user-friendly--should be converted to octal as in the [example](https://doc.rust-lang.org/beta/std/os/unix/fs/trait.PermissionsExt.html#tymethod.mode) and then the last 3 digits yield the [familiar permissions](https://stackoverflow.com/a/35375275) reported/used by common Unix tools.

Issue #2423, opened 2025-03-01
Yazi crashes under SSH after 25.2.26


SpookyYomo:
I do not have in-person access to the machine the bug is occurring on at the moment.
In light of the #2421 plus the triage performed (and explained in the subsequent comments), this might not be an issue with SSH, and just a bad set of commits following from `AsyncStdin` with f25ef0f0.

### What system are you running Yazi on?

Linux Wayland

### What terminal are you running Yazi in?

kitty v0.39.1

### `yazi --debug` output

```Shell
Yazi
    Version: 25.2.26 (57ef8f9e 2025-03-01)
    Debug  : false
    Triple : x86_64-unknown-linux-gnu (linux-x86_64)
    Rustc  : 1.84.1 (e71f9a9a 2025-01-27)

Ya
    Version: 25.2.26 (Arch Linux 2025-02-27)

Emulator
    TERM                : Some("xterm-kitty")
    TERM_PROGRAM        : None
    TERM_PROGRAM_VERSION: None
    Brand.from_env      : Some(Kitty)
    Emulator.detect     : Emulator { kind: Left(Kitty), light: false, cell_size: None }

Adapter
    Adapter.matches: Kgp

Desktop
    XDG_SESSION_TYPE           : Some("tty")
    WAYLAND_DISPLAY            : None
    DISPLAY                    : Some("localhost:15.0")
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: true

WSL
    WSL: false

Variables
    SHELL           : Some("/bin/fish")
    EDITOR          : Some("nvim")
    VISUAL          : None
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
    FZF_DEFAULT_OPTS: None

Text Opener
    default     : Some(Opener { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block-create: Some(Opener { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block-rename: Some(Opener { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })

Multiplexers
    TMUX               : false
[...]

Issue #2421, opened 2025-03-01
Xfce4-terminal fails to start yazi at startup


Crdr0122:
### What system are you running Yazi on?

Linux Wayland

### What terminal are you running Yazi in?

xfce4-terminal 1.1.4

### `yazi --debug` output

```Shell
Terminal response timeout: The request sent by Yazi didn't receive a correct response.
Please check your terminal environment as per: https://yazi-rs.github.io/docs/faq#trt

Yazi
    Version: 25.2.26 (57ef8f9 2025-03-01)
    Debug  : false
    Triple : x86_64-unknown-linux-gnu (linux-x86_64)
    Rustc  : 1.84.0 (9fc6b431 2025-01-07)

Ya
    Version: 25.2.26 (57ef8f9 2025-03-01)

Emulator
    TERM                : Some("xterm-256color")
    TERM_PROGRAM        : None
    TERM_PROGRAM_VERSION: None
    Brand.from_env      : None
    Emulator.detect     : Emulator { kind: Right(Unknown { kgp: false, sixel: false }), light: false, cell_size: None }

Adapter
    Adapter.matches: Chafa

Desktop
    XDG_SESSION_TYPE           : Some("wayland")
    WAYLAND_DISPLAY            : Some("wayland-1")
    DISPLAY                    : Some(":9")
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Variables
    SHELL           : Some("/run/current-system/sw/bin/bash")
    EDITOR          : Some("nvim")
    VISUAL          : None
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
    FZF_DEFAULT_OPTS: None

Text Opener
    default     : Some(Opener { run: "$EDITOR \"$@\"", block: true, orphan: false, desc: "Edit", for_: None, spread: true })
    block-create: Some(Opener { run: "$EDITOR \"$@\"", block: true, orphan: false, desc: "Edit", for_: None, spread: true })
    block-rename: Some(Opener { run: "$EDITOR \"$@\"", block: true, orphan: false, desc: "Edit", for_: None, spread: true })

Multiplexers
    TMUX               : false
    tmux version       : No such file or directory (os error 2)
    tmux build flags   : enable-sixel=Unknown
[...]

PR #2425, opened 2025-03-01
fix: ignore stdin redirection to ensure always accessing the real tty

sxyazi:
Fixes https://github.com/sxyazi/yazi/issues/2421

Fixes https://github.com/sxyazi/yazi/issues/2423

### DIFF
--- a/yazi-adapter/src/stdin.rs
+++ b/yazi-adapter/src/stdin.rs
@@ -1,8 +1,25 @@
-use std::{io::{Error, ErrorKind}, time::{Duration, Instant}};
+use std::{io::{Error, ErrorKind}, ops::Deref, time::{Duration, Instant}};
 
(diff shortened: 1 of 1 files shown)


wylie102:
Can I ask, this is a fix for yazi itself not being redirected and reading/writing to somewhere else? As in the fix isn't supposed to make yazi appear as the TTY when receiving output from another program is it?

I just ask because with my plugin I was having some issues getting duckdb to send it's output as if to the TTY (including ansi codes).  I'm pretty sure the issue is duckdb cleaning its output if it's not directly to the TTY.

I foud a workaround on Mac OS by issuing the command to "script" which then issues the duckdb command and duckdb sees it as the TTY.  I'm working on a solution for linux and windows but I just wanted to check there wasn't something built in that was supposed to be doing this already?

PR #2447, opened 2025-03-06
feat: new "content" option to copy command

Dantescur:
Added the content type handling in the copy command, allowing file content to be copied to the clipboard.

### DIFF
--- a/yazi-config/preset/keymap-default.toml
+++ b/yazi-config/preset/keymap-default.toml
@@ -98,6 +98,7 @@ keymap = [
 	{ on = [ "c", "d" ], run = "copy dirname",          desc = "Copy the directory path" },
 	{ on = [ "c", "f" ], run = "copy filename",         desc = "Copy the filename" },
 	{ on = [ "c", "n" ], run = "copy name_without_ext", desc = "Copy the filename without extension" },
+	{ on = [ "c", "g" ], run = "copy content",         desc = "Copy the content" },
 
 	# Filter
 	{ on = "f", run = "filter --smart", desc = "Filter files" },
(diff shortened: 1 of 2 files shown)


sxyazi:
Thank you for the PR, I have some questions:

- What is the behavior when the user selects multiple files?
- What happens when the user copies a binary file, or a file contains binary content?
- What happens when the user selects a text file along with a binary file?

Dantescur:
> * What is the behavior when the user selects multiple files?

If multiple files are selected it works as expected, copying all file contents with a new line between file contents 

Screenshot_06-Mar_02-35-42_8688



> * What happens when the user copies a binary file, or a file contains binary content?

when the user tries to copy a binary file or a file that contains binary content, the copy command attempts to read the file as text using fs::read_to_string. If the content cannot be interpreted as text (such as with binary files), an error will be caught and logged. 

Screenshot_06-Mar_02-37-09_10992

In this case, the command will return an empty string (Cow::Borrowed(OsStr::new(""))) as a fallback, ensuring that the operation doesn't break.

> * What happens when the user selects a text file along with a binary file?

When the user selects both a text file and a binary file, the copy command will first try to read the text file and copy its content to the clipboard.
When it encounters the binary file, it will return an empty string, as binary content cannot be copied as text.
Both files will still be processed in sequence, with the text content copied and the binary file effectively ignored (in terms of content) and logging the previous error showed.

Hope this help, open to suggestions and thanks for this great project ❤️

Dantescur:
@sxyazi Just following up on this PR—let me know if there's anything that needs improvement! I'm not very familiar with the codebase, so if you have any suggestions or a preferred approach, I'd be happy to adjust the implementation. Thanks for your time!

mikavilpas:
fwiw, here is an 80% implementation I am using in my own config

https://github.com/mikavilpas/dotfiles/blob/7d9379329aaa8c9633ab20bbcc8930462358aa39/.config/yazi/keymap.toml?plain=1#L30-L35

```toml
# keymap.toml
[[mgr.prepend_keymap]]
desc = "Copy contents to the clipboard"
on = ["c", "i"] # copy inside
run = '''
shell 'cat $@ | pbcopy'
'''
```

PR #2485, opened 2025-03-15
feat: file navigation wraparound with new `arrow prev` and `arrow next` commands

sxyazi:
This PR adds new `arrow prev` and `arrow next` commands that provide wraparound navigation at list boundaries and sets them as the default key bindings. 

This decision is based on the following considerations:

- It was requested by multiple users in the past.
- It operates only on a single movement unit (±1), making wraparound navigation more intuitive when continuing to press `j` or `k` at the list boundaries.
- It does not affect the behavior of `arrow 5` (multiple units) or `arrow 50%` (percentage), and the original `arrow -1` and `arrow 1` commands remain available for the previous behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -999,9 +999,9 @@ dependencies = [
 
 [[package]]
 name = "half"
-version = "2.4.1"
+version = "2.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


hankertrix:
This should be marked as a breaking change, as the default configuration for `<Space>` has changed to use wraparound navigation when it previously did not use wraparound navigation.

PR #2494, opened 2025-03-18
feat: allow bulk renaming to include trailing content in addition to the required new names

MikuGeek:
When handling numerous filenames, it's better to give indication to original filenames. I've added the original filenames below, separated by a divider, without changing the code related to renaming.

### DIFF
--- a/yazi-core/src/mgr/commands/bulk_rename.rs
+++ b/yazi-core/src/mgr/commands/bulk_rename.rs
@@ -26,16 +26,12 @@ impl Mgr {
 		tokio::spawn(async move {
 			let tmp = YAZI.preview.tmpfile("bulk");
 			let s = old.iter().map(|o| o.as_os_str()).collect::<Vec<_>>().join(OsStr::new("\n"));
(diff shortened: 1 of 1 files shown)


sxyazi:
This doesn't make sense. Not everyone wants the file names to appear twice, and you can already achieve this through [custom bulk rename commands](https://yazi-rs.github.io/docs/tips#bulk-editor):

```toml
[[opener.bulk-rename]]
block = true
run = '''
  s="$(cat "$1")"
  echo "$s\n------\n$s" > "$1"
  nvim "$1"
  sed -i "" "$(wc -l <<< "$s")q" "$1"
'''

[[open.prepend_rules]]
name = "bulk-rename.txt"
use  = "bulk-rename"
```

However, I can accept removing the strict restriction that the new and old file lists must have the same length, and instead just read the first n lines. 

This way, you wouldn't need the final `sed -i` to remove the separator from the end of the file, which could make implementing such customizations more convenient.

sxyazi (approved):
Thank you!

PR #2525, opened 2025-03-24
feat: add counters to bulk rename function

JustForFun88:
This Pull Request introduces **counter-based bulk renaming** to Yazi, enabling users to specify incrementing placeholders across multiple lines. Each line can include zero or more placeholders of the form:
```
%{<COUNT_TYPE>,<START_VALUE>,<STEP>,<WIDTH>}
```
where:

* `<COUNT_TYPE>`: A single character indicating the counter type:
    - `N`, `n`, `D`, `d` → Numeric digits.
    - `A` → Uppercase ANSI letters.
    - `a` → Lowercase ANSI letters.
    - `R` → Uppercase Roman numerals.
    - `r` → Lowercase Roman numerals.
    - `C` → Uppercase Cyrillic letters.
    - `c` → Lowercase Cyrillic letters.
* `<START_VALUE>` (optional): Initial value, either as a number (e.g., `1`, `2`, etc.) or a value corresponding to the counter type:
    - `A`, `B`, `AA` (ANSI uppercase)
    - `a`, `b`, `aa` (ANSI lowercase)
    - `I`, `II`, `III` (Roman uppercase)
    - `i`, `ii`, `iii` (Roman lowercase)
    - `А`, `Б`, `АБ` (Cyrillic uppercase)
    - `а`, `б`, `аб` (Cyrillic lowercase)
    - `_` for unspecified.
* `<COUNT_STEP>` (optional): Step size, integer (e.g., `2`) or `_` for unspecified.
* `<COUNT_WIDTH>` (optional): Minimum width with zero-padding. Integer (e.g., `3`) or `_` for unspecified.

Optional parameters (`<START_VALUE>`, `<STEP>`, `<WIDTH>`) can be set or left as `_` to take defaults. For instance, `%{n,_,2,3}` starts at the default value 1, increments by 2, and has a minimum width of 3. **They also can be omitted, i.e. not written at all. For example, you can write `%{N}`, `%{N, 1}`**

**_Optional parameters must be specified sequentially:_** `<START_VALUE>` is required if `<COUNT_STEP>` or `<COUNT_WIDTH>` are used, either explicitly (e.g., `1`) or with `_`. Omitting earlier parameters with commas (e.g., `%{N,,2}` or `%{N,,,4}`) is invalid. User should write `%{N,_,2}` or `%{N,_,_,4}` instead.

**Escaping:** To insert a literal `%{` in filenames (in linux for example), user could escape it by writing `%%{`. For example, `file_%%{name}.txt` will become `file_%{name}.txt`.

## Constraints & Error Handling:

- If any line contains counters, all lines must have the same number of placeholders; mismatches result in a comprehensive error report showing all problematic lines simultaneously.
- All errors are collected and reported together.

To avoid adding extra dependencies, the parser is custom-written. While parsing performance could theoretically be improved using Rayon parallelization, I think that typical usage scenarios likely won’t require such optimization.

https://github.com/user-attachments/assets/5c177bc8-89ca-4140-a04c-29a05bf179f9

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4595,6 +4595,7 @@ dependencies = [
  "tokio",
  "tokio-stream",
  "tracing",
+ "unicode-width 0.2.0",
  "yazi-binding",
  "yazi-boot",
  "yazi-config",
--- a/yazi-actor/Cargo.toml
+++ b/yazi-actor/Cargo.toml
@@ -30,16 +30,17 @@ yazi-watcher   = { path = "../yazi-watcher", version = "25.9.15" }
 yazi-widgets   = { path = "../yazi-widgets", version = "25.9.15" }
 
 # External dependencies
-anyhow       = { workspace = true }
-crossterm    = { workspace = true }
-futures      = { workspace = true }
-hashbrown    = { workspace = true }
-mlua         = { workspace = true }
-paste        = { workspace = true }
-scopeguard   = { workspace = true }
-tokio        = { workspace = true }
-tokio-stream = { workspace = true }
-tracing      = { workspace = true }
+anyhow        = { workspace = true }
+crossterm     = { workspace = true }
+futures       = { workspace = true }
+hashbrown     = { workspace = true }
+mlua          = { workspace = true }
+paste         = { workspace = true }
+scopeguard    = { workspace = true }
+tokio         = { workspace = true }
+tokio-stream  = { workspace = true }
+tracing       = { workspace = true }
+unicode-width = { workspace = true }
 
 [target."cfg(unix)".dependencies]
 libc = { workspace = true }
--- a/yazi-actor/src/mgr/bulk_rename.rs
+++ b/yazi-actor/src/mgr/bulk_rename.rs
@@ -19,6 +19,12 @@ use crate::{Actor, Ctx};
 
 pub struct BulkRename;
 
+mod counters;
+mod filename_template;
+mod name_generator;
+
+use name_generator::generate_names;
+
 impl Actor for BulkRename {
 	type Options = VoidOpt;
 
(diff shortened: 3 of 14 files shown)


alexn-s:
great feature. i was looking for such a capability in yazi.


for people waiting, here is my sequential renamer plugin i use (simple and auto names 001, 002, ... with choosen prefix eg img_001.jpg, img_002.jpg, ...)

just place in 
~/.config/yazi/plugins/seq-rename.yazi/main.lua 



```lua
--- Sequential rename: renames selected (or hovered) files to <prefix>_001.ext, …

local collect = ya.sync(function()
	local paths = {}
	for _, url in pairs(cx.active.selected) do
		table.insert(paths, tostring(url))
	end
	if #paths == 0 then
		local h = cx.active.current.hovered
		if h then
			table.insert(paths, tostring(h.url))
		end
	end
	return paths
end)

return {
	entry = function()
		local paths = collect()
		if #paths == 0 then
			return
		end

		local prefix, event = ya.input({
			title = "sequential rename prefix (_001 is appended):",
			value = "file",
			pos = { "top-center", y = 3, w = 40 },
			position = { "top-center", y = 3, w = 40 },
		})
		if event ~= 1 or not prefix or #prefix == 0 then
			return
		end

		local errors = {}
		for i, path in ipairs(paths) do
			local dir = path:match("^(.*)/[^/]*$") or "."
			local base = path:match("[^/]+$") or path
			local ext = base:match("%.[^%.]+$") or ""
			local new_name = string.format("%s_%03d", prefix, i) .. ext
			local new_path = dir .. "/" .. new_name

			local ok, err = os.rename(path, new_path)
			if not ok then
				table.insert(errors, err)
			end
		end

		ya.emit("reload", {})

		if #errors > 0 then
			ya.notify({
				title = "Sequential Rename",
				content = "Some files failed:\n" .. table.concat(errors, "\n"),
				timeout = 5,
				level = "error",
			})
		end
	end,
}

```

or some other folder name and add a hotkey in keymap.toml

```toml
[[mgr.prepend_keymap]]
on = ["R", "n"]
run = "plugin seq-rename"
desc = "Rename: sequential (file_001, file_002, ...)"

```

Issue #2509, opened 2025-03-21
Add a way to set shortcuts for a specific platform


narnaud:
### `yazi --debug` output

```Shell
Yazi
    Version: 25.3.2 (46125ed 2025-03-02)
    Debug  : false
    Triple : x86_64-pc-windows-msvc (windows-x86_64)
    Rustc  : 1.85.0 (4d91de4e 2025-02-17)

Ya
    Version: 25.3.2 (46125ed 2025-03-02)

Emulator
    TERM                : None
    TERM_PROGRAM        : None
    TERM_PROGRAM_VERSION: None
    Brand.from_env      : Some(Microsoft)
    Emulator.detect     : Emulator { kind: Left(Microsoft), light: false, cell_size: Some((10, 20)) }

Adapter
    Adapter.matches: Sixel

Desktop
    XDG_SESSION_TYPE           : None
    WAYLAND_DISPLAY            : None
    DISPLAY                    : None
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Variables
    SHELL           : None
    EDITOR          : None
    VISUAL          : None
    YAZI_FILE_ONE   : Some("C:\\Users\\nicolas\\scoop\\apps\\git\\current\\usr\\bin\\file.exe")
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
    FZF_DEFAULT_OPTS: Some("--ansi")

Text Opener
    default     : Some(Opener { run: "hx %*", block: true, orphan: false, desc: "Edit", for_: None, spread: true })
    block-create: Some(Opener { run: "hx %*", block: true, orphan: false, desc: "Edit", for_: None, spread: true })
    block-rename: Some(Opener { run: "hx %*", block: true, orphan: false, desc: "Edit", for_: None, spread: true })

Multiplexers
    TMUX               : false
    tmux version       : program not found
    tmux build flags   : enable-sixel=Unknown
    ZELLIJ_SESSION_NAME: None
    Zellij version     : program not found

Dependencies
    file          : 5.46
    ueberzugpp    : program not found
    ffmpeg/ffprobe: 7.1.1 / 7.1.1
    pdftoppm      : program not found
    magick        : 7.1.1-46
    fzf           : 0.60.3
    fd/fdfind     : 10.2.0 / program not found
[...]

PR #2526, opened 2025-03-24
feat: platform-specific key binding

sxyazi:
An implementation of the feature request: https://github.com/sxyazi/yazi/issues/2509

Closes https://github.com/sxyazi/yazi/issues/2509

---

This PR adds a new *optional* `for` field to key bindings to specify the platform the key binding applies to.

For example:

```sh
{ on = [ "g", "d" ], run = "cd ~/dev",  desc = "Go dev",    for = "unix" },
{ on = [ "g", "d" ], run = 'cd C:\dev', desc = 'Go C:\dev', for = "windows" },
```

The available values ​​of `for` are the same as https://yazi-rs.github.io/docs/configuration/yazi#opener

### DIFF
--- a/yazi-cli/src/package/hash.rs
+++ b/yazi-cli/src/package/hash.rs
@@ -1,4 +1,4 @@
-use anyhow::{Context, Result};
+use anyhow::{Context, Result, bail};
 use tokio::fs;
 use twox_hash::XxHash3_128;
 use yazi_fs::ok_or_not_found;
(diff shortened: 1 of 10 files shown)


jmlucjav:
does this apply to the prepend_keymap too? I am trying to use it there (a plugin I have is not available in macos, so want I need to prepend the keys only for linux...) and it does not seem to work...

PR #2574, opened 2025-04-05
refactor: deprecate `ui.Padding` and `ui.Rect:padding()`

sxyazi:
This PR deprecates `ui.Padding` in favor of the new `ui.Pad`, which uses the same [parameter order as CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/padding#syntax). This change makes it possible to further simplify syntax in the future:

```diff
- ui.Padding(left, right, top, bottom)
+ ui.Pad(top, right, bottom, left)
```

The `padding()` method of `ui.Rect` is also deprecated to align with the new `ui.Pad` naming:

```diff
- ui.Rect.default:padding(padding)
+ ui.Rect.default:pad(pad)
```

The old `ui.Padding` and `ui.Rect:padding()` are still available, but will trigger a warning.

### DIFF
--- a/yazi-plugin/src/elements/pad.rs
+++ b/yazi-plugin/src/elements/pad.rs
@@ -17,6 +17,13 @@ impl From<ratatui::widgets::Padding> for Pad {
 
 impl Pad {
 	pub fn compose(lua: &Lua, v4: bool) -> mlua::Result<Table> {
+		if !v4 {
+			crate::deprecate!(
+				lua,
+				"The `ui.Padding` has been deprecated, please use `ui.Pad` instead, in your {}"
+			);
+		}
+
 		let new = if v4 {
 			lua.create_function(|_, (_, top, right, bottom, left): (Table, u16, u16, u16, u16)| {
 				Ok(Self(ratatui::widgets::Padding::new(left, right, top, bottom)))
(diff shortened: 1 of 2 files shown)


gauravashisth:
after the latest update, made changes in respective file, ~/.config/yazi/plugins/full-border.yazi/main.lua

previous version:

		local c = self._chunks
		self._chunks = {
			c[1]:padding(ui.Padding.y(1)),
			c[2]:paddding(ui.Padding(c[1].w > 0 and 0 or 1, c[3].w > 0 and 0 or 1, 1, 1)),
			c[3]:padding(ui.Padding.y(1)),
		}

after changes:

		local c = self._chunks
		self._chunks = {
			c[1]:pad(ui.Pad.y(1)),
			c[2]:pad(ui.Pad(c[1].w > 0 and 0 or 1, c[3].w > 0 and 0 or 1, 1, 1)),
			c[3]:pad(ui.Pad.y(1)),
		}

### Changes made:
- padding  ->  pad
- ui.Padding  ->  ui.Pad

**Still can't load yazi, what other changes need to be made to run yazi, please guide.**

sxyazi:
Is the `full-border.yazi` plugin you are using our official one? If so, it is already prepared for the new version of Yazi - you just need to update it to the latest version.

In case you are not using the official plugin, this change was made in https://github.com/yazi-rs/plugins/commit/600614a9dc59a12a63721738498c5541c7923873 , you can refer to it and apply it to your own plugin.

Issue #2575, opened 2025-04-05
Previewing SVG through Magick opens an undesired pop-up window


pierreay:
### What system are you running Yazi on?

Linux X11

### What terminal are you running Yazi in?

Alacritty

### `yazi --debug` output

```Shell
Yazi
    Version: 25.3.2 (Arch Linux 2025-03-05)
    Debug  : false
    Triple : x86_64-unknown-linux-gnu (linux-x86_64)
    Rustc  : 1.85.0 (4d91de4e 2025-02-17)

Ya
    Version: 25.3.2 (Arch Linux 2025-03-05)

Emulator
    TERM                : Some("alacritty")
    TERM_PROGRAM        : None
    TERM_PROGRAM_VERSION: None
    Brand.from_env      : None
    Emulator.detect     : Emulator { kind: Right(Unknown { kgp: false, sixel: false }), light: false, cell_size: None }

Adapter
    Adapter.matches: X11

Desktop
    XDG_SESSION_TYPE           : Some("tty")
    WAYLAND_DISPLAY            : None
    DISPLAY                    : Some(":0")
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Variables
    SHELL           : Some("/usr/bin/zsh")
    EDITOR          : Some("nvim")
    VISUAL          : None
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
    FZF_DEFAULT_OPTS: None

Text Opener
    default     : Some(Opener { run: "nvim \"$@\"", block: true, orphan: false, desc: "Edit (nvim)", for_: None, spread: true })
    block-create: Some(Opener { run: "nvim \"$@\"", block: true, orphan: false, desc: "Edit (nvim)", for_: None, spread: true })
    block-rename: Some(Opener { run: "nvim \"$@\"", block: true, orphan: false, desc: "Edit (nvim)", for_: None, spread: true })

Multiplexers
    TMUX               : false
    tmux version       : tmux 3.5a
    tmux build flags   : enable-sixel=Unknown
    ZELLIJ_SESSION_NAME: None
    Zellij version     : No such file or directory (os error 2)

Dependencies
    file          : 5.46
    ueberzugpp    : 2.9.6
[...]

Issue #2562, opened 2025-04-02
Crash previewing SVG with large dimensions


krir17:
### What system are you running Yazi on?

Linux Wayland

### What terminal are you running Yazi in?

kitty 0.36.4

### `yazi --debug` output

```Shell

Yazi
    Version: 25.3.7 (1765aba 2025-03-30)
    Debug  : false
    Triple : x86_64-unknown-linux-gnu (linux-x86_64)
    Rustc  : 1.85.1 (4eb16125 2025-03-15)

Ya
    Version: 25.3.7 (1765aba 2025-03-30)

Emulator
    TERM                : Some("xterm-256color")
    TERM_PROGRAM        : None
    TERM_PROGRAM_VERSION: None
    Brand.from_env      : None
    Emulator.detect     : Emulator { kind: Right(Unknown { kgp: false, sixel: false }), light: false, cell_size: None }

Adapter
    Adapter.matches    : Wayland
    Dimension.available: WindowSize { rows: 51, columns: 237, width: 0, height: 0 }

Desktop
    XDG_SESSION_TYPE           : Some("wayland")
    WAYLAND_DISPLAY            : Some("wayland-1")
    DISPLAY                    : Some(":0")
    SWAYSOCK                   : Some("/run/user/1000/sway-ipc.1000.458567.sock")
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Variables
    SHELL           : Some("/bin/bash")
    EDITOR          : Some("nvim")
    VISUAL          : Some("nvim")
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
    FZF_DEFAULT_OPTS: None

Text Opener
    default     : Some(OpenerRule { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block-create: Some(OpenerRule { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block-rename: Some(OpenerRule { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })

Multiplexers
    TMUX               : false
    tmux version       : tmux 3.5a
[...]

Issue #2345, opened 2025-02-15
Proc. magick is consuming all available memory.


9a8ri3L:
### What system are you running Yazi on?

Linux Wayland

### What terminal are you running Yazi in?

kitty 0.39.1

### `yazi --debug` output

```Shell
Yazi
    Version: 25.2.11 (22c8f370 2025-02-15)
    Debug  : false
    Triple : x86_64-unknown-linux-gnu (linux-x86_64)
    Rustc  : 1.84.1 (e71f9a9a 2025-01-27)

Ya
    Version: 25.2.11 (22c8f370 2025-02-15)

Emulator
    TERM                : Some("tmux-256color")
    TERM_PROGRAM        : Some("tmux")
    TERM_PROGRAM_VERSION: Some("3.5a")
    Brand.from_env      : Some(Kitty)
    Emulator.detect     : Emulator { kind: Left(Kitty), light: false, cell_size: Some((9, 25)) }

Adapter
    Adapter.matches: Kgp

Desktop
    XDG_SESSION_TYPE           : Some("wayland")
    WAYLAND_DISPLAY            : Some("wayland-0")
    DISPLAY                    : Some(":0")
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Variables
    SHELL           : Some("/usr/bin/zsh")
    EDITOR          : Some("nvim")
    VISUAL          : Some("nvim")
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
    FZF_DEFAULT_OPTS: Some("--height 50% --layout=reverse --bind ctrl-b:preview-up,ctrl-f:preview-down --preview \'bat --style=numbers --color=always --line-range :500 {}\'")

Text Opener
    default     : Some(Opener { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block-create: Some(Opener { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block-rename: Some(Opener { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })

Multiplexers
    TMUX               : true
[...]

PR #2581, opened 2025-04-07
feat!: switch to `resvg` as the SVG renderer

sxyazi:
This PR replaces Yazi's SVG previewer from ImageMagick to `resvg` to address the following issues:

- ImageMagick chooses different backends for SVG rendering based on user build parameters and even the runtime environment (depending on whether a certain backend is installed). It can choose between ImageMagick’s built-in renderer, rsvg, and inkscape, which introduces many uncontrollable factors.
- All these backends, except for rsvg, can exhaust all available memory when rendering certain SVGs, leading to system crashes.
- The parameters required for different backends vary: the built-in backend needs `-density` and `-resize` to maintain image quality, while the rsvg backend only supports `-size`. No single parameter combination works correctly across all backends.

As a result, several users have reported SVG preview issues recently:

- https://github.com/sxyazi/yazi/issues/2345
- https://github.com/sxyazi/yazi/issues/2562
- https://github.com/sxyazi/yazi/issues/2575

Aside from ImageMagick, the remaining two SVG renderers are rsvg-convert and `resvg`. I chose `resvg` because:

- `resvg` is ~2.5 times faster than `rsvg-convert` in my benchmarks
- `resvg` has [better SVG support](https://github.com/linebender/resvg?tab=readme-ov-file#svg-support) than `rsvg-convert`
- `resvg` provides an official Windows binary release, allowing Windows users to use it directly, even install it via Scope

Currently, `resvg` [does not support memory limitation](https://github.com/linebender/resvg/issues/815). So, this PR adds memory restrictions for `resvg` within Yazi:

- On Linux, memory usage is limited with `setrlimit(RLIMIT_AS)`
- On Windows, memory usage is constrained through `CreateJobObjectW()` and `AssignProcessToJobObject()`
- On macOS, where `setrlimit` is ineffective, memory usage is continuously monitored with `proc_pidinfo()`, and the process is terminated if it exceeds the limit

## ⚠️ Breaking change

After this PR, Yazi will use `resvg` instead of `ImageMagick` as the SVG previewer. You can download it from https://github.com/linebender/resvg/releases


sxyazi:
Benchmark with https://commons.wikimedia.org/wiki/File:Cannon-diagram2.svg:

```sh
Benchmark 1: resvg -w 2000 -h 2000 Cannon-diagram2.svg /tmp/abc.png
  Time (mean ± σ):     295.5 ms ±   1.2 ms    [User: 257.2 ms, System: 37.0 ms]
  Range (min … max):   294.5 ms … 298.5 ms    10 runs

Benchmark 2: rsvg-convert -w 2000 -h 2000 Cannon-diagram2.svg -o /tmp/abc.png
  Time (mean ± σ):     747.3 ms ±   6.1 ms    [User: 662.8 ms, System: 59.0 ms]
  Range (min … max):   744.1 ms … 763.9 ms    10 runs

Summary
  resvg -w 2000 -h 2000 Cannon-diagram2.svg /tmp/abc.png ran
    2.53 ± 0.02 times faster than rsvg-convert -w 2000 -h 2000 Cannon-diagram2.svg -o /tmp/abc.png
```

Benchmark with https://commons.wikimedia.org/wiki/File:Moldova_(1483)-ro.svg:

```sh
Benchmark 1: resvg -w 2000 -h 2000 Moldova_\(1483\)-ro.svg /tmp/abc.png
  Time (mean ± σ):     601.6 ms ±   1.7 ms    [User: 545.4 ms, System: 50.0 ms]
  Range (min … max):   598.5 ms … 604.2 ms    10 runs

Benchmark 2: rsvg-convert -w 2000 -h 2000 Moldova_\(1483\)-ro.svg -o /tmp/abc.png
  Time (mean ± σ):      1.466 s ±  0.004 s    [User: 1.365 s, System: 0.061 s]
  Range (min … max):    1.460 s …  1.473 s    10 runs

Summary
  resvg -w 2000 -h 2000 Moldova_\(1483\)-ro.svg /tmp/abc.png ran
    2.44 ± 0.01 times faster than rsvg-convert -w 2000 -h 2000 Moldova_\(1483\)-ro.svg -o /tmp/abc.png
```

JustForFun88:
@sxyazi Aw, too bad 😅! I literally just finished adding resvg support. Plugged it in just like the image crate — and the binary only got about 10% bigger!

But with that we can preview SVGs with text and load fonts for it! Wanna check it out?

sxyazi:
Is there a way to limit its memory usage if it's built-in rather than being
a separate process? Without the limitation, SVGs like the one in
https://github.com/linebender/resvg/issues/814 would immediately consume
all system memory.

JustForFun88:
> Is there a way to limit its memory usage if it's built-in rather than being a separate process? Without the limitation, SVGs like the one in [linebender/resvg#814](https://github.com/linebender/resvg/issues/814) would immediately consume all system memory.

I'm not sure about that — I'll see what I can do. One thing I do know for sure: embedding allows to use about half the memory.

No idea why, but when rendering SVGs, resvg clones the pixel buffer during demultiplying, so we end up with two copies of the pixel data. I changed it to modify the pixels in-place instead. I’ll try to dig into the OOM issue. There's that hint saying “All you can do right now is to walk over nodes and check their bboxes”, though I was kinda hoping to avoid too much headache here.

JustForFun88:
@sxyazi I think that I fixed th OOM in #2588. But the implementation for Ueberzug and Chafa may need further refinement, and I'd appreciate help reviewing or improving those parts.

JustForFun88:
> @sxyazi I think that I fixed the OOM in #2588. But the implementation for Ueberzug and Chafa may need further refinement, and I'd appreciate help reviewing or improving those parts.

PR #2588, opened 2025-04-08
Use resvg as library

JustForFun88:
This PR replaces the external resvg command with the embedded resvg Rust crate, addressing the out-of-memory issue more reliably than the previous setrlimit-based workaround.

Instead of applying OS-level memory limits, this PR calculates the bounding box of all SVG nodes in advance and scales the image down to fit within `max_width` and `max_height`, preventing excessive memory allocation during rendering:

```rust
fn target_size_and_scale(tree: &Tree, max_width: u32, max_height: u32) -> (u32, u32, Transform) {
	// It is Ok. The max_width and max_height could not be larger then monitor
	// detentions which much less then f32::MAX
	let max_width = max_width as f32;
	let max_height = max_height as f32;
	let mut width = tree.size().width();
	let mut height = tree.size().height();

	for node in tree.root().children() {
		if let Some(bounding_box) = node.abs_layer_bounding_box() {
			width = width.max(bounding_box.width());
			height = height.max(bounding_box.height());
		}
	}
	if width <= max_width && height <= max_height {
		return (width.floor() as u32, height.floor() as u32, Transform::from_scale(1.0, 1.0));
	}
	let ratio = f32::min(max_width / width, max_height / height);
	(
		(width * ratio).floor() as u32,
		(height * ratio).floor() as u32,
		Transform::from_scale(ratio, ratio),
	)
}
```
## Build impact summary (Windows)
- Binary size:
   - use_resvg: 21.9 MB (22,964,736 bytes)
   - main: 19.6 MB (20,637,696 bytes)
   - +2.3 MB (~11.3%) increase
- Build time (measured with `cargo build --profile release-windows --locked --timings`):
   - use_resvg: 2m 57.9s (177.9 s)
   - main: 3m 0.9s (180.9 s)
   - Slight improvement (~1.7% faster)
  
## TODO
- [ ] Chech the implementation for Ueberzug
- [ ] Chech the implementation for Chafa


JustForFun88:
@sxyazi I've now implemented the changes as discussed:

- SVGs are integrated into the existing `image_*` APIs. The decoder first attempts to open the image as a raster. If that fails with a format-related error, it then falls back to decoding it as SVG.

- Font-related configuration has been moved to `Lua` for flexibility. The plugin now exposes Lua functions such as `set_svg_font_family`, `load_system_fonts`, `load_font_file`, and `load_fonts_dir`, allowing to modify the resvg font database at runtime.
As noted, resvg won’t parse `<text>` elements unless at least one font is loaded. With the Lua API, users should explicitly load fonts.
- The image scaling logic is left here for now. Some parts of it could potentially be upstreamed later.

The only thing left is to measure build time and binary size impact — I’ll post that shortly.

Issue #2598, opened 2025-04-10
Large size images


sibi-venti:
### What system are you running Yazi on?

Linux X11

### What terminal are you running Yazi in?

ghossty and gnome-terminal

### `yazi --debug` output

```Shell
runtime error
```

### Describe the bug

when the directory have large images it tries to preview and fills up the RAM and pc dies or Hangs
image size is 40 mb or 4k image 8k ,
you can have something to turn off for high res images

### Minimal reproducer

open yazi with high res images

### Checklist

- [x] I tried the [latest nightly build](https://yazi-rs.github.io/docs/installation#binaries), and the issue is still reproducible
- [x] I updated the debug information (`yazi --debug`) input box to the nightly that I tried
- [x] I can reproduce it after disabling all custom configs/plugins (`mv ~/.config/yazi ~/.config/yazi-backup`)

sxyazi:
Seems that the `magick` previewer currently only passes `MAGICK_MEMORY_LIMIT` (corresponding to `image_alloc`):

https://github.com/sxyazi/yazi/blob/7cc250180f61e0e4e8a5d0aa0271d31c3fc797b7/yazi-plugin/preset/plugins/magick.lua#L51-L53

I found `MAGICK_WIDTH_LIMIT` and `MAGICK_HEIGHT_LIMIT` in the [ImageMagick documentation](https://imagemagick.org/script/resources.php), which correspond to `image_bound`, maybe we can pass them to ImageMagick as well

sxyazi:
Libvips looks promising, nice find! @ze0987

Perhaps we could explore replacing ImageMagick with libvips, but need to
look into further its cross-platform support and image size/quality ratio -
would you like to file a new issue to track it?

PR #2602, opened 2025-04-10
feat: limit memory usage for previewing large images

ze0987:
~Fixes https://github.com/sxyazi/yazi/issues/2598~ This limits the memory available to `magick`, but regardless, `magick` tries to scale the image anyway and fails. This is very wasteful, therefore it is not a solution.


sxyazi:
I think this still makes sense, at the very least, it will not result in ImageMagick exhausting the memory and crashing the system, and it can serve as a temporary solution until we find a better way to limit ImageMagick's memory usage, so let's reopen it.

sxyazi (approved):
Thank you!

PR #2609, opened 2025-04-12
feat: allow custom exit code with `quit --code`

Nicholas42:
Hi! Very cool project, I recently started using it and have a (not too obscure in my opinion) use case: When integrating yazi with other tools, it is useful to be able to return with a non-zero exit code. An example for this is the pager for https://github.com/aurutils/aurutils 

I don't think it is possible to do this via a plugin (except for maybe somehow purposefully panicking from lua), so I added an option to the quit command.

Hope this is helpful.

### DIFF
--- a/yazi-core/src/mgr/commands/close.rs
+++ b/yazi-core/src/mgr/commands/close.rs
@@ -1,16 +1,24 @@
-use yazi_shared::event::CmdCow;
+use yazi_shared::event::{CmdCow, Data};
 
 use crate::{mgr::{Mgr, commands::quit}, tasks::Tasks};
 
 #[derive(Default)]
 struct Opt {
 	no_cwd_file: bool,
+	exit_code:   i32,
 }
 impl From<CmdCow> for Opt {
-	fn from(c: CmdCow) -> Self { Self { no_cwd_file: c.bool("no-cwd-file") } }
+	fn from(c: CmdCow) -> Self {
+		Self {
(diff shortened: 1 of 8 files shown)


sxyazi:
It's unclear to me why this is needed: 

- Does aurutils specifically require it — what is it used for? 
- You added it to the `quit` command, do you intend to bind different keys to `quit` in order to exit Yazi with different exit codes?
- Does a command line parameter like `yazi --exit-code` make sense?

Nicholas42:
> * Does aurutils specifically require it — what is it used for?

The workflow in aurutils is the following: It opens a directory (the build directory for an aur package) with an external program (aurutils calls it a "pager", though I find this a bit of a weird name). The user inspects the directory (peeking into files and such) and then returns from the external program, choosing an exit code. With zero exit code, aurutils proceeds, with non-zero exit code it aborts.

To be honest, I haven't seen this workflow much outside of aurutils, but I think it is actually a very general approach that makes a lot of sense to me.

> * You added it to the `quit` command, do you intend to bind different keys to `quit` in order to exit Yazi with different exit codes?

correct

> * Does a command line parameter like `yazi --exit-code` make sense?

I don't know. Personally, I have no immediate use case for determining the exit code of yazi before I open it (and could emulate that with `yazi && return <EXIT_CODE>`

sxyazi (approved):
Thank you!

PR #2657, opened 2025-04-21
feat: make the hover state for `reveal`, `sort`, and `hidden` commands stable

sxyazi:
Resolves https://github.com/sxyazi/yazi/discussions/2348


boydaihungst:
The default `sort_dir_first = true` setting is not working due to a regression introduced in this PR:
image

sxyazi:
Fixed in https://github.com/sxyazi/yazi/commit/a29dbb9f068beda29a1ef5424d905fca7808bbfc

PR #2676, opened 2025-04-25
feat(json_decode): add optional 'ordered' argument for ordered tables

xiaoxigua-1:
### What this PR does

Adds an optional `ordered` argument to the Yazi plugin’s `json_decode` function.  
When `ordered = true`, the decoded JSON is wrapped in an `OrderedTable`, preserving key order during Lua `pairs()` iteration.

### Why this is useful

Lua tables do not guarantee key order.  
This change allows users to retain JSON key ordering, which is useful for:

- UI rendering  
- Config parsing  
- Other order-sensitive tasks

### To do

- [x] Add `ordered` argument to `json_decode`
- [x] Wrap decoded JSON in `OrderedTable` when ordered is true
- [x] Implement `__newindex` in `OrderedTable` (currently read-only)
- [x] Fix encoding of `OrderedTable`: 
    - Ensure only `data` is encoded (exclude `ordered`)
    - Maintain key order during encoding

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3474,6 +3474,7 @@ dependencies = [
  "parking_lot",
  "paste",
  "ratatui",
+ "serde",
  "serde_json",
  "syntect",
  "tokio",
--- a/yazi-plugin/Cargo.toml
+++ b/yazi-plugin/Cargo.toml
@@ -43,6 +43,7 @@ tokio-util    = { workspace = true }
 tracing       = { workspace = true }
 twox-hash     = { workspace = true }
 unicode-width = { workspace = true }
+serde        = { workspace = true }
 yazi-prebuilt = "0.1.0"
 
 [target."cfg(unix)".dependencies]
--- a/yazi-plugin/src/utils/json.rs
+++ b/yazi-plugin/src/utils/json.rs
@@ -1,126 +1,240 @@
-use mlua::{Function, IntoLuaMulti, Lua, LuaSerdeExt, MetaMethod, Result, Table, Value};
-use yazi_binding::Error;
+use mlua::{
+	Function, IntoLuaMulti, Lua, LuaSerdeExt, MetaMethod, UserData, UserDataMethods, Value,
+};
 
-use std::cell::RefCell;
+use serde::Serialize;
+use serde_json::value::Value as JsonValue;
+
+use std::cell::{Ref, RefCell, RefMut};
 use std::rc::Rc;
+
+use yazi_binding::Error;
 
 use super::Utils;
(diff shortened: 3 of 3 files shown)


xiaoxigua-1:
Perhaps using `UserData` to wrap the `OrderedTable` would be a more efficient solution.

xiaoxigua-1:
The use of RefCell in async code may cause race conditions; consider using Arc<Mutex<T>> for safe shared state.

sxyazi:
I'm sorry, but I'm afraid I cannot accept this PR as its implementation is too complicated, and relying on the order of keys is not best practice and Yazi has always tried to avoid such designs. Is there any real-world use case that requires it at all?

xiaoxigua-1:
> I'm sorry, but I'm afraid I cannot accept this PR as its implementation is too complicated, and relying on the order of keys is not best practice and Yazi has always tried to avoid such designs. Is there any real-world use case that requires it at all?

JSON is inherently an ordered data format, but due to Lua tables being unordered and lacking order-preserving capabilities, it's not possible to reliably retrieve the original key order in Lua. This can lead to issues in plugins, especially when rendering UI elements where the order of fields matters — potentially resulting in a randomly ordered display.

xiaoxigua-1:
Maybe I can just return the order directly, instead of implementing a full OrderedTable — no need to make things unnecessarily complicated.

sxyazi:
> JSON is inherently an ordered data format

No, it isn't. From [RFC 7159](https://www.rfc-editor.org/rfc/rfc7159.html#section-4):

>   JSON parsing libraries have been observed to differ as to whether or
>   not they make the ordering of object members visible to calling
>   software.  Implementations whose behavior does not depend on member
>   ordering will be interoperable in the sense that they will not be
>   affected by these differences.

This means that the order of keys depends on the specific implementation, and Yazi *can* choose not to preserve their order.

> especially when rendering UI elements where the order of fields matters

Can you elaborate on this?

xiaoxigua-1:
> No, it isn't. From [RFC 7159](https://www.rfc-editor.org/rfc/rfc7159.html#section-4):

You're right — JSON is not inherently ordered. As noted in RFC 7159

This means key order is implementation-dependent, and Yazi is free to ignore it.
> Can you elaborate on this?

That said, my intent behind preserving key order is purely for UI rendering purposes — especially in cases like JSON previews or metadata panels, where a logical and predictable field order greatly improves readability. For instance, fields like name, type, size, and modified are easier to scan when consistently ordered. So even though ordering isn’t required by the spec, maintaining it can enhance the user experience in these contexts.

PR #2678, opened 2025-04-26
feat: add `s` and `S` keybinds in the input component

hankertrix:
Also changed the behaviour of "cc" to match Neovim's behaviour.

### DIFF
--- a/yazi-config/preset/keymap-default.toml
+++ b/yazi-config/preset/keymap-default.toml
@@ -286,16 +286,16 @@ keymap = [
 	{ on = "<A-d>", run = "kill forward",  desc = "Kill forwards to the end of the current word" },
 
 	# Cut/Yank/Paste
-	{ on = "d", run = "delete --cut",                                     desc = "Cut selected characters" },
-	{ on = "D", run = [ "delete --cut", "move eol" ],                     desc = "Cut until EOL" },
-	{ on = "c", run = "delete --cut --insert",                            desc = "Cut selected characters, and enter insert mode" },
-	{ on = "C", run = [ "delete --cut --insert", "move eol" ],            desc = "Cut until EOL, and enter insert mode" },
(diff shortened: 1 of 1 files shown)


hankertrix:
@sxyazi any updates on this PR?

yazi-widgets/src/input/commands/delete.rs
```diff
@@ -25,7 +25,7 @@ impl Input {
 				self.move_(0);
 			}
 			InputOp::Delete(..) => {
-				self.snap_mut().op = InputOp::Delete(opt.cut, opt.insert, 0);
+				self.snap_mut().op = InputOp::Delete(opt.cut, opt.insert, self.snap().value.chars().enumerate().find(|(_, c)| !c.is_whitespace()).map_or(0, |(i, _)| i));
```
sxyazi, line 28: Can you explain this change a little bit please?

sxyazi: Thanks for explaining!

I'm not sure if it change makes in the filename editing context - `autoindent` applies to the current line and the new line, that is, to a multi-line editing case:

```
'autoindent' 'ai'    boolean   (default on)
            local to buffer
    Copy indent from current line when starting a new line (typing <CR>
    in Insert mode or when using the "o" or "O" command).  If you do not
```

But Yazi's input is only a single line, so there is no indentation reference line.

Consider: if the filename begins with spaces, and you press `cc`, do you want to delete the entire line or preserve the leading space? Why?

hankertrix: I see. Personally, I don't think it makes sense in the file editing context either, but the change was more of just to match Neovim's behaviour.

I'll edit the PR to change it back to the original behaviour, and edit the `S` keybind to match as well.

sxyazi (approved):
Thank you!

hankertrix:
You’re welcome!

Issue #1403, opened 2024-08-03
Make tabs clickable


Jelenkee:
### `yazi --debug` output

<details>

```sh
##### ↓↓↓ Paste the output here: ↓↓↓ #####


```

</details>

### Please describe the problem you're trying to solve

You cannot change the current tab by clicking on one of them.
I already tried to figure out how I can fix it on my own. But I don't know how to determine the x position of the tabs in lua

### Would you be willing to contribute this feature?

- [X] Yes, I'll give it a shot

### Describe the solution you'd like

Make tabs clickable

### Validations

- [X] I have searched the existing issues
- [X] The latest nightly build of Yazi doesn't already have this feature

Ali1997bhatti:
https://github.com/sxyazi/yazi/issues/1403

PR #2702, opened 2025-04-30
feat: make tabs in header clickable

coolkiid:
according to the x-offset, calculate in which tab the mouse clicks.

issue: #1403

### DIFF
--- a/yazi-plugin/preset/components/header.lua
+++ b/yazi-plugin/preset/components/header.lua
@@ -110,7 +110,36 @@ function Header:redraw()
 end
 
 -- Mouse events
-function Header:click(event, up) end
+function Header:click(event, up)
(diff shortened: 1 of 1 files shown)


coolkiid:
I've tested this patch on a Mac machine (zsh, iTerm2, Sonoma 14.7.4) and a Windows machine (powershell, windows-terminal, Windows 10).

coolkiid:
@sxyazi ptal <3

sxyazi:
Hi, thanks for the PR! I'm currently planning to redesign the tabs as a
standalone, single-line component with better defaults, for example,
showing the tab name by default, adaptive truncation for long names,
clickable support, etc. This PR will be superseded once the new
implementation landed.

PR #2709, opened 2025-05-02
feat: add `--hovered` to the `copy` command

hankertrix:
Similar to the `open`, `rename` and `remove` commands, the `--hovered` flag will cause the copy command to operate only on the hovered file.


sxyazi (approved):
Thank you!

hankertrix:
You're welcome!

PR #2745, opened 2025-05-11
feat!: redesign tabs

sxyazi:
This PR decouples tabs from the header into an independent tab bar component to enhance UX:

- Standalone single-line component to increase the available space for displaying tabs
- Tab titles are displayed by default to increase recognizability
- Supports switching tabs via mouse for easier navigation

The new tab bar is only shown when there is more than one tab, ensuring no space waste for users who don't use tabs.

Closes https://github.com/sxyazi/yazi/issues/1403

Closes https://github.com/sxyazi/yazi/pull/2702

## Demo

```toml
# ~/.config/yazi/theme.toml
[tabs]
active   = { fg = "#24273a", bg = "#8aadf4", bold = true }
inactive = { fg = "#8aadf4", bg = "#363a4f" }
```

https://github.com/user-attachments/assets/5ac79095-9a4a-444e-894a-e6ee0e992815

## ⚠️ Breaking changes

- `tab_active` and `tab_inactive` have been moved to the new `[tabs]` section as `active` and `inactive` from `[manager]`
- `tab_width` under `[manager]` has been removed because the new implementation automatically calculates and truncates tab titles, no need for manual set widths

### DIFF
--- a/yazi-config/preset/theme-dark.toml
+++ b/yazi-config/preset/theme-dark.toml
@@ -38,11 +38,6 @@ marker_cut      = { fg = "lightred",    bg = "lightred" }
 marker_marked   = { fg = "lightcyan",   bg = "lightcyan" }
 marker_selected = { fg = "lightyellow", bg = "lightyellow" }
 
-# Tab
-tab_active   = { reversed = true }
-tab_inactive = {}
-tab_width    = 1
-
 # Count
 count_copied   = { fg = "white", bg = "green" }
 count_cut      = { fg = "white", bg = "red" }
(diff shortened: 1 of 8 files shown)


GOLD875379:
Hey, I updated my Yazi and noticed the change that was made by this PR. I didn't like the new UI or the way the TABs work at all. How can I go back to normal or roll back this specific feature without having to stay on an outdated build?

sxyazi:
see https://github.com/sxyazi/yazi/pull/2782 @GOLD875379

GOLD875379:
@sxyazi Thx, worked

Ticonderoga:
It looks really nice but is it possible to put the tabs on the right side ? something like

`
[tabs]
position  = right
`

Thanks for your work

somnico:
Awesome work. Was wondering though if there is a setting for removing the one wide gap that shows the background that's present to the left of the first tab and to the right of the last tab?

PR #2754, opened 2025-05-13
feat: truncate long items in the file list

sxyazi:
An implementation of the feature request: https://github.com/sxyazi/yazi/issues/903

Resolves https://github.com/sxyazi/yazi/issues/903

### DIFF
--- a/yazi-plugin/preset/components/current.lua
+++ b/yazi-plugin/preset/components/current.lua
@@ -32,15 +32,16 @@ function Current:redraw()
 		return self:empty()
 	end
 
-	local entities, linemodes = {}, {}
+	local left, right = {}, {}
 	for _, f in ipairs(files) do
-		entities[#entities + 1] = Entity:new(f):redraw()
-		linemodes[#linemodes + 1] = Linemode:new(f):redraw()
+		left[#left + 1] = Entity:new(f):redraw()
+		right[#right + 1] = Linemode:new(f):redraw()
+		left[#left]:truncate { max = math.max(0, self._area.w - right[#right]:width()) }
 	end
 
 	return {
-		ui.List(entities):area(self._area),
-		ui.Text(linemodes):area(self._area):align(ui.Text.RIGHT),
+		ui.List(left):area(self._area),
+		ui.Text(right):area(self._area):align(ui.Text.RIGHT),
 	}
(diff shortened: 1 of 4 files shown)


PoisonFrog:
So the extension won’t be shown? I don’t see the difference between a file name being cut off because it doesn’t fit in the pane and one ending with "..." since the extension is hidden either way. Also, why waste three characters with "..." when a single character like "~" could do the job more efficiently?

sxyazi:
No, it's `…` (width is 1, U+2026 HORIZONTAL ELLIPSIS), not `...` (width is
3).

The former is a symbol specifically defined by the Unicode standard for
ellipsis, and CSS also uses it,
https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow#ellipsis

PoisonFrog:
> No, it's `…` (width is 1, U+2026 HORIZONTAL ELLIPSIS), not `...` (width is 3). The former is a symbol specifically defined by the Unicode standard for ellipsis, and CSS also uses it, https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow#ellipsis

Oh, I see. That's much better.

> For example, if the user has enabled [show_symlink](https://yazi-rs.github.io/docs/configuration/yazi/#manager.show_symlink), then for symlinks, " -> /path/linked/to" will be appended after the filename, in which it's hard to determine how long the filename part should occupy

Is there any other case where this would be an issue? If `show_symlink` is the only setting that would conflict with showing the extension we could have it show extensions when `show_symlink` is set to `false` and not show extensions when it is set to `true`.

sxyazi:
Yes, there are, `show_symlink` is just one of the cases built-in supported
by Yazi, but users can freely add any content before or after the filename
through the Lua API:

https://github.com/sxyazi/yazi/blob/5a4e6e40684751e600f927aa9ed32e8baabe66f9/yazi-plugin/preset/components/entity.lua#L3-L9

So it's uncertain how much space should be allocated for the filename part.

The current implementation truncates the entire part of the line that
doesn't include `linemode`, so all these cases work correctly.

sxyazi:
@PoisonFrog `"…file extension"` is now supported in https://github.com/sxyazi/yazi/pull/2759

PR #2777, opened 2025-05-19
feat: add a `/` to the interactive cd auto-completion candidates

ineu:
To be able to jump directly to `/` via `g SPACE`

Resolves #2760


sxyazi (approved):
Thank you!

PR #2803, opened 2025-05-28
refactor!: deprecate `[manager]` in favor of `[mgr]`

sxyazi:
Yazi's design philosophy is to use the same terminology for the same things to avoid confusion. 

This PR deprecates `[manager]` and introduces the new `[mgr]` as its direct replacement, ensuring consistency with the others:

- `rt.mgr` - [user preferences API](https://yazi-rs.github.io/docs/plugins/runtime#rt.mgr)
- `th.mgr` - [theme configuration API](https://yazi-rs.github.io/docs/plugins/runtime#th.mgr)
- `ya.emit("mgr:...")` - [Lua API](https://yazi-rs.github.io/docs/plugins/utils#ya.emit)
- `ya emit mgr:...` - `ya` subcommand
- `mgr:...` - [cross-layer command invocation](https://github.com/sxyazi/yazi/pull/2399)

## Deprecation

```diff
# yazi.toml / keymap.toml / theme.toml / flavor.toml
- [manager]
+ [mgr]
```

This PR includes an automatic migration for `yazi.toml`, `keymap.toml`, and `theme.toml`

When Yazi starts up, it will check if the user's configuration contains `[manager]`, if it does, backup the existing configuration and update the user's configuration with the new `[mgr]` to ensure a smoother transition to the new `[mgr]`.

Note that `flavor.toml` will not be automatically migrated, as it's not a user-owned file. If you are using a flavor from [`yazi-rs/flavors`](https://github.com/yazi-rs/flavors), run `ya pkg upgrade` to update to the latest commit for support.

### DIFF
--- a/yazi-config/src/lib.rs
+++ b/yazi-config/src/lib.rs
@@ -112,7 +112,7 @@ fn migrate(p: std::path::PathBuf) -> String {
 	if let Err(e) = std::fs::copy(&p, backup) {
 		_ = TTY.writer().write_all(
 			format!("WARNING: `[manager]` has been deprecated in favor of the new `[mgr]`, see #2803 for more details: https://github.com/sxyazi/yazi/pull/2803\n
-Trying to migrate your config automatically failed, please rename the file manually, error while backuping {p:?}: \n{e:?}\n").as_bytes(),
+Trying to migrate your config automatically failed, please edit the file manually, error while backuping {p:?}: \n{e:?}\n").as_bytes(),
 		);
 		return new;
 	}
@@ -120,7 +120,7 @@ Trying to migrate your config automatically failed, please rename the file manua
 	if let Err(e) = std::fs::write(&p, &new) {
 		_ = TTY.writer().write_all(
 			format!("WARNING: `[manager]` has been deprecated in favor of the new `[mgr]`, see #2803 for more details: https://github.com/sxyazi/yazi/pull/2803\n
-Trying to migrate your config automatically failed, please rename the file manually, error while writing {p:?}: \n{e:?}\n").as_bytes(),
+Trying to migrate your config automatically failed, please edit the file manually, error while writing {p:?}: \n{e:?}\n").as_bytes(),
 		);
 	}
 	new


uncenter:
This makes the configuration file far less readable. Why are we opting for shorter, more ambiguous wording?

sxyazi:
Shortening "manager" to "mgr" cuts down on a few keystrokes, I hope it boosts the QoL, especially since it's the term used most frequently across all layers.

Ideally, we should balance practicality (keeping it short) with correctness (using the complete word), and I've always tried to do that - naming things when designing a feature takes a lot of effort, but sometimes there's an irreconcilable conflict between the two and when it's hard to find a short, complete word, based on my past patterns, I tend to lean more towards practicality.

uncenter:
I'd rather `[yazi-file-manager-manager-configuration-section]` than `[mgr]`. Extreme brevity should not be the goal while naming configuration options. I, and others - judging by the reactions to my previous comment - feel that the UX is actually worsened, not improved, by this change. A major and breaking change like this might serve well as an RFC to get community feedback first.

uncenter:
I'd propose moving the `manager` section options to the top-level of configuration (no specific section) or calling the section something like "[general]" -- though the latter is no more "optimal" in brevity than the original "[manager]" as they are both 7 character words.

sxyazi:
> moving the `manager` section options to the top-level of configuration (no specific section)

How should I access them with `rt` and `th`? Currently, I use `rt.mgr.sort_by` to access:

https://github.com/sxyazi/yazi/blob/5cb6e1f2245dd876838fa6de9c463d7b8cfc80d3/yazi-config/preset/yazi-default.toml#L7

Does that mean, after this change, it becomes `rt.sort_by`? That sounds odd, because it would mean mixing the values from `yazi.toml` with others that might not even belong to `yazi.toml` (like `rt.args`, `rt.term`) in the same level:

https://github.com/sxyazi/yazi/blob/5cb6e1f2245dd876838fa6de9c463d7b8cfc80d3/yazi-plugin/src/config/runtime.rs#L14-L23


> calling the section something like "[general]" -- though the latter is no more "optimal" in brevity than the original "[manager]" as they are both 7 character words.

No, `[manager]` is not for general configuration, it's the configuration for the `[manager]` layer, which is used to display the 3-column layout. For general configuration, I plan to introduce a new `[app]` in the future, and pull out some of the existing configs of `[manager]` into it, such as `mouse_events` and `title_format` - it already exists, just hasn't been used yet:

https://github.com/sxyazi/yazi/blob/5cb6e1f2245dd876838fa6de9c463d7b8cfc80d3/yazi-shared/src/layer.rs#L7-L10

Besides, it isn't any shorter than the original `[manager]`, which would make code like:

https://github.com/sxyazi/yazi/blob/5cb6e1f2245dd876838fa6de9c463d7b8cfc80d3/yazi-plugin/preset/components/rail.lua#L13-L14

even longer.

uncenter:
> Besides, it isn't any shorter than the original `[manager]`, which would make code like:

Right lol, my point is that there is no need to make that code shorter. It was fine as it was and would continue to be fine.

sxyazi:
That would make it inconsistent with all other APIs listed in the OP, and I really want to avoid that as much as possible.

Also, using abbreviations in Yazi is a very common practice/coding style — not just a special case. A quick look at the codebase will find many such abbreviations: `app`, `mgr`, `cmp`, `url`, `cha`, `rect`, `pad`, `pos`, `cx`, `rt`, `th`, `fs`, `ui`, `ya`, `ps`.

As I said in https://github.com/sxyazi/yazi/pull/2803#issuecomment-2917169860, when designing Yazi, I aimed to find a balance between correctness and practicality. I don't know if you are using these APIs listed in OP, but I use them A LOT. 

Using `mgr` instead of `manager` not only reduces the keystrokes but also makes it easier for me to spot errors (`"mgr"` vs `"manager"`, the former is recognizable to me immediately). Also, it shortens the codeline, making reviews easier - I can read less code and focus more on other more critical parts.

PR #2807, opened 2025-05-28
feat: support `.deb` packages

termit-uanic:
Added packaging of already built binary files into a valid deb package.

Tested on Ubuntu 24.04

I haven't had a chance to test on different Ubuntu/Debian versions, but as far as I can tell, resulting deb file doesn't have any distribution and version-specific features. So it should work everywhere.

Besides bin files the package installs completion for bash, README.md, LICENSE all dependencies specified in the documentation

recommended way to install the deb package:
```
sudo apt install ./yazi-fm_<version>.deb
```

list of files:
```
❯ dpkg-query -L yazi-fm
/usr
/usr/share
/usr/share/bash-completion
/usr/share/bash-completion/completions
/usr/share/bash-completion/completions/ya
/usr/share/bash-completion/completions/yazi
/usr/share/doc
/usr/share/doc/yazi-fm
/usr/share/doc/yazi-fm/README
/usr/share/doc/yazi-fm/copyright
/usr/bin
/usr/bin/ya
/usr/bin/yazi
```

### DIFF
--- a/.github/workflows/draft.yml
+++ b/.github/workflows/draft.yml
@@ -57,8 +57,10 @@ jobs:
       - name: Upload artifact
         uses: actions/upload-artifact@v4
         with:
-          name: yazi-${{ matrix.target }}.zip
-          path: yazi-${{ matrix.target }}.zip
+          name: ${{ matrix.target }}
+          path: |
+            yazi-${{ matrix.target }}.zip
+            yazi-${{ matrix.target }}.deb
 
   build-windows:
     strategy:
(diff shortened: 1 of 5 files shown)


sxyazi:
Thanks for the patch!

Is it possible to pull out the CLI-related items from `yazi-fm/Cargo.toml` into their dedicated config file, i.e. `yazi-cli/Cargo.toml`?

termit-uanic:
Hi
Yes, that's possible, but then we'd get two deb packages.
These packages could be linked together.
But it would be inconvenient to download two linked deb files.

PR #2825, opened 2025-06-02
chore(nix): use ffmpeg instead of ffmpegthumbnailer (#1928)

chinrw:

Now yazi is using ffmpeg instead of ffmpegthumbnailer, and nix system hasn't updated its deps.

Resolves #

Replace ffmpegthumbnailer with ffmpeg for generating previews, bringing the Nix build in line with changes introduced in commit 0f9a3195 (feat!: video spotter & use ffmpeg as previewer backend).

### DIFF
--- a/nix/yazi.nix
+++ b/nix/yazi.nix
@@ -14,7 +14,7 @@
   jq,
   poppler_utils,
   _7zz,
-  ffmpegthumbnailer,
+  ffmpeg,
   fd,
   ripgrep,
   fzf,
@@ -40,7 +40,7 @@ let
     jq
     poppler_utils
     _7zz
-    ffmpegthumbnailer
+    ffmpeg
     fd
     ripgrep
     fzf


sxyazi:
Thanks for the patch!

While you're on this, could you also add [resvg](https://github.com/linebender/resvg) to the dependency list, which was introduced in https://github.com/sxyazi/yazi/pull/2581

chinrw:
> Thanks for the patch!
> 
> While you're on this, could you also add [resvg](https://github.com/linebender/resvg) to the dependency list, which was introduced in #2581

Updated, thanks for this great tool :)

sxyazi:
Thanks for the update, can you also add `resvg` to the `defaultDeps` array?

chinrw:
> Thanks for the update, can you also add `resvg` to the `defaultDeps` array?

Updated

sxyazi (approved):
LGTM, thank you!

Issue #2836, opened 2025-06-04
Allow updating subset of packages


csswright:
### `yazi --debug` output

```Shell
N/A
```

### Please describe the problem you're trying to solve

Currently, it is only possible to update *all* packages. For users with many packages, updating all at once makes it difficult to debug potential issues or incorporate new features into workflow. Manually pinning/unpinning subsets of packages is not convenient and updating only specified packages fits in line with a typical package manager.

### Would you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

Allow updating subset of packages. 

### Additional context

Considering that this is a simple feature request, throwing an idea for one more useful and related feature request: show updates of packages in the form of new commits messages for each package updated. This is convenient to give an idea of what changes can be expected from the update without manually checking themselves in e.g. a web browser. This is a common feature in similar package managers in other applications like vim and system package managers.

### Checklist

- [x] I have searched the existing issues/discussions
- [x] The [latest nightly build](https://yazi-rs.github.io/docs/installation/#binaries) doesn't already have this feature

sxyazi:
> Allow updating subset of packages.

Supported in https://github.com/sxyazi/yazi/pull/2841

> show updates of packages in the form of new commits messages for each package updated

There's no plan to support it because it doesn't integrate well with the current UI, which would cause the output to become very long. Perhaps it could be considered when the package manager UI is refactored into an interactive one in the future, but that won't happen anytime soon.

PR #2841, opened 2025-06-05
feat: allow upgrading only specific packages

sxyazi:

Resolves https://github.com/sxyazi/yazi/issues/2836

### DIFF
--- a/yazi-cli/src/args.rs
+++ b/yazi-cli/src/args.rs
@@ -76,7 +76,12 @@ pub(super) enum CommandPkg {
 	/// List all packages.
 	List,
 	/// Upgrade all packages.
-	Upgrade,
+	#[command(arg_required_else_help = true)]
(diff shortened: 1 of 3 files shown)


yazi-cli/src/args.rs
```diff
@@ -76,7 +76,12 @@ pub(super) enum CommandPkg {
 	/// List all packages.
 	List,
 	/// Upgrade all packages.
-	Upgrade,
+	#[command(arg_required_else_help = true)]
```
dexxiez, line 79: @sxyazi was this intentional? not sure on the codebase here but looking at the name, it does break if no IDs given. Just tested on `main` and found it.

Tyarel8: Same, how do you update all packages now? Just tried to do `ya pkg upgrade` and it shows the help

PR #2849, opened 2025-06-08
feat: port several widespread GUI keys to the input component

Axlefublr:
## Rationale of this PR

ctrl+left, ctrl+right, ctrl+delete, and ctrl+backspace are incredibly widespread and expected hotkeys


sxyazi:
Thanks for the patch!

Are these keys standard Vim or Emacs readline keybindings?

Axlefublr:
nope, they're just standard everywhere type keybindings. basically every app supports them, when it comes to gui applications especially; simplest example I can think of is the browser, but you can probably try most apps where you have an inputbox, and notice that all three are likely supported! :3

Axlefublr:
it might be that you're going for support of vim/emacs style bindings, but my patch includes universal bindings that users very much might expect. think of how Home and End are; ^delete, ^backspace, ^←→ are in the same type of field in terms of universalness

sxyazi (approved):
I see, seems reasonable to me, thank you!

Axlefublr:
:o fancy pr title \
thanks for the merge!

Issue #2874, opened 2025-06-14
Inaccessible directories/files with special names


TKperson:
### What system are you running Yazi on?

Linux Wayland

### What terminal are you running Yazi in?

Ghostty 1.1.3-arch1

### `yazi --debug` output

```Shell
Yazi
    Version: 25.6.11 (aae3f9e 2025-06-14)
    Debug  : false
    Triple : x86_64-unknown-linux-gnu (linux-x86_64)
    Rustc  : 1.87.0 (17067e9a 2025-05-09)

Ya
    Version: 25.5.31 (Arch Linux 2025-06-01)

Emulator
    TERM                : Some("screen-256color")
    TERM_PROGRAM        : Some("tmux")
    TERM_PROGRAM_VERSION: Some("3.5a")
    Brand.from_env      : Some(Ghostty)
    Emulator.detect     : Emulator { kind: Left(Ghostty), light: false, csi_16t: (10, 25), force_16t: false }

Adapter
    Adapter.matches    : Kgp
    Dimension.available: Dimension { rows: 39, columns: 189, width: 1890, height: 975 }

Desktop
    XDG_SESSION_TYPE           : Some("wayland")
    WAYLAND_DISPLAY            : Some("wayland-1")
    DISPLAY                    : Some(":0")
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: Some("9958d297641b5c84dcff93f9039d80a5ad37ab00_1748308084_1061943535")
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Variables
    SHELL           : Some("/usr/bin/zsh")
    EDITOR          : Some("nvim")
    VISUAL          : None
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
[...]

PR #2884, opened 2025-06-16
fix: proper invalid UTF-8 path support

sxyazi:
Fixes https://github.com/sxyazi/yazi/issues/2874


boydaihungst:
`ui.Span("test"):style({ bold = true, fg = "cyan", bg = "#080A0E" })`

This is not work anymore. I got this error message `expected a Style or nil`

Edit: Ok, I see we can use `:style(ui.Style():fg("cyan"):bg("#080A0E"):bold())`. But for a configurable plugin, how can we allow user to input their own style? Like this `:style(ui.Style():fg(opts.fg and opts.fg or ""):bg(...):bold():italic()......)`?

sxyazi:
That's expected, the parameter type for `style()` [can only be `ui.Style`](https://yazi-rs.github.io/docs/plugins/layout#span.style), and table is an undocumented undefined behavior that depends on the implementation, which can't be relied on reliably, for example, in your case, the `bold = true` is never worked since the bold style (among others) is represented and stored in `modifier` as a number. 

> But for a configurable plugin, how can we allow user to input their own style?

`ui.Style()` itself can be passed as a configuration value - it can be passed in during `setup()` or set globally in `th` like [the `git.yazi` plugin does](https://github.com/yazi-rs/plugins/tree/main/git.yazi#advanced). 

In the future, users will be able to set plugin-specific styles in their `theme.toml`, for example:

```toml
# theme.toml
[plugin_id]
foo = { bold = true, fg = "cyan", bg = "#080A0E" }
```

At that time, it will be accessible via `th.plugin_id.foo`, making the global `th` way an easier transition as it is implemented.

boydaihungst:
That's really cool. One more thing, `ui.Style()` can pass as a value in `setup()` in stable yazi right?

sxyazi:
Yes, for example:

```lua
require("foo"):setup {
  foo = ui.Style():bold():fg("red"),
  bar = ui.Style():bg("blue"):underline(),
}
```

boydaihungst:
After testing it, it can't set or get `foo` or `bar` style from `state`, so I guess the only option is `th` :

```lua
local set_state = ya.sync(function(state, key, value)
	state[key] = value
end)

local get_state = ya.sync(function(state, key)
	return state[key]
end)
```

boydaihungst:
This is not work `Url(path)`, if path is the invalid UTF-8 path.

PR #2892, opened 2025-06-18
feat: support opening and previewing files inside of Directories with…

TKperson:

* Resolves opening and previewing files inside of directories with invalid UTF-8 names in #2874

### DIFF
--- a/yazi-plugin/src/process/command.rs
+++ b/yazi-plugin/src/process/command.rs
@@ -146,8 +146,33 @@ impl UserData for Command {
 						me.inner.arg(String::from_utf8_lossy(&s.as_bytes()).as_ref());
 					}
 					Value::Table(t) => {
-						for s in t.sequence_values::<mlua::String>() {
-							me.inner.arg(String::from_utf8_lossy(&s?.as_bytes()).as_ref());
+						for v in t.sequence_values::<Value>() {
+							match &v? {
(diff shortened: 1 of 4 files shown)


yazi-plugin/src/process/command.rs
```diff
@@ -146,8 +146,33 @@ impl UserData for Command {
 						me.inner.arg(String::from_utf8_lossy(&s.as_bytes()).as_ref());
 					}
 					Value::Table(t) => {
-						for s in t.sequence_values::<mlua::String>() {
-							me.inner.arg(String::from_utf8_lossy(&s?.as_bytes()).as_ref());
+						for v in t.sequence_values::<Value>() {
```
sxyazi, line 149: `Command:arg()` should not accept `Url` as a parameter, because a `Url` may represent a remote file (along with remote file management will be added). 

Unlike APIs under `fs` (such as [`fs.remove()`](https://yazi-rs.github.io/docs/plugins/utils/#fs.remove)), which are designed to work across file systems, `Command` only searches for and executes user-local executables within the user's local file system, so passing `Url`s in those APIs are undefined behavior, which should be avoided.

Instead here, should be changed to support accepting bytes via `mlua::String`, and these bytes should be converted to `OsStr` before being passed to Rust's `Command::arg()`.

PR #2904, opened 2025-06-22
fix: expose `ya` CLI in the Snap build

### LINKED ISSUE #2903: ya is not exposed from the snap build

Linux X11

kitty v0.42.1

### `yazi --debug` output

```Shell
Irrelevant
```

### Describe the bug

Currently the snap build only exposes yazi but not ya, so when installing yazi from the snap store you can't run the package manager unless you explicitly run it from `/snap/yazi/current/ya`. The easy solution for this would be to expose a second app from the snap which points to ya.

### Minimal reproducer

1. Install yazi from snap `snap install yazi --classic`
2. Try running `ya` or `yazi.ya` => `Command not found`

### Anything else?

_No response_

### Checklist

- [ ] I can reproduce it after disabling all custom configs/plugins (`mv ~/.config/yazi ~/.config/yazi-backup`)

Sebastian-0:

Resolves #2903

## Rationale of this PR
Currently the snap build only exposes yazi but not ya, so when installing yazi from the snap store you can't run the package manager unless you explicitly run it from `/snap/yazi/current/ya`.

This PR solves this by exposing a second app from the snap which points to ya. Note that `ya` will be exposed as `yazi.ya` because snap does not support exposing multiple (non-prefixed) entrypoints for a single snap. However, this is still better than not exposing ya at all.

As far as I know this change should be sufficient, but if someone with stronger snap knowledge knows better please chime in.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -18,6 +18,10 @@ apps:
     command: yazi
     environment:
       PATH: $SNAP/bin:$SNAP/usr/bin:$PATH
+  ya:
+    command: ya
+    environment:
+      PATH: $SNAP/bin:$SNAP/usr/bin:$PATH
 
 parts:
   yazi:


sxyazi:
Thanks for the patch!

> Note that `ya` will be exposed as `yazi.ya` because snap does not support exposing multiple (non-prefixed) entrypoints for a single snap.

I don't know Snapcraft well, so I asked ChatGPT, and it seems we can rename binary files under `organize` in `parts` like this:

```yaml
name: my-multi-bin-app
base: core22
version: '1.0'
summary: Example snap with two binaries
description: |
  This is an example snap that installs and exposes two executables: foo and bar.

grade: stable
confinement: strict

parts:
  my-binaries:
    plugin: dump
    source: .
    organize:
      foo: bin/foo
      bar: bin/bar

apps:
  foo:
    command: bin/foo
  bar:
    command: bin/bar
```

Sebastian-0:
Thanks for the input!

According to [the docs](https://documentation.ubuntu.com/snapcraft/stable/reference/project-file/snapcraft-yaml/#parts.%3Cpart-name%3E.organize) organize only changes the target location of the binaries, it doesn't affect how the commands are exposed. You still need to expose them as apps, and according to the [corresponding docs](https://documentation.ubuntu.com/snapcraft/stable/reference/project-file/snapcraft-yaml/#apps) it will be exposed as `yazi.ya`. 

I also prompted chatgpt a bit about this problem but it claims there is no solution because the namespacing with the snap name is an underlying limitation of Snap to avoid colliding binary names between different snaps (allegedly, always take the LLM with a grain of salt).

sxyazi:
Ah, I see, thanks for the correction!

Ideally, we should always uses `ya` as the CLI name to make the built-in features like extraction work:

https://github.com/sxyazi/yazi/blob/ab0fe931506bb4f9fcbb114f107321a75145ec27/yazi-config/preset/yazi-default.toml#L50-L53

However, if it's not possible, I think I can accept `yazi.ya` - it's better than nothing.

CC @jnsgruk - sorry to bother you, but you're the most knowledgeable Snap expert I know, do you think there's any way to achieve that?

jnsgruk:
Hey! Change looks good.

On the command, you can request a "global alias" on the snapcraft discourse, as per this process: https://forum.snapcraft.io/t/process-for-aliases-auto-connections-and-tracks/455

Users of the snap can also use the following to achieve it locally:

```
sudo snap alias yazi.ya ya
```

Sebastian-0:
Great! I didn't know you can request an alias from snap, that is a neat solution.

The first step is to merge this patch, and after the next release we can request an alias :)

sxyazi:
That's awesome, didn't know that, thanks for the info @jnsgruk!

Will try to submit a "global alias" request in the next few days, let's go ahead and merge this PR first.

Issue #2894, opened 2025-06-18
chain fd (find) and rg (ripgrep) (search improvement)


HansMartinJ:
### `yazi --debug` output

```Shell
no issue
```

### Please describe the problem you're trying to solve

Loving yazi. One thing i use it for a lot is browsing massive folders in my org. fd and rg are my two most used operations.

A usual use case for me is to fd certain folders or files in project with a huge amount of them. Using yazi this is super easy, but one thing i often need to do is only find files that are in "prod" folders

e.g. say I'm in a dir like this
root/{projects...}/{test|prod}/{files...}

and I'm looking for the files in prod containing some name

I'd do [fd hotkey] {targetString}

Now i get a result, but it contains test files, so i want to [fd hotkey] prod. But this won't work, it will just do a normal search on all paths

(I can't use filter because that only works the name of the final file)

Same goes for rg, i search for some content but ah its in both test and prod, so I'd want to further filter out test

### Would you be willing to contribute this feature
No

### Describe the solution you'd like

Allow chaining rg and fd, so that e.g. when you get the fd result, you can do fd again and it only searches on the previous result. Or maybe extend filter to work on path prefix.

Hitting escape would either go one back in the search (to the prior result) or just completely removes the search

Out of scope side thoughts:
- it could be cool to have some sort of system in place to create views that are chains of rg and fd.
- introduce OR and NOT searching (e.g. searching in all non-test folders, or search both prod AND preprod)

### Additional context

No

### Checklist

- [x] I have searched the existing issues/discussions
- [x] The [latest nightly build](https://yazi-rs.github.io/docs/installation/#binaries) doesn't already have this feature

HansMartinJ:
Not sure what info it needs

sxyazi:
I think this should be achieved in some form through the `filter` command rather than `search`, because each search requires calling external tools (`rg`, `fd`), and these tools themselves don't provide a session-like concept. So when you search within search results (a second search), it's invisible to them, meaning Yazi must pass *all* previous search results to them. If the number of search results is large, this will cause performance issues and may also exceed the operating system's maximum command line length limit.

For your specific case, perhaps the `filter` command could be enhanced to support searching full paths instead of just filenames (this was also requested in https://github.com/sxyazi/yazi/discussions/2565).

> - it could be cool to have some sort of system in place to create views that are chains of rg and fd.
> - introduce OR and NOT searching (e.g. searching in all non-test folders, or search both prod AND preprod)

[...]

PR #2915, opened 2025-06-24
feat: `filter` by full path (prefix + filename) in search view instead of just filename

sxyazi:
An implementation of the feature request: https://github.com/sxyazi/yazi/discussions/2565, https://github.com/sxyazi/yazi/issues/2894

---

The search view flatly shows search results, for example, `a/b/c.txt`, where `a/b` is the prefix and `c.txt` is the file name.

Previously, users could only use the `filter` command to filter for the file name portion (`c.txt`) within the search results. 

With this PR, the full path `a/b/c.txt` can now be filtered, which makes it more intuitive, as users should be able to filter for the path as they see it in the list, rather than only the final part of it.

https://github.com/user-attachments/assets/ed9798af-345b-43bc-9a21-a8762178d053

### DIFF
--- a/yazi-fs/src/files.rs
+++ b/yazi-fs/src/files.rs
@@ -186,10 +186,9 @@ impl Files {
 		}
 
 		let (mut hidden, mut items) = if let Some(filter) = &self.filter {
-			urns.into_iter().partition(|u| {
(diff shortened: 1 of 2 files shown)


JeremieAlcaraz:
Love you work and your app ! Thanks for the new feat🙏♥️

PR #2095, opened 2024-12-24
feat: new `ya.confirm()` API

sxyazi:
An implementation of the feature request: https://github.com/sxyazi/yazi/issues/2082

Closes https://github.com/sxyazi/yazi/issues/2082

---

```lua
local answer = ya.confirm {
  pos = { "center", w = 40, h = 10 },
  title = "Test",
  content = "Hello, World!",
}
```

You can also apply a UI element to the title and content, to customize the appearance:

```lua
local answer = ya.confirm {
  pos = { "center", w = 40, h = 10 },
  title = ui.Line("Test"):fg("yellow"):bold(),
  content = ui.Text("Hello, World!"):bg("red"):fg("#ffffff"),
}
```

The available values for `pos` are the same as those for `ya.input()`, https://yazi-rs.github.io/docs/plugins/utils#ya.input

PR #2921, opened 2025-06-26
feat!: make `ya.confirm()` and `ui.Pos()` public API

sxyazi:
Follow up to https://github.com/yazi-rs/yazi-rs.github.io/commit/5db68d0a8ed9026aba3c51a13e7479ceb2f24e01

---

This PR makes the final changes to `ya.confirm()` and `ui.Pos()` before going public, and these changes are now reflected in the documentation:

- `ya.confirm()`: https://yazi-rs.github.io/docs/plugins/utils#ya.confirm
- `ui.Pos()`: https://yazi-rs.github.io/docs/plugins/layout#pos

### Use `body` instead of the term `content` in confirmations

A confirmation consists of three parts:

- Top: the title
- Middle: the confirmation prompt text
- Bottom: an additional list shown upon confirmation, such as a file list when deleting files

Previously, the middle part was referred to as `content`, but this was inaccurate because all parts except the title are "content" (i.e. the middle and bottom parts). So, the middle part is now changed to `body` from `content`.

```diff
# theme.toml / flavor.toml
[confirm]
- content = {}
+ body    = {}
```

```diff
# yazi.toml
[confirm]
- overwrite_content = "Will overwrite the following file:"
+ overwrite_body    = "Will overwrite the following file:"

- quit_content = "The following tasks are still running, are you sure you want to quit?"
+ quit_body    = "The following tasks are still running, are you sure you want to quit?"
```

## Deprecation

The `position` property of `ya.input()` has been deprecated and replaced by the new `pos` to align with `ya.confirm()` and its type name `ui.Pos`.

The original `position` property is still available, but it will trigger a deprecation warning.

Resolves: https://github.com/sxyazi/yazi/pull/2095#issuecomment-2561700908

### DIFF
--- a/yazi-config/preset/theme-dark.toml
+++ b/yazi-config/preset/theme-dark.toml
@@ -124,7 +124,7 @@ separator_style = { fg = "darkgray" }
 [confirm]
 border     = { fg = "blue" }
 title      = { fg = "blue" }
-content    = {}
+body       = {}
 list       = {}
 btn_yes    = { reversed = true }
 btn_no     = {}
--- a/yazi-config/preset/theme-light.toml
+++ b/yazi-config/preset/theme-light.toml
@@ -124,7 +124,7 @@ separator_style = { fg = "darkgray" }
 [confirm]
 border     = { fg = "blue" }
 title      = { fg = "blue" }
-content    = {}
+body       = {}
 list       = {}
 btn_yes    = { reversed = true }
 btn_no     = {}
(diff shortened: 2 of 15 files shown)


boydaihungst:
image

There's a typo.

sxyazi:
Fixed in https://github.com/yazi-rs/yazi-rs.github.io/commit/fe56791babb2a812808065eca173c88cd8264d6d and https://github.com/yazi-rs/plugins/commit/5513ea0c379170d0d50a676c9f7b8e43aea61059

PR #2959, opened 2025-07-07
fix: respect user's system media opener instead of hardcoding `mpv`

yezhizhen:
All platforms fails to play media file if "mpv" is not installed. There is also no response shown.

## Rationale of this PR

The default `open --interactive` menu for Media type is: 
https://github.com/sxyazi/yazi/blob/917e1f54a10445f2e25147c4b81a3c77d8233632/yazi-config/preset/yazi-default.toml#L68-L69

For most platforms, `mpv` is not built in. This PR forces `"play"` to use the default associated application defined in the OS, so that it is always responsive when pressing "Enter" (which triggers [mgr open](https://yazi-rs.github.io/docs/configuration/keymap#mgr.open) and first item in `use` list: "play"`).

### DIFF
--- a/yazi-config/preset/yazi-default.toml
+++ b/yazi-config/preset/yazi-default.toml
@@ -52,10 +52,10 @@ extract = [
 	{ run = 'ya pub extract --list %*',   desc = "Extract here", for = "windows" },
 ]
 play = [
-	{ run = 'xdg-open "$1"',                desc = "Play with OS default", for = "linux" },
-	{ run = 'open "$@"',                    desc = "Play with OS default", for = "macos" },
-	{ run = 'start "" "%1"', orphan = true, desc = "Play with OS default", for = "windows" },
-	{ run = 'termux-open "$1"',             desc = "Play with OS default", for = "android" },
+	{ run = 'xdg-open "$1"',                desc = "Play", for = "linux" },
+	{ run = 'open "$@"',                    desc = "Play", for = "macos" },
+	{ run = 'start "" "%1"', orphan = true, desc = "Play", for = "windows" },
+	{ run = 'termux-open "$1"',             desc = "Play", for = "android" },
 	{ run = '''mediainfo "$1"; echo "Press enter to exit"; read _''', block = true, desc = "Show media info", for = "unix" },
 ]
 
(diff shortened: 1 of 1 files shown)


sxyazi:
We should keep platform consistency here, i.e. if system default app is used for media flies on Windows, then other systems should also use the system default

yezhizhen:
Do you mean we should make similar change for `unix` as well? (For which I think `mpv` is also not built in)

I suppose we need to replicate following in `play`, but add `desc = "Play with OS default"`
https://github.com/sxyazi/yazi/blob/917e1f54a10445f2e25147c4b81a3c77d8233632/yazi-config/preset/yazi-default.toml#L37-L42

yezhizhen:
Or even simpler: Change 
`{ mime = "{audio,video}/*", use = [ "play", "reveal" ] },`
to
`{ mime = "{audio,video}/*", use = [ "open", "reveal" ] },` same as image, without special "play" handling.

sxyazi:
> Do you mean we should make similar change for `unix` as well? (For which I think `mpv` is also not built in)

Yes

> without special "play" handling

Please keep the `play` so that users can overwrite the media open rules more easier without touching the `[open]` section just like this:

```
[opener]
play = [
  # ...
]
```

yazi-config/preset/yazi-default.toml
```diff
@@ -53,7 +53,7 @@ extract = [
 ]
 play = [
 	{ run = 'mpv --force-window "$@"', orphan = true, for = "unix" },
-	{ run = 'mpv --force-window %*', orphan = true, for = "windows" },
+	{ run = 'start "" "%1"', orphan = true, desc = "Play with OS default", for = "windows" },
 	{ run = '''mediainfo "$1"; echo "Press enter to exit"; read _''', block = true, desc = "Show media info", for = "unix" },
```
yezhizhen, line 59: btw I believe this duplicate "unix" rule was redundant and left here by accident?
But can be used as fallback now.

sxyazi: It's not redundant, as it can be used in interactive open (with key `O`), where users can select and use non-default openers

sxyazi (approved):
Thank you!

yezhizhen:
Thanks for developing this amazing tool :) Really love it.

yezhizhen:
May I ask why the project is indented by tab instead of space? Just curious. (I just started using yazi today)

sxyazi:
https://sxyz.blog/why-i-use-tab-for-indent/

PR #2962, opened 2025-07-08
feat: new "Show media info" for `play` opener on Windows

yezhizhen:

For media types, Unix has "Show media info" option from interactive open. Windows doesn't.

## Rationale of this PR

Make the feature available for Windows as well.

### DIFF
--- a/yazi-config/preset/yazi-default.toml
+++ b/yazi-config/preset/yazi-default.toml
@@ -57,6 +57,7 @@ play = [
 	{ run = 'start "" "%1"', orphan = true, desc = "Play", for = "windows" },
 	{ run = 'termux-open "$1"',             desc = "Play", for = "android" },
 	{ run = '''mediainfo "$1"; echo "Press enter to exit"; read _''', block = true, desc = "Show media info", for = "unix" },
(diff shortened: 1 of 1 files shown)


yezhizhen:
https://github.com/user-attachments/assets/da4511a0-1871-4bd3-bced-82d8b3fe55f4

sxyazi:
Thanks for the patch!

`"%1"` is incorrect here, should use just `%1` instead, see https://yazi-rs.github.io/docs/configuration/yazi#opener

Also the `.exe` extension seems unnecessary, as it would prevent cases like `mediainfo.bat` or `mediainfo.cmd` from working

yezhizhen:
Thanks for the review!

> `"%1"` is incorrect here, should use just `%1` instead, see https://yazi-rs.github.io/docs/configuration/yazi#opener

Actually it is correct. Because Windows can contain space for file names... If I use `%1` instead, it fails to work.
image



> Also the .exe extension seems unnecessary

Thanks! Tested and made the change accordingly.

sxyazi:
> Actually it is correct. Because Windows can contain space for file names... If I use `%1` instead, it fails to work.

Ah you're right, thanks for the correction!

sxyazi (approved):
Thank you!

Issue #611, opened 2024-02-01
Tracking issues of Remote file management


Xuanwo:
This issue is used to track progress of yazi's remote file management.

Discuss at discord: https://discord.com/channels/1136203602898194542/1187669236672368681

## Goal

Make yazi a super terminal file manager that can manage remote files.

First of all, users can define new shemas in config:

```toml
[schemas]
rules = [
  { name = "backup", type = "s3", bucket = "my-backup-bucket", ... },
  { name = "google", type = "gdrive", referesh_token = "my-token", ... },
  { name = "nextcloud", type = "webdav", endpoint = "http://my-nextcloud-server", ... }
]
```

Then, users can use `yazi` to manage them like local files:

```shell
# manage remote s3 bucket!
yazi backup://path/to/dir

# manage my gdrive
yazi google://root

# manage my nextcloud
yazi nextcloud://path/to/my/notes
```

## Tasks

- [ ] Add shemas in config
- [ ] Add remote url scheme in URL
- [ ] Build opendal operator and handling io operations

dyxushuai:
Amazing idea, really like it.

dyxushuai:
I want to claim this.

og900aero:
It would be nice if they were also available on the go, so these "drives" could also be called up with a keyboard shortcuts.

dyxushuai:
> It would be nice if they were also available on the go, so these "drives" could also be called up with a keyboard shortcuts.

Opendal already supports Go binding https://github.com/apache/opendal/tree/main/bindings/go

sxyazi:
Can we implement it using the Go binding?

Xuanwo:
What's the relation between keyboard shortcuts and go? I would rather directly integrate yazi with rust because the go binding is still in its early stages and it requires CGO and the c binding, which could complicate our cross-platform support efforts.

Xuanwo:
> Would SSH / SFTP / SCP be within scope of this feature?

SFTP, Yes. 

Other protocols are considered as obsolete that not supported in opendal. See https://github.com/apache/opendal/issues/2643

danilevy1212:
Thank you! very much looking forward to this.

Xuanwo:
Hello everyone, I have started my work at https://github.com/sxyazi/yazi/pull/1152.

Current progress: opendal has been integrated into yazi, I'm working out the layout of yazi to make it works correctly.

[...]

PR #3034, opened 2025-08-04
feat!: virtual file system

sxyazi:
Prepare for https://github.com/sxyazi/yazi/issues/611

---

## ⚠️ Breaking changes

The `name` rule for [open](https://yazi-rs.github.io/docs/configuration/yazi#open), [fetchers](https://yazi-rs.github.io/docs/configuration/yazi#plugin.fetchers), spotters, [preloaders](https://yazi-rs.github.io/docs/configuration/yazi#plugin.preloaders), [previewers](https://yazi-rs.github.io/docs/configuration/yazi#plugin.previewers), [filetype](https://yazi-rs.github.io/docs/configuration/theme#filetype), and [`globs` icon rules](https://yazi-rs.github.io/docs/configuration/theme#icon) has been renamed to `url`, e.g.:

```diff
# yazi.toml
[plugin]
prepend_previewers = [
-   { name = "*/", run = "folder" },
+   { url = "*/", run = "folder" },
]
```

```diff
# theme.toml / flavor.toml
[filetype]
rules = [
-   { name = "*/", fg = "blue" },
+   { url = "*/", fg = "blue" },
]
```

This change addresses:

- The original term `name` was inaccurate since the `name` rule doesn't apply to the filename actually, but to the full file path, e.g., `{ name = "/Downloads/**/*.zip", ... }`.
- With this PR, virtual file system has been supported, and there needs to be a way for matching URL schemes, for instance, to disable previews for SFTP servers: `{ url = "sftp://**", run = "noop" }`.

URL is a superset of path, which means that the new `url` rule will inherit the functionality of the original `name` rule while also providing support for URL scheme matching.

## Deprecated

The `frag` property of `Url` has been deprecated and replaced by the new `domain`:

```diff
- print(url.frag)
+ print(url.domain)
```

This change is part of the redesign of the URL system. Previously, the fragment was included after the URL's path:

```sh
search:///path/to/file#keyword
```

But this means that the character `#` could no longer appear in the path, otherwise it would be unclear where the fragment starts, hence it required the path to be URL encoded:

```sh
search://%2Fpath%2Fto%2Ffile#keyword
```

This reduced readability and made typing URLs more cumbersome since the path is the most critical part of a URL. The new implementation stores the original fragment as the URL's domain:

```sh
search://keyword//path/to/file
```

Fragments occur far less frequently than paths, and they are typically only used for internal data exchange in Yazi, and users usually do not have to input them, so encoding fragments instead is acceptable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3556,6 +3556,7 @@ dependencies = [
  "clap_complete",
  "clap_complete_fig",
  "clap_complete_nushell",
+ "futures",
  "regex",
  "serde",
  "vergen-gitcl",
--- a/yazi-actor/src/cmp/trigger.rs
+++ b/yazi-actor/src/cmp/trigger.rs
@@ -1,7 +1,7 @@
 use std::{ffi::OsString, mem, path::MAIN_SEPARATOR_STR};
 
 use anyhow::Result;
-use yazi_fs::{CWD, expand_url, services::Local};
+use yazi_fs::{CWD, expand_url, provider};
 use yazi_macro::{act, render, succ};
 use yazi_parser::cmp::{CmpItem, ShowOpt, TriggerOpt};
 use yazi_proxy::CmpProxy;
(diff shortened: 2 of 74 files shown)


boydaihungst:
Also, `[open] > rules`: `name` -> `url`

Issue #2050, opened 2024-12-15
Multiple tmux clients causes buggy inputs


musjj:
### What system are you running Yazi on?

Linux Wayland

### What terminal are you running Yazi in?

kitty 0.37.0
<details>
<summary><h3><code>yazi --debug</code> output</h3></summary>

```Shell
Yazi
    Version: 0.4.0 (0b340b6 2024-12-09)
    Debug  : false
    Triple : x86_64-unknown-linux-gnu (linux-x86_64)
    Rustc  : 1.82.0 (f6e511ee 2024-10-15)

Ya
    Version: 0.4.0 (0b340b6 2024-12-09)

Emulator
    Brand.from_env      : Some(Kitty)
    Emulator.detect     : Emulator { kind: Left(Kitty), light: false, cell_size: Some((8, 18)) }
    Emulator.detect_full: Ok(Emulator { kind: Left(Kitty), light: false, cell_size: Some((8, 18)) })

Adapter
    Adapter.matches: Kgp

Desktop
    XDG_SESSION_TYPE           : Some("wayland")
    WAYLAND_DISPLAY            : Some("wayland-0")
    DISPLAY                    : Some(":0")
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

NVIM
    NVIM          : false
    Neovim version: 0.10.2

Variables
    SHELL           : Some("/run/current-system/sw/bin/zsh")
    EDITOR          : Some("nvim")
    VISUAL          : None
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None

Text Opener
    default     : Some(Opener { run: "$EDITOR \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block-create: Some(Opener { run: "$EDITOR \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block-rename: Some(Opener { run: "$EDITOR \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })

Multiplexers
    TMUX               : 2
    tmux version       : tmux 3.5a
    tmux build flags   : enable-sixel=Unknown
    ZELLIJ_SESSION_NAME: None
    Zellij version     : No such file or directory (os error 2)

[...]

PR #3118, opened 2025-08-28
Fix for tmux: correct cursor restoration and fix CSI u detection

alberti42:
### Description

This PR fixes two long-standing issues with Yazi's `tmux` integration:
1.  It corrects the cursor restoration on exit, preventing the cursor from being changed to a block.
2.  It fixes the detection of the "CSI u" advanced keyboard protocol.

The solution involves avoiding ANSI queries inside `tmux`, which are currently not supported by `tmux` itself, and instead relying on `tmux`'s native features and configuration.

***

### 1. Root Cause of the Cursor Issue

When Yazi starts, it attempts to query the current cursor style. However, the foundational issue is that **`tmux`'s own terminal emulator does not respond to the `DECRQSS` query (`\x1bP$q q\x1b\\`)** that Yazi uses.

This can be verified by sending the query directly to `tmux` with the following script. Inside `tmux`, the script will time out with no response.

```bash
#!/bin/bash
# test_cursor.sh

# Ensure the terminal is in a sane state on exit
trap 'stty echo' EXIT

# Hide characters typed and disable canonical mode
stty -echo -icanon

echo "Sending cursor shape query..."
printf '\x1bP$q q\x1b\\'

# Loop briefly to capture the asynchronous response
response=""
while read -s -t 1 -n 1 char; do
    response+="$char"
done

# Print the captured response in a human-readable hex format
echo "Response received:"
echo -n "$response" | hexdump -C

echo "Test finished."
```

Because the query receives no response, Yazi's parsing logic fails and triggers a fallback case that unconditionally sets a steady block cursor (`\e[2 q`) upon exit.

***

### 2. Fix for CSI u Keyboard Protocol Detection

Similarly, Yazi attempts to detect support for the modern "CSI u" keyboard protocol (for unambiguous key reporting, e.g., distinguishing `Ctrl+i` from `Tab`) using an ANSI query (`\x1b[?u`). This query also receives no response from `tmux`, meaning this important feature was non-functional for `tmux` users.

This PR replaces the unreliable ANSI query with a robust, `tmux`-specific command. A new function, `Mux::tmux_has_csi_u()`, has been added to `yazi-adapter/src/mux.rs`. It directly queries `tmux`'s own settings by running:

```bash
tmux display-message -p '#{extended-keys},#{extended-keys-format}'
```

This change correctly detects and enables the CSI u protocol when configured in `tmux`, significantly improving keyboard handling. The original ANSI query method is preserved as a fallback for non-`tmux` terminals.

***

### 3. The Solution for the Cursor Issue

The most robust and correct solution for the cursor problem is to trust `tmux`'s native restoration mechanism.

When an application uses the alternate screen (`EnterAlternateScreen`/`LeaveAlternateScreen`), `tmux` automatically saves a "snapshot" of the pane's state (including the cursor style) and restores it upon exit. Yazi's faulty manual `RestoreCursor` command was overriding this correct, native behavior.

This PR introduces a conditional check (`if !yazi_adapter::TMUX.get()`) to prevent `RestoreCursor::store()` from running inside `tmux`. This disables Yazi's faulty restore logic and allows `tmux`'s superior alternate screen restoration to function correctly, ensuring any cursor style the user had set is perfectly restored.

***

### 4. Practical Tip for `tmux` Users (Context for Developers)

This investigation also uncovered a notable quirk in `tmux`'s default behavior. Without an explicit `cursor-style` set in `~/.tmux.conf`, `tmux` operates in a **"lazy initialization"** mode. The first command to set a cursor style in any pane acts as a trigger, which correctly sets the cursor in the target pane but also sets a default "block" cursor in all other panes.

For a more predictable environment, `tmux` users should be encouraged to set an initial cursor style in their configuration. This proactively initializes `tmux`'s cursor management, ensuring all cursor changes are local from the start.

**Example for `~/.tmux.conf`:**
```tmux
# Set a default cursor style to ensure predictable behavior
set-option -g cursor-style bar
```

***

### 6. Note on Code Structure

To address the tmux-specific issues, we have deliberately separated the startup logic into distinct if/else blocks. This approach was chosen to make the code more readable and to clearly isolate the tmux-specific execution path, avoiding multiple conditional checks scattered throughout the function.

A future refactoring could potentially abstract this further to reduce code duplication while still cleanly handling the different logic required for tmux versus other terminals. However, for the purpose of this fix, this clear separation was deemed the most effective and understandable solution.

***

### 7. Testing Environment

This behavior and the corresponding fix were tested and verified under the following environment:

* **Operating System:** macOS
* **`tmux` Version:** 3.5a
* **Terminal Emulators:**
    * WezTerm `20240203-110809-5046fc22`
    * iTerm2 `3.5.14`

### DIFF
--- a/yazi-adapter/src/mux.rs
+++ b/yazi-adapter/src/mux.rs
@@ -20,6 +20,37 @@ impl Mux {
 			Cow::Borrowed(s)
 		}
 	}
+	
+	pub fn tmux_has_csi_u() -> bool {
+	    let output = time!(
+	        "Querying tmux for CSI u support",
+	        std::process::Command::new("tmux")
+	            .args(["display-message", "-p", "#{extended-keys},#{extended-keys-format}"])
+	            .output()
+	    );
+
+	    match output {
+	        Ok(o) if o.status.success() => {
+	            // Success case: The command ran and exited with code 0.
+	            let stdout = String::from_utf8_lossy(&o.stdout);
+	            stdout.trim() == "on,csi-u"
+	        }
+	        Ok(o) => {
+	            // Execution Error: The command ran but failed.
+	            error!(
+	                "Querying tmux for CSI u support failed with status {:?}: {}",
+	                o.status,
+	                String::from_utf8_lossy(&o.stderr)
+	            );
+	            false
+	        }
+	        Err(e) => {
+	            // Spawn Error: The command could not be started.
+	            error!("Failed to spawn `tmux display-message` for CSI u query: {}", e);
+	            false
+	        }
+	    }
+	}
 
 	pub fn tmux_passthrough() {
 		let output = time!(
--- a/yazi-fm/src/term.rs
+++ b/yazi-fm/src/term.rs
@@ -30,23 +30,54 @@ impl Term {
 			yazi_adapter::Mux::tmux_passthrough();
 		}
 
-		execute!(
-			TTY.writer(),
-			yazi_term::If(!TMUX.get(), EnterAlternateScreen),
-			Print("\x1bP$q q\x1b\\"), // Request cursor shape (DECRQSS query for DECSCUSR)
-			Print(Mux::csi("\x1b[?12$p")), // Request cursor blink status (DECSET)
-			Print("\x1b[?u"),         // Request keyboard enhancement flags (CSI u)
-			Print(Mux::csi("\x1b[0c")), // Request device attributes
-			yazi_term::If(TMUX.get(), EnterAlternateScreen),
-			EnableBracketedPaste,
-			yazi_term::If(!YAZI.mgr.mouse_events.get().is_empty(), EnableMouseCapture),
-		)?;
+		let mut csi_u_supported = false; // This will hold the result of our detection.
 
-		let resp = Emulator::read_until_da1();
-		Mux::tmux_drain()?;
-		yazi_term::RestoreCursor::store(&resp);
+    if TMUX.get() {
+        // --- TMUX-SPECIFIC LOGIC ---
+        // 1. Reliably detect CSI u support by querying tmux's own options.
+        csi_u_supported = Mux::tmux_has_csi_u();
 
-		CSI_U.store(resp.contains("\x1b[?0u"), Ordering::Relaxed);
+        // 2. Send only the essential queries, skipping the problematic cursor ones.
+        execute!(
+					TTY.writer(),
+					Print(Mux::csi("\x1b[0c")), // Request device attributes
+					EnterAlternateScreen,
+					EnableBracketedPaste,
+					yazi_term::If(!YAZI.mgr.mouse_events.get().is_empty(), EnableMouseCapture),
+				)?;
+
+        // 3. Read the terminal's response, but do NOT call store().
+        //    This allows tmux's native alternate screen restoration to work.
+    		Emulator::read_until_da1();
+				Mux::tmux_drain()?;
+
+    } else {
+        // --- ORIGINAL LOGIC FOR NON-TMUX TERMINALS ---
+        // 1. Send all original queries, as they may be supported by other terminals.
+				execute!(
+					TTY.writer(),
+					yazi_term::If(!TMUX.get(), EnterAlternateScreen),
+					Print("\x1bP$q q\x1b\\"), // Request cursor shape (DECRQSS query for DECSCUSR)
+					Print(Mux::csi("\x1b[?12$p")), // Request cursor blink status (DECSET)
+					Print("\x1b[?u"),         // Request keyboard enhancement flags (CSI u)
+					Print(Mux::csi("\x1b[0c")), // Request device attributes
+					yazi_term::If(TMUX.get(), EnterAlternateScreen),
+					EnableBracketedPaste,
+					yazi_term::If(!YAZI.mgr.mouse_events.get().is_empty(), EnableMouseCapture),
+				)?;
+
+        // 2. Use the original logic to parse the response and store the cursor state.
+				let resp = Emulator::read_until_da1();
+				Mux::tmux_drain()?;
+				yazi_term::RestoreCursor::store(&resp);
+
+        // 3. Check the ANSI response for CSI u support, as before.
+        if resp.contains("\x1b[?0u") {
+            csi_u_supported = true;
+        }
+    }
+
+		CSI_U.store(csi_u_supported, Ordering::Relaxed);
 		if CSI_U.load(Ordering::Relaxed) {
 			PushKeyboardEnhancementFlags(
 				KeyboardEnhancementFlags::DISAMBIGUATE_ESCAPE_CODES


yazi-adapter/src/mux.rs
```diff
@@ -20,6 +20,37 @@ impl Mux {
 			Cow::Borrowed(s)
 		}
 	}
+	
+	pub fn tmux_has_csi_u() -> bool {
+	    let output = time!(
+	        "Querying tmux for CSI u support",
+	        std::process::Command::new("tmux")
+	            .args(["display-message", "-p", "#{extended-keys},#{extended-keys-format}"])
```
sxyazi, line 28: Thank you for your patch, but this is not the correct way to query CSI u, as it does not work in remote SSH sessions because it always calls the system's local `tmux`.

alberti42: Ok. Perhaps we should think of a different way to probe CSI u in `tmux`. I will test tmux in ssh myself to reproduce the behavior you mentioned.

However, I hope you will consider the other part of the patch—namely, avoiding restoring the cursor in `tmux`, because `tmux` misbehaves. Rather, we should skip:

```rust
yazi_term::RestoreCursor::store(&resp);
```

when in `tmux` because restoring the cursor is handled by `tmux` locally in the selected pane, and any other attempt to restore the cursor would only create problems, since the cursor style cannot be queried in `tmux` using `\x1bP$q q\x1b\\`.

alberti42: Thank you for the quick feedback. I have given some extra thought to what you stated. I'd like to clarify my understanding of the issue.

When Yazi is running inside a tmux session on a remote server, my understanding is that

```rust 
std::process::Command::new("tmux")
	            .args(["display-message", "-p", "#{extended-keys},#{extended-keys-format}"])
```

will execute the tmux binary on that same remote server. Could you please elaborate on the scenario where the local system's tmux would be called instead?

In general, what would it be the ideal behavior in the situation when running `tmux` runs on a remote machine? Are you also considering the case of nested `tmux` servers i.e., one running on the local and one on the remote machine?   

I agree that a solution using standard terminal protocols would be better. However, since the ANSI query for CSI u (\x1b[?u) is not supported by tmux, what would be the preferred method for detecting this feature? I am happy to implement the correct approach.

sxyazi:
> However, I hope you will consider the other part of the patch—namely, avoiding restoring the cursor in `tmux`, because `tmux` misbehaves. Rather, we should skip:

This breaks the existing tmux cursor restoration behavior:

- Start tmux
- Run `echo -e -n "\x1b[\x32 q"` to set the cursor to *steady block*
- Compare Yazi with this PR, and without this PR

## With this PR: The cursor becomes a blinking block

https://github.com/user-attachments/assets/7231c44f-545a-4ea7-8521-09a01262422b

## Without this PR: The cursor remains a steady block

https://github.com/user-attachments/assets/95377e3d-1969-4573-ab78-74e7c6ed316f

sxyazi:
> Could you please elaborate on the scenario where the local system's tmux would be called instead?

Sorry for the confusion, I meant tmux running on the host machine (not remote), and runs Yazi on the remote.

> In general, what would it be the ideal behavior in the situation when running `tmux` runs on a remote machine? Are you also considering the case of nested `tmux` servers i.e., one running on the local and one on the remote machine?   

tmux on the host and tmux on remote should behave the same, that is, this detection should always work, regardless of the user's environment or where the user runs tmux.

> since the ANSI query for CSI u (\x1b[?u) is not supported by tmux, what would be the preferred method for detecting this feature?

I'm not familiar with tmux, but as far as I know, the ANSI query is the most reliable and standard way to detect CSI u support. If tmux doesn't support it, then we should ask tmux to support it, rather than implementing a problematic hack in Yazi.

alberti42:
Thank you for taking the time to try out my pull request (PR).

### Cursor restoration

My main concern, which affects me directly, is the (failed) restoration of the cursor. There must be a difference in our setups because I see the opposite behavior from what you showed in the video. If I took a screencast, it would show the exactly opposite of what you experience. The fact that the cursor changed in all my tmux windows has always bothered me.

I would really like to understand the fundamental difference between our setups. I am running tmux 3.5a. I tried it without a .tmux.conf file, using only the default settings. I tried this on Linux and macOS. On all these systems, I also tried running a shell without RC files, like with `zsh -df` or `bash --noprofile --norc`. In every attempt, when I use the non-patched version of `yazi`, the cursor reverts to the block cursor (the fallback case you have hardcoded).

Have you tested your setup with the default settings? I would just like to rule out the possibility that the different behavior originates from a particular setting of yours.

Finally, I ran further tests using yazi's log:

```bash
YAZI_LOG=debug yazi
```

In WezTerm **without tmux**, I obtain:

```text
  2025-08-29T07:53:36.692729Z DEBUG yazi_adapter::emulator: Terminal responded to DA1 in 4.202583ms: [27, 80, 62, 124, 87, 101, 122, 84, 101, 114, 109, 32, 50, 48, 50, 52, 48, 50, 48, 51, 45, 49, 49, 48, 56, 48, 57, 45, 53, 48, 52, 54, 102, 99, 50, 50, 27, 92, 27, 91, 54, 59, 51, 56, 59, 49, 56, 116, 27, 93, 49, 49, 59, 114, 103, 98, 58, 50, 56, 50, 56, 47, 50, 57, 50, 57, 47, 51, 53, 51, 53, 27, 92, 27, 91, 63, 54, 53, 59, 52, 59, 54, 59, 49, 56, 59, 50, 50, 99]
    at yazi-adapter/src/emulator.rs:129

  2025-08-29T07:53:36.693358Z DEBUG yazi_adapter::emulator: Detected background color: 2828/2929/3535 (luma = 0.16)
    at yazi-adapter/src/emulator.rs:192

  2025-08-29T07:53:36.714729Z DEBUG yazi_adapter::emulator: Terminal responded to DA1 in 4.162166ms: [27, 80, 48, 36, 114, 27, 92, 27, 91, 63, 49, 50, 59, 50, 36, 121, 27, 91, 63, 48, 117, 27, 91, 63, 54, 53, 59, 52, 59, 54, 59, 49, 56, 59, 50, 50, 99]
    at yazi-adapter/src/emulator.rs:129
```

This translates to:

```text
┌────────┬─────────────────────────┬─────────────────────────┬────────┬────────┐
│00000000│ 1b 50 3e 7c 57 65 7a 54 ┊ 65 72 6d 20 32 30 32 34 │•P>|WezT┊erm 2024│
│00000010│ 30 32 30 33 2d 31 31 30 ┊ 38 30 39 2d 35 30 34 36 │0203-110┊809-5046│
│00000020│ 66 63 32 32 1b 5c 1b 5b ┊ 36 3b 33 38 3b 31 38 74 │fc22•\•[┊6;38;18t│
│00000030│ 1b 5d 31 31 3b 72 67 62 ┊ 3a 32 38 32 38 2f 32 39 │•]11;rgb┊:2828/29│
│00000040│ 32 39 2f 33 35 33 35 1b ┊ 5c 1b 5b 3f 36 35 3b 34 │29/3535•┊\•[?65;4│
│00000050│ 3b 36 3b 31 38 3b 32 32 ┊ 63                      │;6;18;22┊c       │
└────────┴─────────────────────────┴─────────────────────────┴────────┴────────┘
```

and the important part with the correct cursor information:

```
┌────────┬─────────────────────────┬─────────────────────────┬────────┬────────┐
│00000000│ 1b 50 30 24 72 1b 5c 1b ┊ 5b 3f 31 32 3b 32 24 79 │•P0$r•\•┊[?12;2$y│
│00000010│ 1b 5b 3f 30 75 1b 5b 3f ┊ 36 35 3b 34 3b 36 3b 31 │•[?0u•[?┊65;4;6;1│
│00000020│ 38 3b 32 32 63          ┊                         │8;22c   ┊        │
└────────┴─────────────────────────┴─────────────────────────┴────────┴────────┘
```

If I repeat the test in tmux, I obtain the following:

```text
 cat /Users/andrea/.local/state/yazi/yazi.log
  2025-08-29T08:10:35.602393Z DEBUG yazi_adapter::brand: Detected special environment variable: WEZTERM_EXECUTABLE
    at yazi-adapter/src/brand.rs:85

  2025-08-29T08:10:35.603207Z DEBUG yazi_adapter::emulator: Terminal responded to DA1 in 374.791µs: [27, 80, 62, 124, 116, 109, 117, 120, 32, 51, 46, 53, 97, 27, 92, 27, 91, 54, 59, 51, 56, 59, 49, 56, 116, 27, 93, 49, 49, 59, 114, 103, 98, 58, 50, 56, 50, 56, 47, 50, 57, 50, 57, 47, 51, 53, 51, 53, 7, 27, 91, 63, 49, 59, 50, 99]
    at yazi-adapter/src/emulator.rs:129

  2025-08-29T08:10:35.603480Z DEBUG yazi_adapter::emulator: Detected background color: 2828/2929/3535 (luma = 0.16)
    at yazi-adapter/src/emulator.rs:192

  2025-08-29T08:10:35.612107Z DEBUG yazi_adapter::mux: Running `tmux set -p allow-passthrough on` took 8.31125ms
    at yazi-adapter/src/mux.rs:23

  2025-08-29T08:10:35.616650Z DEBUG yazi_adapter::mux: Running `tmux show-environment` took 4.5265ms
    at yazi-adapter/src/mux.rs:78

  2025-08-29T08:10:35.616909Z DEBUG yazi_adapter::brand: Detected special environment variable: WEZTERM_EXECUTABLE
    at yazi-adapter/src/brand.rs:85

  2025-08-29T08:10:35.620945Z DEBUG yazi_adapter::emulator: Terminal responded to DA1 in 3.782125ms: [27, 91, 54, 59, 51, 56, 59, 49, 56, 116, 27, 93, 49, 49, 59, 114, 103, 98, 58, 50, 56, 50, 56, 47, 50, 57, 50, 57, 47, 51, 53, 51, 53, 7, 27, 80, 62, 124, 87, 101, 122, 84, 101, 114, 109, 32, 50, 48, 50, 52, 48, 50, 48, 51, 45, 49, 49, 48, 56, 48, 57, 45, 53, 48, 52, 54, 102, 99, 50, 50, 27, 92, 27, 91, 63, 54, 53, 59, 52, 59, 54, 59, 49, 56, 59, 50, 50, 99]
    at yazi-adapter/src/emulator.rs:129

  2025-08-29T08:10:35.624710Z DEBUG yazi_adapter::emulator: Terminal responded to DSR in 3.755667ms: [27, 91, 48, 110]
    at yazi-adapter/src/emulator.rs:145

  2025-08-29T08:10:35.624717Z DEBUG yazi_adapter::emulator: Detected background color: 2828/2929/3535 (luma = 0.16)
    at yazi-adapter/src/emulator.rs:192

  2025-08-29T08:10:35.645178Z DEBUG yazi_adapter::emulator: Terminal responded to DA1 in 4.077875ms: [27, 91, 63, 49, 50, 59, 50, 36, 121, 27, 91, 63, 54, 53, 59, 52, 59, 54, 59, 49, 56, 59, 50, 50, 99]
    at yazi-adapter/src/emulator.rs:129

  2025-08-29T08:10:35.649107Z DEBUG yazi_adapter::emulator: Terminal responded to DSR in 3.921625ms: [27, 91, 48, 110]
    at yazi-adapter/src/emulator.rs:145
```

This translates to:

```text
┌────────┬─────────────────────────┬─────────────────────────┬────────┬────────┐
│00000000│ 1b 50 3e 7c 74 6d 75 78 ┊ 20 33 2e 35 61 1b 5c 1b │•P>|tmux┊ 3.5a•\•│
│00000010│ 5b 36 3b 33 38 3b 31 38 ┊ 74 1b 5d 31 31 3b 72 67 │[6;38;18┊t•]11;rg│
│00000020│ 62 3a 32 38 32 38 2f 32 ┊ 39 32 39 2f 33 35 33 35 │b:2828/2┊929/3535│
│00000030│ 07 1b 5b 3f 31 3b 32 5a ┊                         │••[?1;2Z┊        │
└────────┴─────────────────────────┴─────────────────────────┴────────┴────────┘
```

```
┌────────┬─────────────────────────┬─────────────────────────┬────────┬────────┐
│00000000│ 1b 5b 36 3b 33 38 3b 31 ┊ 38 74 1b 5d 31 31 3b 72 │•[6;38;1┊8t•]11;r│
│00000010│ 67 62 3a 32 38 32 38 2f ┊ 32 39 32 39 2f 33 35 33 │gb:2828/┊2929/353│
│00000020│ 35 07 1b 50 3e 7c 57 65 ┊ 7a 54 65 72 6d 20 32 30 │5••P>|We┊zTerm 20│
│00000030│ 32 34 30 32 30 33 2d 31 ┊ 31 30 38 30 39 2d 35 30 │240203-1┊10809-50│
│00000040│ 34 36 66 63 32 32 1b 5c ┊ 1b 5b 3f 36 35 3b 34 3b │46fc22•\┊•[?65;4;│
│00000050│ 36 3b 31 38 3b 32 32 63 ┊                         │6;18;22c┊        │
└────────┴─────────────────────────┴─────────────────────────┴────────┴────────┘
```

```text
┌────────┬─────────────────────────┬─────────────────────────┬────────┬────────┐
│00000000│ 1b 5b 3f 31 32 3b 32 24 ┊ 79 1b 5b 3f 36 35 3b 34 │•[?12;2$┊y•[?65;4│
│00000010│ 3b 36 3b 31 38 3b 32 32 ┊ 5a                      │;6;18;22┊Z       │
└────────┴─────────────────────────┴─────────────────────────┴────────┴────────┘
```

and

```text
┌────────┬─────────────────────────┬─────────────────────────┬────────┬────────┐
│00000000│ 1b 5b 30 6e             ┊                         │•[0n    ┊        │
└────────┴─────────────────────────┴─────────────────────────┴────────┴────────┘
```

tmux responds with the cursor blink status (?12;2$y) but **it omits** the cursor shape and `CSI u` responses.

I could open a feature request in `tmux` GitHub or try to implement the proper fix in `tmux` code. The developer is quite responsive. However, I still want to emphasize the fact that by omitting in yazi the cursor restoration, `tmux` properly handles cursor restoration on its own! I am just surprised that this is not happening on your setup. If we figured out why it does not work for you, we could have a solution in much a shorter time than modifying `tmux` code.

Eventually, I will open a feature request in `tmux` because it makes sense on the long run to support this query as well as that for `CSI u`.

***

### CSI u support

I am less interested in supporting CSI u; I don't even know if tmux supports it properly. I can enable it in tmux.

```tmux
# Enable support for advanced key reporting (CSI u)
set-option -g extended-keys on
set-window-option -g extended-keys-format csi-u
```

but, honestly, I could not see any difference in tmux, as I was not even able to differentiate key combinations like Ctrl+i from Tab.

Even if this was not my primary concern, I became aware of it while reading `yazi`'s code, and I just want to point out that, on all the macOS and Linux machines I tested, tmux does not support this terminal query to test the CSI u support.

sxyazi:
I'm using kitty 0.42.2 , tmux tmux 3.5a, both with default config

alberti42:
I tried your setup with default configurations for `kitty` and `tmux`, same versions. The problem still remains for me. I would be interested to know what your `yazi` log file shows when you run in tmux and compare to mine. It remains for me a mystery why your setup behaves differently.

alberti42:
I tried an additional Linux Ubuntu machine and a Windows machine on site (i.e., avoiding any SSH connection). I still experience the same problem with the cursor. At this point, I am quite certain it must be something affecting many people. I don't know why for you it works differently.

For your information. I experience the same problem of wrong cursor restoration with `neovim`. There are other people complaining about it: https://github.com/neovim/neovim/issues/21480 and https://github.com/neovim/neovim/issues/4396. It does not look like `neovim` was properly fixed. People resorted to using a hook function on exit that restores the cursor to the desired one.

I realize that terminal emulators can be really frustrating. One tries to bring in modernity while keeping compatibility with standards from 1980 or perhaps even from earlier times...

alberti42:
Thank you for the feedback. After extensive further testing, I believe I've uncovered the complete picture, including why our experiences with cursor restoration have been different.

### The Key Discovery: `tmux` Configuration is Crucial

The critical factor for successful cursor restoration is whether the `cursor-style` option is explicitly set in `~/.tmux.conf`.

* **When `cursor-style` is set** (e.g., `set-option -g cursor-style bar`), `tmux`'s native alternate screen feature correctly restores the cursor to this configured style upon application exit. This is the environment where my simple patch works perfectly.

* **When `cursor-style` is unset** (the default `tmux` behavior), the restoration is undefined and inconsistent. This is likely your environment and explains why you could not reproduce the successful restoration.

This reliance on user configuration highlights the need for a more robust solution within Yazi itself until `tmux` provides a better support for handling the cursor style.

### Diagnostic Proof

To illustrate the underlying technical limitations, the following script can be used to test `DECRQSS` queries.

```bash
#!/bin/bash
# query_cursor.sh

echo "Preparing terminal..."
stty_orig=$(stty -g)
trap 'stty "$stty_orig"' EXIT

# Raw-ish mode so replies aren't buffered/echoed
stty -echo -icanon time 0 min 0

# The query to be sent
DECRQSS_QUERY='\eP$q q\e\\'

if [[ "$1" == "--passthrough" ]]; then
    # Save and set allow-passthrough
    ALLOW_PASSTHROUGH=$(tmux show-options -pv allow-passthrough)
    tmux set-option -p allow-passthrough on
    trap 'tmux set-option -p allow-passthrough ${ALLOW_PASSTHROUGH}; stty "$stty_orig"' EXIT

    echo "Sending PASSTHROUGH query to underlying terminal..."
    PASSTHROUGH_WRAPPED_QUERY="\ePtmux;\e${DECRQSS_QUERY}\e\\"
    printf "$PASSTHROUGH_WRAPPED_QUERY" > /dev/tty
else
    echo "Sending DIRECT query to the current terminal..."
    printf "$DECRQSS_QUERY" > /dev/tty
fi

# Give the terminal time to respond and read the result
sleep 0.3
dd bs=256 count=1 2>/dev/null | hexdump -C

echo "Test finished."
```

#### Test Results:
1.  **Native Terminal (e.g., Kitty, WezTerm):** The query works perfectly and returns the full cursor style (e.g., `|.P1$r6 q.\|` for a steady bar).
2.  **Inside `tmux` (Direct Query):** `tmux` provides **no response**.
3.  **Inside `tmux` (Passthrough Query):** The underlying terminal returns a generic acknowledgment (`|.P0$r.\|`) with **no actual cursor style data**.

### The Conceptual Flaw in Yazi's Current Approach

These tests prove that querying the cursor inside `tmux` is unreliable.

**More importantly, the passthrough strategy is conceptually flawed.** The cursor the user sees is managed and rendered by `tmux` on a per-pane basis. The state of the underlying terminal's cursor is irrelevant, as `tmux` completely controls it. Therefore, asking the underlying terminal for its cursor style is asking the wrong question, even if it worked reliably.

### Proposed Solutions

Given these limitations in `tmux`, I see two viable paths forward.

**Option A: The Simple Fix (Rely on User Configuration)**
* **Action:** Merge the existing fix, which disables Yazi's faulty restore logic inside `tmux`.
* **Result:** This allows `tmux`'s native alternate screen restoration to work. We would add documentation explaining that for perfect restoration, users **must** set `cursor-style` in their `.tmux.conf`.
* **Pros:** The code change is already provided with the PR.
* **Cons:** It relies on `tmux` non-documented behavior, where the cursor style is restored to `cursor-style` after exiting the alternate screen.

**Option B: The Robust Fix (Application-Level Handling)**
* **Action:** Refactor the `tmux` adapter to query `tmux`'s `cursor-style` and `extended-keys` options directly using a single, efficient `tmux display-message` command at startup.
* **Result:** Yazi would store this value and, on exit, send a specific `DECSCUSR` command to restore the cursor to match the user's configured `tmux` style.
* **Pros:** Works for all users "out of the box," regardless of their configuration, and it is the best we can do until `tmux` provides proper support for cursor style.
* **Cons:** Some more code to write.

Both approaches are significant improvements over the current behavior. Please let me know which direction you prefer. I am happy to update the PR to implement either solution.

alberti42:
For your information, I committed a PR for tmux. I added support for the relevant queries Yazi needs and more. When this [PR](https://github.com/tmux/tmux/pull/4614) is integrated upstream, there will be no need to differentiate between tmux and other terminal emulators—or at least this is the goal, making tmux behave as any other terminal emulator.

I must say, I am quite impressed with how fast `yazi` runs. The integration with tmux is also super fast in generating previews for images. It now takes only a bunch of ms to create a preview image.

sxyazi:
Awesome, thanks for the update!

This seems like the best "Option C" here - making tmux properly support these sequences, allowing Yazi to interpret terminal behavior consistently, just like it does with other terminal emulators, which means Yazi no longer needs to include tmux-specific hacks, and other TUI applications like NeoVim should also benefit, since they rely on these sequences as well. Great job @alberti42!

In this case, I'll go ahead and close this PR in favor of https://github.com/tmux/tmux/pull/4614

alberti42:
My patch to tmux has been accepted. The 'CSI u' query has been removed from the submitted pull request (PR) for now because it is not yet a standard protocol. In the next few days, I will file a separate PR for `tmux` to implement the 'CSI u' query.

This means that you will no longer need to use the 'allow-passthrough' option, which was certainly the wrong approach for handling the cursor. Regarding 'CSI u', I am not yet 100% sure whether we will no longer need it. I have almost no experience with extended CSI u.

I am wondering whether we need to check whether the underlying terminal emulator supports CSI u, and then check whether `tmux` supports it, or whether it is sufficient to query `tmux` and disregard the underlying terminal emulator.

In any case, I am happy that everything related to cursor handling is now handled perfectly by tmux (though only in the upstream version and not yet in the official releases).

sxyazi:
Happy to see tmux has merged and now natively supports it, I've removed the hack to it in https://github.com/sxyazi/yazi, but passthrough is still needed to workaround https://github.com/sxyazi/yazi/issues/2050

> I am wondering whether we need to check whether the underlying terminal emulator supports CSI u, and then check whether `tmux` supports it, or whether it is sufficient to query `tmux` and disregard the underlying terminal emulator.


This depends on how tmux reports and handles CSI u:

- If the underlying terminal does not support CSI u and tmux reports to Yazi that it's not supported (even if tmux itself does support), then Yazi does not need to do anything extra.
- If the underlying terminal does not support CSI u, but tmux reports support, yet tmux can safely ignore the CSI u mode request sent by Yazi since the underlying terminal doesn't support CSI u, then Yazi doesn't need to do anything extra.

alberti42:
Cool. Yes, the `CSI u` is a bit trickier. It can be done and I actually had already implemented in tmux the fix, but it was decide to remove it from the PR because Nicholas (the creator of tmux) was skeptical/unfamiliar with the nonstandard query for probing `CSI u` support. I could open a new PR and see how it goes in a second attempt.

Regarding #2050, I saw the instructions to reproduce the bug. What can I do to reproduce it on my setup? What should I by-pass in Yazi's code? Is there a quick way to test the bug? I am asking this because I think instead of the workaround, we should fix tmux. Hopefully with a couple of lines of code, the problem is fixed in tmux at the source.

PR #3121, opened 2025-08-29
feat: progression status for copy I/O in Task window

### LINKED ISSUE #3106: Show Copy/Move Progress in Tasks Window
### `yazi --debug` output

```Shell
Not needed.
```

Currently, when copying or moving files, the Tasks window only shows the queued operations without any progress details. 

It would be helpful to display:

- Current file being processed
- Progress bar or percentage
- Transfer speed and estimated time remaining

This would improve user experience, especially for large file operations or slow networks.

### Additional context

_No response_

### Checklist

cattyman919:

Resolves #3106 

## Rationale of this PR
The existing task window only provides limited feedback for long-running I/O operations, particularly file copies. When a user copies a large file or directory, the UI only display a static "Copy" message in the task window, offering no insight into the progress. 

This PR introduce real-time progress for the copy file operation in the task window

## Key Enchancements
- Visual progress bar, showing percentage of completion based on total byte size.
- Task view now shows total byes copied vs. total size (e.g., 2.1 GB / 5 GB)
- When copying directory, UI now displays specific file currently being processed and the overall file count (e.g., "(15/100) Copying video.mp4").
- The progress polling interval is currently hardcoded with 1 second

### Copy File

**Before**
image

**After**
image

### Copy Directory

**Before**
image

**After**
image

## Future Work: Transfer Speed and ETA
While the initial goal of #3106 included transfer speed and ETA, this part of the feature has been deferred to a future enhancement to ensure accuracy.

The primary reason is that the current progress tracking is based on the bytes processed within the application's copy loop. This doesn't always reflect the actual disk I/O speed due to factors like OS-level disk caching and buffering. A naive speed calculation could therefore report very high speeds initially (as data is written to a fast memory buffer) that don't represent the true time it will take to flush to disk, leading to a misleading ETA.


sxyazi (approved):
Thank you!

PR #3128, opened 2025-08-31
fix: add `fd` as a dependency to the `.deb` packaging workflow

philclifford:
## Rationale of this PR

The dependency on `fd-find` causes a removal of `fd` due to a noted conflict and if the user has installed `fd` (i.e. from github or via `deb-get` ) because the renamed `fd-find` in their LTS installation is too old they'll suffer a surprise downgrade or get into a blocked situation.

```
$ apt-cache policy fd-find fd
fd-find:
  Installed: (none)
  Candidate: 8.3.1-1ubuntu0.1
  Version table:
     8.3.1-1ubuntu0.1 500
        500 http://gb.archive.ubuntu.com/ubuntu jammy-updates/universe amd64 Packages
        500 http://gb.archive.ubuntu.com/ubuntu jammy-security/universe amd64 Packages
     8.3.1-1 500
        500 http://gb.archive.ubuntu.com/ubuntu jammy/universe amd64 Packages
fd:
  Installed: 10.3.0
  Candidate: 10.3.0
  Version table:
 *** 10.3.0 100
        100 /var/lib/dpkg/status
```

It might be preferable even to limit the `fd` dependency to >='10.3.0' , but I think the simple alternative would do.

```
deb-get show fd yazi
fd
  Package:	fd
  Repository:	01-main
  Updater:	deb-get
  Installed:	10.3.0
  Published:	10.3.0
  Architecture:	amd64 arm64 armhf
  Download:	https://github.com/sharkdp/fd/releases/download/v10.3.0/fd_10.3.0_amd64.deb
  Website:	https://github.com/sharkdp/fd
  Summary:	A simple, fast and user-friendly alternative to 'find'.
Yazi
  Package:	yazi
  Repository:	99-local
  Updater:	deb-get
  Installed:	No
  Published:	25.5.28-1
  Architecture:	amd64 arm64
  Download:	https://github.com/sxyazi/yazi/releases/download/nightly/yazi-x86_64-unknown-linux-gnu.deb
  Website:	https://yazi-rs.github.io/
  Summary:	💥 Blazing fast terminal file manager written in Rust, based on async I/O.
```

I've re-opened https://github.com/wimpysworld/deb-get/issues/1158 and created a [draft PR](https://github.com/wimpysworld/deb-get/pull/1508) there ready for your next release.

### DIFF
--- a/yazi-packing/Cargo.toml
+++ b/yazi-packing/Cargo.toml
@@ -11,7 +11,7 @@ repository  = "https://github.com/sxyazi/yazi"
 	[package.metadata.deb]
 	name                      = "yazi"
 	license-file              = [ "../LICENSE", "0" ]
-	depends                   = "file, ffmpeg, 7zip, jq, poppler-utils, fd-find, ripgrep, fzf, zoxide, imagemagick, xsel|xclip|wl-clipboard"
+	depends                   = "file, ffmpeg, 7zip, jq, poppler-utils, fd-find|fd, ripgrep, fzf, zoxide, imagemagick, xsel|xclip|wl-clipboard"
 	recommends                = "bash-completion"
 	extended-description-file = "README.md"
 	section                   = "utility"


sxyazi:
Thank you for your patch!

I'm not familiar with Debian, so I have a few questions:

- What is the difference between `fd-find` and `fd`? Are they two different packages of the [`fd`](https://github.com/sharkdp/fd) project?
- The `fd` package seems to be more up-to-date, is it possible to use it directly as Yazi's dependency instead of keeping both packages?

On a personal note, thank you for including Yazi in deb-get, appreciate your work!

philclifford:
`fd` is the newer package direct from the author's github.  `fd-find` is the same software (but an older release) re-packaged by the ubuntu maintainers with the application (`fd` -> `fdfind`)  and package (`fd` -> `fd-find`) renamed due to a clash with another package (`fdclone`) which contains `/usr/bin/fd`


```
apt info fd-find fd
Package: fd-find
Version: 9.0.0-1
Built-Using: rust-nix (= 0.26.2-1), rustc (= 1.73.0+dfsg0ubuntu1-0ubuntu1)
Priority: optional
Section: universe/utils
Source: rust-fd-find
Origin: Ubuntu
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Original-Maintainer: Debian Rust Maintainers <pkg-rust-maintainers@alioth-lists.debian.net>
Bugs: https://bugs.launchpad.net/ubuntu/+filebug
Installed-Size: 2,570 kB
Depends: libc6 (>= 2.34), libgcc-s1 (>= 4.2)
Homepage: https://github.com/sharkdp/fd
[...]

Package: fd
Version: 10.3.0
Status: install ok installed
Priority: optional
Section: utils
Maintainer: David Peter <mail@david-peter.de>
Installed-Size: unknown
Provides: fd
Conflicts: fd-musl, fd-find
Homepage: https://github.com/sharkdp/fd
 
```

So it seems that if a user has installed `fd` and then installs a package that depends on `fd-find` `apt` will not consider a downgrade is  being sought ( because the package names differ) and will install the older `fd-find` and remove `fd` due to the noted conflict. If the dependency could be resolved with either `fd-find` or `fd` then no surprise downgrade should occur. Since the `fd` package is not in the ubuntu repositories you need to depend on `fd-find` but adding `| fd` will satisfy the dependency if `fd` is installed or if not will get the older `fd-find` from the repository if needed. So it is desirable to add `fd` and necessary to have both to preserve the dependency. The `fd` package adds a symlink for `fdfind` so should be a completely compatible upgrade, and `yazi` seems to be working fine for me with it installed.

I don't believe there is a way to depend directly on `fd` (from github) and if you only depended on `fd` most debian/ubuntu users would get an unsatisfiable dependency error. 

Thank **you** for making Yazi! I'm delighted to have found it through a request for `deb-get` to support it and to have been able to make a trivial PR to do that now you are publishing a `.deb`

sxyazi (approved):
I see, thank you!

Issue #2781, opened 2025-05-20
Proper multi file plugin support


Grauly:
### `yazi --debug` output

```Shell
Yazi
    Version: 25.4.8 (Nixpkgs 2025-04-08)
    Debug  : false
    Triple : x86_64-unknown-linux-gnu (linux-x86_64)
    Rustc  : 1.86.0 (05f9846f 2025-03-31)

Ya
    Version: 25.4.8 (Nixpkgs 2025-04-08)

Emulator
    TERM                : Some("xterm-kitty")
    TERM_PROGRAM        : None
    TERM_PROGRAM_VERSION: None
    Brand.from_env      : Some(Kitty)
    Emulator.detect     : Emulator { kind: Left(Kitty), light: false, cell_size: Some((9, 20)) }

Adapter
    Adapter.matches    : Kgp
    Dimension.available: WindowSize { rows: 24, columns: 78, width: 702, height: 480 }

Desktop
    XDG_SESSION_TYPE           : Some("wayland")
    WAYLAND_DISPLAY            : Some("wayland-1")
    DISPLAY                    : Some(":0")
    SWAYSOCK                   : Some("/run/user/1000/sway-ipc.1000.2379.sock")
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Variables
    SHELL           : Some("/run/current-system/sw/bin/zsh")
    EDITOR          : Some("nvim")
    VISUAL          : None
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
    FZF_DEFAULT_OPTS: None

Text Opener
    default     : Some(OpenerRule { run: "nvim \"$@\"", block: false, orphan: true, desc: "", for_: None, spread: true })
    block-create: None
    block-rename: None

Multiplexers
    TMUX               : false
    tmux version       : No such file or directory (os error 2)
    tmux build flags   : enable-sixel=Unknown
    ZELLIJ_SESSION_NAME: None
    Zellij version     : No such file or directory (os error 2)

Dependencies
    file          : 5.46
    ueberzugpp    : No such file or directory (os error 2)
    ffmpeg/ffprobe: 7.1.1 / 7.1.1
    pdftoppm      : 25.05.0
    magick        : 7.1.1-47
    fzf           : 0.62.0
[...]

PR #3154, opened 2025-09-08
feat: multi-entry support for plugin system

sxyazi:
An implementation of the feature request: https://github.com/sxyazi/yazi/issues/2781

Closes https://github.com/sxyazi/yazi/issues/2781

---

With this PR, Yazi will support Lua files other than `main.lua` as plugin entry points. To achieve this, a new dot (`.`) syntax has been introduced. For example:

```sh
~/.config/yazi/plugins/demo.yazi/
├── main.lua
├── foo.lua
├── bar.lua
├── README.md
└── LICENSE
```

Now, in the [`plugin` command](https://yazi-rs.github.io/docs/plugins/overview#functional-plugin), and fetchers, spotters, preloaders, and previewers rules, in addition to specifying `demo` (which corresponds to `main.lua`) as the entry point as before:

```toml
[[plugin.prepend_previewers]]
mime = "text/*"
run  = "demo"
```

You can also explicitly specify other Lua scripts as entry points, such as `foo.lua`:

```toml
[[plugin.prepend_previewers]]
mime = "text/*"
run  = "demo.foo"
```

You can explicitly call `main.lua` with `demo.main` for sure, but internally this will be normalized to just `demo`:

```toml
[[plugin.prepend_previewers]]
mime = "text/*"
run  = "demo.main"
```

For the `require()` function, the new dot (`.`) syntax is also supported, which means you can have:

```lua
-- ~/.config/yazi/plugins/demo.yazi/foo.lua

local bar = require("demo.bar")  -- Loads the `bar.lua` entry of the `demo` plugin
```

`require()` also supports relative paths, which means the above code can also be written as:

```lua
-- ~/.config/yazi/plugins/demo.yazi/foo.lua

local bar = require(".bar")  -- Loads the `bar.lua` entry of the current plugin (`demo`)
```

### DIFF
--- a/yazi-cli/src/args.rs
+++ b/yazi-cli/src/args.rs
@@ -24,9 +24,6 @@ pub(super) enum Command {
 	/// Manage packages.
 	#[command(subcommand)]
 	Pkg(CommandPkg),
-	#[command(hide = true)]
-	/// Manage packages.
-	Pack(CommandPack), // TODO: remove
 	/// Publish a message to the current instance.
 	Pub(CommandPub),
 	/// Publish a message to the specified instance.
@@ -81,26 +78,6 @@ pub(super) enum CommandPkg {
 		#[arg(index = 1, num_args = 0..)]
 		ids: Vec<String>,
 	},
-}
-
-#[derive(clap::Args)]
-#[command(arg_required_else_help = true)]
-pub(super) struct CommandPack {
-	/// Add packages.
-	#[arg(short = 'a', long, num_args = 1..)]
-	pub(super) add:     Option<Vec<String>>,
-	/// Delete packages.
-	#[arg(short = 'd', long, num_args = 1..)]
-	pub(super) delete:  Option<Vec<String>>,
-	/// Install all packages.
-	#[arg(short = 'i', long)]
-	pub(super) install: bool,
-	/// List all packages.
-	#[arg(short = 'l', long)]
-	pub(super) list:    bool,
-	/// Upgrade all packages.
-	#[arg(short = 'u', long)]
-	pub(super) upgrade: bool,
 }
 
 #[derive(clap::Args)]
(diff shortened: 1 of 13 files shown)


hankertrix:
Does this work with directories too? For example, `require(".utils.bar")`, with the plugin directory containing a directory called `utils`, and a Lua file called `bar.lua`.

sxyazi:
Subdirectories are not supported because that would require coming up with
another notation besides the dot (`.`) to express the parent directory, and
something like `local foo = require("..foo")` looks a bit weird

hankertrix:
Ah I see. As an alternative to the notation for the parent directory, I think it’s possible to just set the base directory for the plugin as the plugin directory, then all imports must be relative to that base directory. For example, if a file `plug.yazi/utils/archive/bar.lua` wants to use a function from `plug.yazi/foo.lua`, the require would be `require(“.foo”)`, and if `plug.yazi/baz.lua` wants to use a function from `plug.yazi/utils/archive/bar.lua`, the require would be `require(“.utils.archive.bar”)`.

This is how Neovim does plugins.

Either way, love this change and I can’t wait to split up my plugin code into separate files. Makes it far easier to maintain. But I think the package manager might need to be updated to support pulling down all the Lua files as it currently only pulls down the `main.lua` file.

sxyazi:
I'm a bit surprised that this is how Neovim works, because this is different from the typical expectation of "dot" (`.`) being relative to the current directory. In this implementation, the "current directory" is the same as the "plugin root", so "." can refer to both, but in my opinion, their semantics are different.

Nice, glad it works for you! This PR added multi-entry support for the package manager, as well!

hankertrix:
Oh, Neovim doesn’t use the “dot” syntax. It doesn’t have the “.” in front of requires, just the file name without “.lua”. The main thing is that all `require`s are relative to the plugin’s root directory instead of doing current directory and parent directory. But yes, I agree that it can be confusing since “.” usually refers to the current directory.

That’s awesome! Looking forward to it!

Issue #3197, opened 2025-09-23
Plugin revision pinning is ignored upon install


philocalyst:
### What system are you running Yazi on?

macOS

### What terminal are you running Yazi in?

Ghostty 1.1.3

### `yazi --debug` output

```Shell
Yazi
    Version: 25.9.15 (9a804bac 2025-09-23)
    Debug  : true
    Triple : aarch64-apple-darwin (macos-aarch64)
    Rustc  : 1.88.0 (6b00bc38 2025-06-23)

Ya
    Version: 25.5.31 (Nixpkgs 2025-05-30)

Emulator
    TERM                : Some("xterm-256color")
    TERM_PROGRAM        : Some("ghostty")
    TERM_PROGRAM_VERSION: Some("1.1.3")
    Brand.from_env      : Some(Ghostty)
    Emulator.detect     : Emulator { kind: Left(Ghostty), light: false, csi_16t: (27, 56), force_16t: false }

Adapter
    Adapter.matches    : Kgp
    Dimension.available: Dimension { rows: 32, columns: 111, width: 3016, height: 1804 }

Desktop
    XDG_SESSION_TYPE           : None
    WAYLAND_DISPLAY            : None
    DISPLAY                    : None
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Variables
    SHELL           : Some("/bin/zsh")
    EDITOR          : Some("hx")
    VISUAL          : None
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
    FZF_DEFAULT_OPTS: None

Text Opener
    default     : Some(OpenerRule { run: "$EDITOR \"$@\"", block: true, orphan: false, desc: "$EDITOR", for: None, spread: true })
    block-create: Some(OpenerRule { run: "$EDITOR \"$@\"", block: true, orphan: false, desc: "$EDITOR", for: None, spread: true })
    block-rename: Some(OpenerRule { run: "$EDITOR \"$@\"", block: true, orphan: false, desc: "$EDITOR", for: None, spread: true })

Multiplexers
    TMUX               : false
    tmux version       : tmux 3.5a
    tmux build flags   : enable-sixel=Unknown
    ZELLIJ_SESSION_NAME: None
    Zellij version     : 0.43.1

[...]

PR #3200, opened 2025-09-24
fix: renew package `rev` only when it's empty

sxyazi:
Fixes https://github.com/sxyazi/yazi/issues/3197

### DIFF
--- a/yazi-cli/src/package/install.rs
+++ b/yazi-cli/src/package/install.rs
@@ -19,7 +19,10 @@ impl Dependency {
 		}
 
 		self.deploy().await?;
-		self.rev = Git::revision(&path).await?;
+		if self.rev.is_empty() {
+			self.rev = Git::revision(&path).await?;
+		}
+
 		Ok(())
 	}
 }


philocalyst:
Wouldn't a write still be attempted? This stops the pin from being overwritten; but yazi will still attempt to write to the file even if the contents haven't been changed (Just tested on nightly).

My guess is that it would be trivial to derive equality or partial equality on the data structure you're using internally to represent the package file, and only write when it's different than the old. I could submit a PR for this.

This is useful mainly for dealing with setups (like those encouraged by nix) that are designed to be immutable. But also should make Yazi more reliable.

sxyazi:
Perhaps it's possible to check for manifest differences before writing to
the file, but I'm not sure if it really makes sense, because there's no
guarantee of immutability at the file content or format level on `install`,
which means if new fields that need to change druing `install` are
introduced in the future, it could break this check, which makes it pretty
fragile

philocalyst:
I suppose you're right, but as of _right now_, those fields don't exist, and yazi will fail on an immutable package.toml even if it has _nothing to change_. Is this really the desired behavior? Couldn't yazi just throw a WARN? 
As in:
WARNING: Couldn't write to package.toml, updates may not be saved.

sxyazi:
Yes, that's by design, `ya pkg` expects `package.toml` (as well as the
`plugins` and `flavors` directories) to be readable and writable, and any
write failure should be considered a fatal error as `ya pkg` is not
supposed to work on a read-only filesystem.

If you need this level of isolation with Nix, using Nix's own package
manager to manage Yazi itself along with its plugins is a better way to
achieve immutability - they are already available in nixpkgs:
https://search.nixos.org/packages?channel=unstable&query=yazi

That said, if you really think it's worth the effort, I can accept a PR to
check if writing can be avoided before writing, but this would be
best-effort and implementation-dependent - I can't guarantee it will always
work as future iterations evolve.

philocalyst:
It's not worth the effort. (I didn't know about those on nixpkgs)

Thank you for the kind and informed responses.

PR #3235, opened 2025-10-09
feat: new `casefy` command for case conversion

### LINKED ISSUE #3233: "u" and "U" keybindings for vim input in visual mode
### `yazi --debug` output

```Shell

Yazi
    Version: 25.5.31 (Arch Linux 2025-07-23)
    Debug  : false
    Rustc  : 1.88.0 (6b00bc38 2025-06-23)

Ya
    Version: 25.5.31 (Arch Linux 2025-07-23)

Emulator
    TERM                : Some("xterm-ghostty")
    TERM_PROGRAM        : Some("ghostty")
    TERM_PROGRAM_VERSION: Some("1.2.0-arch1")
    Brand.from_env      : Some(Ghostty)
    Emulator.detect     : Emulator { kind: Left(Ghostty), light: false, csi_16t: (14, 31), force_16t: false }

Adapter
    Adapter.matches    : Kgp
    Dimension.available: Dimension { rows: 33, columns: 137, width: 1918, height: 1023 }

Desktop
    XDG_SESSION_TYPE           : Some("x11")

SSH

WSL
    WSL: false

Variables
    SHELL           : Some("/bin/bash")
    EDITOR          : Some("nvim")
    VISUAL          : None
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
    FZF_DEFAULT_OPTS: None

Text Opener
    default     : Some(OpenerRule { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for: None, spread: true })
    block-create: Some(OpenerRule { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for: None, spread: true })
    block-rename: Some(OpenerRule { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for: None, spread: true })

Multiplexers
    TMUX               : false
    ZELLIJ_SESSION_NAME: None
    Zellij version     : 0.43.1

Dependencies
    file          : 5.46
    ffmpeg/ffprobe: 8.0 / 8.0
    pdftoppm      : 25.09.1
    magick        : 7.1.2-3
    fzf           : 0.65.2
    fd/fdfind     : 10.3.0 / No such file or directory (os error 2)
    rg            : 14.1.1
    zoxide        : No such file or directory (os error 2)
    7zz/7z        : No such file or directory (os error 2) / No such file or directory (os error 2)
    jq            : 1.8.1

Clipboard
    xclip        : 0.13
    xsel         : 1.2.1

Routine


```

Missing vim keybinds for inputs :
- `u` in visual mode to make highlighted text lowercase
- `U` in visual mode to make highlighted text uppercase

I often try to use this motion and then realize it doesn’t work in Yazi. Having it available would really help me organize my files more efficiently.

Implement the missing vim actions for "u" and "U" in visual mode, and add corresponding keybindings.

### Additional context

Similar to https://github.com/sxyazi/yazi/issues/770 and https://github.com/sxyazi/yazi/issues/2119

### Checklist

lmnek:
Resolves #3233

I created this as a draft PR because I encountered an issue while adding the `lowercase` and `uppercase` actions. The problem is that `u` is already bound to `undo`, and the current Yazi vim emulation doesn't differentiate between keybindings across different modes (normal and visual). Therefore, I can't assign `u` to `lowercase`, as the two would conflict and only one would function properly.

How do you think this issue should be addressed?

Potential solutions I see are:
- Adding a property to each keybinding in `keymap-default.toml` to specify which modes it should be active in, but that would require more fundamental structural changes to the vim input system.
- Adding a visual mode (currently, there are only normal, insert, and replace modes), and defining the lowercase and uppercase actions only for the visual mode in `command.rs`, while keeping undo solely in normal mode.


sxyazi:
Thank you for the patch!

> The problem is that `u` is already bound to `undo`, and the current Yazi vim emulation doesn't differentiate between keybindings across different modes (normal and visual)

That's a great question - appreciate you bringing it up. Here are some possible solutions I can think of:

- By default, do not bind the new `lowercase` command and let users decide whether to use it and what key they want to use
- By default, choose a non-conflicting key, such as `g` => `u` instead of just `u`
- Bind both `undo` and `lowercase` to `u`, and check the current mode in the `undo` and `lowercase` commands - `undo` should only trigger in normal mode, and `lowercase` only in visual mode, so that only one command is triggered at any time

> Adding a visual mode (currently, there are only normal, insert, and replace modes)

Actually, there is a visual mode, and the `visual` command is used to switch to it:

https://github.com/sxyazi/yazi/blob/554cb52cc581df9a41e0778ebd448925cd3aca55/yazi-widgets/src/input/commands/visual.rs#L9

which binds to the `v` key by default:

https://github.com/sxyazi/yazi/blob/554cb52cc581df9a41e0778ebd448925cd3aca55/yazi-config/preset/keymap-default.toml#L240

lmnek:
I worked around the issue by changing the keybindings for lowercase and uppercase to `gu` and `gU`. This works fine functionally, but it’s not the best user experience since it differs from the classic vim behavior. The only reason for this change is the difficulty of implementing proper mode-aware keybinding handling. Adjusting Yazi’s vim input architecture is out of scope for this PR, so I’ll keep it like this for now :c. Please feel free to start the review!

---

**More context on the encountered issues**: 

Currently, it’s not possible to support multiple bindings on the same key, as duplicates are  removed in a HashSet ([`/yazi-config/src/keymap/rules.rs`](https://github.com/sxyazi/yazi/blob/main/yazi-config/src/keymap/rules.rs)) . As a result, any additional bindings using the same key are overwritten or dropped.

Neovim addresses this by allowing keymaps to specify a list of modes in which a binding is active (see [`vim.keymap.set()`](https://neovim.io/doc/user/lua.html#vim.keymap.set%28%29)), enabling the same key to perform different actions in Normal, Visual, or Insert mode. Introducing a similar mode-aware keybinding system in Yazi could be a clean long-term solution and would make the keymap configuration more expressive and flexible beyond this specific case.


> Actually, there is a visual mode, and the visual command is used to switch to it

There’s still one more problem related to this: visual mode isn’t currently represented in the `InputMode` enum (only in `InputOp`), which means vim-style actions for it specifically can’t be added cleanly in the [`commands.rs`](https://github.com/sxyazi/yazi/blob/main/yazi-widgets/src/input/commands/commands.rs) switch. A simple (but somewhat hacky) solution for this exists in the [`change-case-same-key`](https://github.com/lmnek/yazi/commit/da149c57f995ade17de0a4fb19e570768fcd4d61) branch, which avoids larger architectural changes while keeping the functionality working.

sxyazi:
Thank you for your update!

> Currently, it's not possible to support multiple bindings on the same key

This is possible - we can bind two commands to "u" at the same time:

```
{ on = "u", run = ["undo", "lowercase"] }
```

Both commands will be ran simultaneously, but we can check the current mode in each command: `undo` can check if it's not normal mode and do nothing if so; `lowercase` checks for visual mode.

This should be a simple solution that doesn't require fundamental architectural changes, and it should also address the second issue you mentioned. What do you think?

lmnek:
Okay, that's amazing, I completely missed the multiple keybindings feature. That makes things much simpler. Thanks for the input!

I’ve force-pushed a solution that uses “u” for both commands.

PR #3267, opened 2025-10-20
fix: make `nix-shell` use the nightly Rust toolchain

carlosedp:
This pull request updates the Nix build configuration to use the Rust nightly toolchain instead of stable, and makes minor formatting and structural improvements for clarity and maintainability.

**Toolchain updates:**

* Switched from `rust-bin.stable.latest.default` to `rust-bin.nightly.latest.default` for both the main build (`flake.nix`) and the development shell (`nix/shell.nix`), ensuring the project uses the latest nightly Rust features. [[1]](diffhunk://#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0L11-R48) [[2]](diffhunk://#diff-dd972f906c9914eb70fae1db9cf66baa653e6b643bbdedeaa0070939abc3fb56L1-R6)

**Formatting and structure:**

* Improved code formatting and grouping in both `flake.nix` and `nix/shell.nix` for better readability and maintainability. [[1]](diffhunk://#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0L11-R48) [[2]](diffhunk://#diff-dd972f906c9914eb70fae1db9cf66baa653e6b643bbdedeaa0070939abc3fb56L1-R6).

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -23,7 +23,7 @@
           inherit system;
           overlays = [ rust-overlay.overlays.default ];
         };
-        toolchain = pkgs.rust-bin.nightly.latest.default;
+        toolchain = pkgs.rust-bin.stable.latest.default;
         rustPlatform = pkgs.makeRustPlatform {
           cargo = toolchain;
           rustc = toolchain;


nix/shell.nix
```diff
@@ -1,20 +1,9 @@
-{
```
sxyazi, line 1: Which formatter are you using to format the file? Yazi uses `nixfmt-rfc-style` which follows the Nix RFC standard, see https://github.com/sxyazi/yazi/pull/1376

flake.nix
```diff
@@ -8,58 +8,44 @@
     };
```
sxyazi, line 8: `flake.nix` is user-facing and it should not use the nightly Rust toolchain, see https://github.com/sxyazi/yazi/pull/314#issuecomment-1776731390

nix/shell.nix
carlosedp: Oh I had nixfmt-classic installed. Replaced it with nixfmt-rfc-style.
Thanks for the tip!

flake.nix
carlosedp: I've realized we'd need the nightly since it's what rustfmt uses with the `+nightly` flag and since I load my flakes automatically using direnv, it would make the correct rustfmt (nightly) readily available in the shell.

sxyazi: `flake.nix` is not supposed to be used for development purposes, but `shell.nix` does

nix/shell.nix
sxyazi: Please re-format both files with `nixfmt-rfc-style`

carlosedp:
Done

flake.nix
carlosedp: Sorry to ask but what is the purpose of flake.nix? I don't understand the idea of user-facing... what are the user actions in the codebase that it needs a flake?

sxyazi: flake.nix is used to distribute the tip version, https://yazi-rs.github.io/docs/installation#flakes 

Apart from being built from the latest Yazi source code, everything else should be kept as consistent as possible with the yazi package in nixpkgs.

carlosedp: Ah, got it... so I'll just keep the shell.nix with the nightly toolchain.
Thanks for the clarification

sxyazi (approved):
LGTM, thank you!

PR #3336, opened 2025-11-18
Fix `copy name_without_ext` trimming folder names with dots 

### LINKED ISSUE #3335: `copy name_without_ext` trims folder names containing dots

Windows

wezterm

### `yazi --debug` output

```Shell
Yazi
    Version: 25.9.15 (bb9e1e97 2025-11-18)
    Debug  : false
    Triple : x86_64-pc-windows-msvc (windows-x86_64)
    Rustc  : 1.93.0-nightly (82ae0ee6 2025-10-31)

Ya
    Version: 25.9.15 (bb9e1e97 2025-11-18)

Emulator
    TERM                : Some("xterm-256color")
    TERM_PROGRAM        : Some("WezTerm")
    TERM_PROGRAM_VERSION: Some("20250308-192419-dd546c9d")
    Brand.from_env      : Some(WezTerm)
    Emulator.detect     : Emulator { kind: Left(WezTerm), light: false, csi_16t: (12, 25), force_16t: true }

Adapter
    Adapter.matches    : Iip
    Dimension.available: Dimension { rows: 52, columns: 209, width: 0, height: 0 }

Desktop

SSH

WSL
    WSL: false

Variables
    SHELL              : None
    EDITOR             : None
    VISUAL             : None
    YAZI_FILE_ONE      : Some("C:\\Program Files\\Git\\usr\\bin\\file.exe")
    YAZI_CONFIG_HOME   : None
    YAZI_ZOXIDE_OPTS   : None
    FZF_DEFAULT_OPTS   : None
    FZF_DEFAULT_COMMAND: None

Text Opener
    default     : Some(OpenerRule { run: "hx %s", block: true, orphan: false, desc: "helix", for: None, spread: true })
    block-create: Some(OpenerRule { run: "hx %s", block: true, orphan: false, desc: "helix", for: None, spread: true })
    block-rename: Some(OpenerRule { run: "hx %s", block: true, orphan: false, desc: "helix", for: None, spread: true })

Multiplexers
    TMUX               : false
    tmux version       : program not found
    ZELLIJ_SESSION_NAME: None
    Zellij version     : program not found

Dependencies
    file          : 5.46
    ueberzugpp    : program not found
    ffmpeg/ffprobe: program not found / program not found
    pdftoppm      : 25.01.0
    magick        : 7.1.2-8
    fzf           : 0.66.0
    fd/fdfind     : 10.3.0 / program not found
    rg            : 15.1.0
    chafa         : program not found
    zoxide        : 0.9.8
    7zz/7z        : program not found / 25.01
    resvg         : 0.45.1
    jq            : 1.8.1

Clipboard
    wl-copy/paste: program not found / program not found
    xclip        : program not found
    xsel         : program not found

Routine


```

### Describe the bug

I ran into an issue where pressing `c + n` (`copy name_without_ext`) on folders that have a dot in their name (for example: `Input data - 18.11.2025`) copies only the part before the dot — so it becomes `Input data - 18.11`, and the rest gets trimmed. This was quite confusing 🙂.

### Minimal reproducer

1. Create a folder with a dot in the name, for example: **Input data - 18.11.2025**
2. Select it in Yazi
3. Press: `c + n`

### Anything else?

_No response_

### Checklist

- [x] I can reproduce it after disabling all custom configs/plugins (`mv ~/.config/yazi ~/.config/yazi-backup`)

JustForFun88:
Fixes #3335

### DIFF
--- a/yazi-actor/src/mgr/copy.rs
+++ b/yazi-actor/src/mgr/copy.rs
@@ -39,7 +39,8 @@ impl Actor for Copy {
 					s.extend_from_slice(&opt.separator.transform(&u.name().unwrap_or_default()));
 				}
 				"name_without_ext" => {
-					s.extend_from_slice(&opt.separator.transform(&u.stem().unwrap_or_default()));
+					let name = if u.as_local().is_some_and(|p| p.is_dir()) { u.name() } else { u.stem() };
+					s.extend_from_slice(&opt.separator.transform(&name.unwrap_or_default()));
 				}
 				_ => bail!("Unknown copy type: {}", opt.r#type),
 			};


sxyazi:
Thank you for the patch, but this is not the right place to fix the issue - `stem()` is just a wrapper around Rust's `file_stem()`, and it should be behave the exact same way, also calling `is_dir()` on a `&Path` will involve a IO operation, but `stem()` should not do any IO (or even allocations). Instead, fix the issue at the upper-level `copy` command

JustForFun88:
> Thank you for the patch, but this is not the right place to fix the issue - `stem()` is just a wrapper around Rust's `file_stem()`, and it should be behave the exact same way, also calling `is_dir()` on a `&Path` will involve a IO operation, but `stem()` should not do any IO (or even allocations). Instead, fix the issue at the upper-level `copy` command

Done

yazi-actor/src/mgr/copy.rs
```diff
@@ -39,7 +39,8 @@ impl Actor for Copy {
 					s.extend_from_slice(&opt.separator.transform(&u.name().unwrap_or_default()));
 				}
 				"name_without_ext" => {
-					s.extend_from_slice(&opt.separator.transform(&u.stem().unwrap_or_default()));
+					let name = if u.as_local().is_some_and(|p| p.is_dir()) { u.name() } else { u.stem() };
```
sxyazi, line 42: Don't use `as_local()` since that only supports local files, whereas `copy` command should be cross-filesystem and support non-local filesystems.

Also, don't do any sync I/O on the main thread, since it can be slow, especially when the user selects many files or the files reside on a remote filesystem mount. Instead, detach those operations to an async blocking thread.

Currently the [file selection state](https://github.com/sxyazi/yazi/blob/d3ad1c0920719f0779ac0f345d8477f1000cad4e/yazi-core/src/tab/tab.rs#L21) only stores the file URL, maybe we could also store a flag indicating whether the file is a directory, which would introduce new cache-invalidation problems though. Maybe the current implementation of real-time I/O isn't a big issue as long as the user doesn't do something crazy...

sxyazi:
I'm going to close this because it still needs some work and hasn't seen any activity recently. @JustForFun88 thanks for the effort!

PR #3358, opened 2025-11-23
Support 8-bit images in RGB, CIELAB, and GRAY color spaces

leiserfg:

Currently yazi can't correctly preview images that are not srgb, this PRs transforms the colorspace while making the cache entry so they work fine.

## Rationale of this PR

It reads the icc of the image, checks if it's not srgb and if that's the case, makes it into srgb before writing to the cache folder.

This is an example of an image that works fine after the change and didn't before:

P1060211

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/):
 ### Added
 
 - Support VFS for preset previewers that rely on external commands ([#3477])
-- Support 8-bit images in CMYK, CIELAB, and GRAY color spaces ([#3358])
+- Support 8-bit images in RGB, CIELAB, and GRAY color spaces ([#3358])
 
 ### Fixed
 
--- a/yazi-adapter/src/icc.rs
+++ b/yazi-adapter/src/icc.rs
@@ -1,12 +1,6 @@
 use anyhow::Context;
-use image::{
-	ColorType, DynamicImage, GrayAlphaImage, GrayImage, ImageDecoder, RgbImage, RgbaImage,
-	metadata::Cicp,
-};
-use moxcms::{
-	CicpColorPrimaries, ColorProfile, DataColorSpace, Layout, TransferCharacteristics,
-	TransformOptions,
-};
+use image::{ColorType, DynamicImage, GrayAlphaImage, GrayImage, ImageDecoder, RgbImage, RgbaImage, metadata::Cicp};
+use moxcms::{CicpColorPrimaries, ColorProfile, DataColorSpace, Layout, TransferCharacteristics, TransformOptions};
 
 pub(super) struct Icc;
 impl Icc {
@@ -63,10 +57,13 @@ impl Icc {
 	}
 
 	fn requires_transform(profile: &ColorProfile) -> bool {
-		!(profile.color_space == DataColorSpace::Cmyk
-			|| profile.cicp.is_some_and(|c| {
-				c.color_primaries == CicpColorPrimaries::Bt709
-					&& c.transfer_characteristics == TransferCharacteristics::Srgb
-			}))
+		if profile.color_space == DataColorSpace::Cmyk {
+			return false;
+		}
+
+		profile.cicp.is_none_or(|c| {
+			c.color_primaries != CicpColorPrimaries::Bt709
+				|| c.transfer_characteristics != TransferCharacteristics::Srgb
+		})
 	}
 }


leiserfg:
By the way, it was a headache to deal with jemalloc in nix-shell, no idea why it does not compile. I had to disable it while making the PR and then enable it again before pushing.
Edit: added another commit to solve for that.

sxyazi:
This change looks reasonable to me, thank you for the patch!

I don't know Nix well, so would like someone else to review the Nix-related change some, would you like to pull it out to a separate PR so it won't block this one?

yazi-adapter/src/image.rs
```diff
@@ -12,7 +13,7 @@ pub struct Image;
 
 impl Image {
 	pub async fn precache(src: PathBuf, cache: &Path) -> Result<()> {
-		let (mut img, orientation, icc) = Self::decode_from(src).await?;
+		let (mut img, orientation, icc) = Self::decode_from(src.clone()).await?;
```
sxyazi:
```suggestion
		let (mut img, orientation, icc) = Self::decode_from(src).await?;
```

yazi-adapter/src/image.rs
```diff
@@ -31,11 +32,25 @@ impl Image {
 				encoder.write_image(&rgba, rgba.width(), rgba.height(), ExtendedColorType::Rgba8)?;
 			} else {
 				let mut encoder = JpegEncoder::new_with_quality(&mut buf, YAZI.preview.image_quality);
-				icc.map(|b| encoder.set_icc_profile(b));
-				encoder.encode_image(&img.into_rgb8())?;
+
+				let mut data = img.to_rgb8().into_raw();
```
sxyazi:
```suggestion
				let mut data = img.into_rgb8().into_raw();
```

yazi-adapter/src/image.rs
```diff
@@ -31,11 +32,25 @@ impl Image {
...
+					&& let Some(input) = Profile::new_from_slice(icc_profile, false)
+					&& !input.is_sRGB()
+				{
+					let mut output = Profile::new_sRGB();
+					output.precache_output_transform();
+
+					let xfm = Transform::new(&input, &output, qcms::DataType::RGB8, qcms::Intent::default())
+						.context("Failed creating the ICC transformer")?;
+					xfm.apply(&mut data);
+				}
```
sxyazi: Please pull these out into a new function called `apply_profile`

leiserfg:
> would you like to pull it out to a separate PR so it won't block this one?

The thing is that without this change I can't build in the nix shell, cargo fails to build jemalloc. The change is only to tell jemalloc-sys where to find the pre built one.

sxyazi:
That sounds like a problem specific to your environment since our CI [passed the build](https://github.com/sxyazi/yazi/actions/runs/19625150099) and other people seem not to have reported similar issues.

CC @XYenon who is the maintainer of Yazi's Nix package - do you think that change makes sense?

leiserfg:
> That sounds like a problem specific to your environment since our CI

In CI it uses `nix build` which does work fine cause it's replacing rust-jemalloc-sys with the one ni nixpkgs, same thing  I did here for nix-shell. Still if you don't like it I can remove it after you approve the other changes, or move it to another pr.

XYenon:
> By the way, it was a headache to deal with jemalloc in nix-shell, no idea why it does not compile. I had to disable it while making the PR and then enable it again before pushing. 

What error message occurred on which platform?

leiserfg:
> > By the way, it was a headache to deal with jemalloc in nix-shell, no idea why it does not compile. I had to disable it while making the PR and then enable it again before pushing.
> 
> What error message occurred on which platform?

In nix develop / nix-shell / direnv, `cargo build` triggers a compilation of jemalloc and that breaks (the `configure` step cant figure out the type of stderr_r). I don't think solving that in this project is worthy given that nixpkgs ships a jemalloc pre-built. 

I'm in nixos btw.

XYenon:
I found the issue because `shell.nix` references `yazi.nix`, and `yazi.nix` is a wrapper created by directly referencing the path of the `yazi-unwrapped` package. Nix cannot infer the dependencies that should be introduced by the `yazi-unwrapped` package through the wrapper. I will submit a fix shortly.

leiserfg:
Ok, then I will rebase the PR and remove the commit for the jemalloc thingy as soon as the one from @XYenon  gets merged.

yazi-adapter/src/image.rs
```diff
@@ -31,17 +32,36 @@ impl Image {
 				encoder.write_image(&rgba, rgba.width(), rgba.height(), ExtendedColorType::Rgba8)?;
```
sxyazi: Should the same process of sRGB to RGB(A) conversion be applied to the PNG encoder as well for transparent images?

yazi-adapter/src/image.rs
```diff
@@ -31,17 +32,36 @@ impl Image {
...
 			} else {
 				let mut encoder = JpegEncoder::new_with_quality(&mut buf, YAZI.preview.image_quality);
-				icc.map(|b| encoder.set_icc_profile(b));
-				encoder.encode_image(&img.into_rgb8())?;
+				let width = img.width();
+				let height = img.height();
+				let mut data = img.into_rgb8().into_raw();
+				Self::apply_profile(icc, &mut data)?;
```
sxyazi: I noticed that `apply_profile` only applied to the `precache` function, which means it will only work if the user doesn't disable the [`image` preloader](https://github.com/sxyazi/yazi/blob/2d55c9427dd67a359cd1e9908303bbf6156053d1/yazi-config/preset/yazi-default.toml#L124) and the image has [generated a cache](https://github.com/sxyazi/yazi/blob/2d55c9427dd67a359cd1e9908303bbf6156053d1/yazi-plugin/preset/plugins/image.lua#L23) at the time, which is problematic as image preview should behave the same way regardless of whether it has cached, i.e. it should also be implemented for [`downscale`](https://github.com/sxyazi/yazi/blob/2d55c9427dd67a359cd1e9908303bbf6156053d1/yazi-adapter/src/image.rs#L64) which is directlly called by [`image_show`](https://github.com/sxyazi/yazi/blob/2d55c9427dd67a359cd1e9908303bbf6156053d1/yazi-plugin/preset/plugins/image.lua#L11) (instead of `image_precache`), maybe this process should placed under [`decode_from`](https://github.com/sxyazi/yazi/blob/2d55c9427dd67a359cd1e9908303bbf6156053d1/yazi-adapter/src/image.rs#L118) which `precache` and `downscale` both call

yazi-adapter/src/image.rs
leiserfg: I didn't do it there as I don't know of a case where that affects images with alpha, therefore I don't have how to test it.

leiserfg: Ok, I think I found some, I'll give it a try.

leiserfg:
I don't like the `_decode` name either, so let me know if you have any alternative.

leiserfg:
If this PR is too convoluted, it will work for me as well if there was a way to use icat (the kitty kitten) for rendering the original images, as it handles icc profiles.

sxyazi:
Hi @leiserfg, I did some refactoring:

- Replaced the `qcms` crate with `moxcms`, which is more actively maintained and [faster](https://github.com/awxkee/moxcms?tab=readme-ov-file#icc-transform-8-bit) and is already in the dep tree.
- Moved `_decode()` into its own module `icc`.

Could you please review and see if there are any issues? If everything looks good, we're good to merge.

leiserfg:
Now that you are using moxcms you can handle also cmky (and other inked layouts), currently it does not work, you can try with https://upload.wikimedia.org/wikipedia/en/2/25/Channel_digital_image_CMYK_color.jpg

leiserfg:
Maybe is fine to merge it as is and work on inks later tho (just make sure not to lie in the changelog :smile: )

sxyazi:
Nice catch!

Do you know where the code is wrong? I can't seem to find any obvious errors.

leiserfg:
Well one thing is that you are using the same layout for in and out (I fixed that locally already) but still after that fix, it seems like the dynamic decoder is miss-reporting the color type.

The two layouts thingy does not matters for most of the colorspaces, but for cmyk it needs to be rgba for the input and rgb for the output.

sxyazi:
> it seems like the dynamic decoder is miss-reporting the color type.

Color type only tells if it's an Rgb8 image, but for the color space would need `profile.color_space`

> but for cmyk it needs to be rgba for the input and rgb for the output.

Interesting, is there a reason for doing so?

leiserfg:
> Interesting, is there a reason for doing so? 

cmyk has 4 components, so it has the same layout as rgba (4 int8)

but somehow Image is reporting it as a rgb image with a cmyk color_space (which makes no sence).
Maybe they are doing the conversion internally for cmyk already :thinking:

leiserfg:
That's it, for cmyk Image does de transformation but does not report it :facepalm: so we just need to skip the transformations in that case.

sxyazi:
Can you push a fix?

leiserfg:
Sure I will.

sxyazi:
Thanks for the update!

https://github.com/sxyazi/yazi/pull/3358/changes/ab01858bb2a756a4fe6a0704105c8d723572da7b no longer errors out, but the colors look a bit weird (left is the original image, right is Yazi preview):

7wlYAweQ

Do you know why this is happening?

leiserfg:
Gosh, then the transform Image does to the cymk is :hankey:. Let me se what can I do.

leiserfg:
For what I see, it's caused by the dynamic decoder, which transforms the cmyk to srgb (but does not take icc into account). So the only way to fix it is to use the specific decoder (JpegDecoder in this case), to avoid the color transforms.

leiserfg:
This image in particular has cmyk + custom icc, that's why it looks weird, I maged to replicate the colors shown in yazi by opening it in krita and changing the color_space to the default one in cmyk. 

I checked into Image code and there is no way to solve it (unless we do the reading by hand), cause they force zune_jpeg::JpegDecoder  to output RGB, so there is no way to get the raw bytes in order to apply the proper icc :face_exhaling:

IMHO we better merge this in the current state. I will later try to replace Image with Zune-Image directly (for more flexibility and better performance).

sxyazi:
Gotcha, thanks for checking! Gonna remove CMYK from the changelog for now.

sxyazi (approved):
Thank you!

PR #3365, opened 2025-11-25
feat: toggle selection using Shift-cursor

Skeeve:

Resolves no issue reported

## Rationale of this PR

Comming from Commander One or, even older, PC-Tools, I'm used to Shift-Up/Down for (de-)selecting items. T.b.h. I had to look up the documentation to find that space is the selection-key.

So I added my common keys to my keymap. 

Perhaps this method of selection is more widespread and beginners are more accustomed to it? I therefore suggest incorporating this as the default setting.

But it's okay to reject the pull request.

### DIFF
--- a/yazi-config/preset/keymap-default.toml
+++ b/yazi-config/preset/keymap-default.toml
@@ -43,7 +43,12 @@ keymap = [
 	{ on = "L", run = "forward", desc = "Forward to next directory" },
 
 	# Toggle
-	{ on = "<Space>", run = [ "toggle", "arrow next" ], desc = "Toggle the current selection state" },
+	{ on = "<Space>",  run = [ "toggle", "arrow next" ], desc = "Toggle the current selection state" },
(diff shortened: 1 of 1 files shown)


sxyazi:
Hi, `K` and `J` are already used for scrolling the preview:

https://github.com/sxyazi/yazi/blob/a1fb206a591be9a6937e606cbb85cfcd9158bc76/yazi-config/preset/keymap-default.toml#L54-L56

`Up` corresponds to `k`, so `Shift + Up` should map to `K` (i.e. `Shift + k`) for consistency, but `K` controls the preview pane rather than selecting files as the PR proposed, which feels a bit odd to me. The same applies for `Down`.

So I think changing it in the user's config instead of adding it as a default is a better fit, but thanks for the suggestion anyway!

Skeeve:
Okay! So I missed the mapping of "K" and "J". My main goal was to map the cursor selection.

But it's okay, if you don't want to make it a default.

PR #3377, opened 2025-11-28
Crosscompile

Skeeve:

Resolves  https://github.com/sxyazi/yazi/discussions/3367

## Rationale of this PR

I do not like to have multiple different development dependencies installed on my computer, so I usually use [Docker](https://www.docker.com/) to develop and build.

Here are some dockerfiles to help in cross-compiling yazi for different platforms which I use.

QNAP TS 431 with OS 4.3.6
macOS

## HINT

There are **no** changes to any existing yazi sources.
Just new files under `platform`.


sxyazi:
Thank you for your patch, but I think this would be better shared in ["Show and Tell"](https://github.com/sxyazi/yazi/discussions/categories/2-show-and-tell), or added somewhere in the documentation, perhaps published on https://yazi-rs.github.io/blog/ as a new post introducing how to cross‑compile for different platforms!

Skeeve:
> Thank you for your patch, but I think this would be better shared in ["Show and Tell"]

As I couldn't figure out how to write a blog article, I choose ["Show and Tell"](https://github.com/sxyazi/yazi/discussions/3388).

> or added somewhere in the documentation,

Feel free to do so or tell me where to add it. I will happily extend the documentation.

> perhaps published on https://yazi-rs.github.io/blog/ as a new post introducing how to cross‑compile for different platforms!

If you tell me how to, I will also do that.

sxyazi:
Awesome, thank you for putting it together! 

I think this is enough for the people who need it to find, but if you think it makes sense, feel free to raise a PR in https://github.com/yazi-rs/yazi-rs.github.io, which is the source of the docs.

Issue #2687, opened 2025-04-28
[Feature Request] Native Jump Back functionality


PoisonFrog:
### `yazi --debug` output

```Shell
Yazi
    Version: 25.4.8 (Arch Linux 2025-04-09)
    Debug  : false
    Triple : x86_64-unknown-linux-gnu (linux-x86_64)
    Rustc  : 1.86.0 (05f9846f 2025-03-31)

Ya
    Version: 25.4.8 (Arch Linux 2025-04-09)

Emulator
    TERM                : Some("xterm-kitty")
    TERM_PROGRAM        : None
    TERM_PROGRAM_VERSION: None
    Brand.from_env      : Some(Kitty)
    Emulator.detect     : Emulator { kind: Left(Kitty), light: false, cell_size: Some((9, 18)) }

Adapter
    Adapter.matches    : Kgp
    Dimension.available: WindowSize { rows: 44, columns: 127, width: 1143, height: 792 }

Desktop
    XDG_SESSION_TYPE           : Some("x11")
    WAYLAND_DISPLAY            : None
    DISPLAY                    : Some(":0.0")
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Variables
    SHELL           : Some("/usr/bin/zsh")
    EDITOR          : Some("nvim")
    VISUAL          : None
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
    FZF_DEFAULT_OPTS: None

Text Opener
    default     : Some(OpenerRule { run: "${EDITOR:=vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block-create: Some(OpenerRule { run: "${EDITOR:=vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block-rename: Some(OpenerRule { run: "${EDITOR:=vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })

Multiplexers
    TMUX               : false
    tmux version       : tmux 3.5a
    tmux build flags   : enable-sixel=Unknown
    ZELLIJ_SESSION_NAME: None
    Zellij version     : No such file or directory (os error 2)

Dependencies
    file          : 5.46
[...]

PR #3380, opened 2025-11-28
feat: new `stash` command to let users customize `back` and `forward` command behavior

sxyazi:
This PR partially reverts https://github.com/sxyazi/yazi/commit/ca96c5bd987c6c6a32df6e38c99dcf40f493035f and https://github.com/sxyazi/yazi/commit/d55e002d5a82ad7eeb816f78512cdbee576c2d31 to address https://github.com/sxyazi/yazi/discussions/2851#discussioncomment-13710380.

To fix https://github.com/sxyazi/yazi/issues/2687 I changed the granularity of the `back` and `forward` commands to only record [non-linear navigations](https://github.com/sxyazi/yazi/discussions/2851#discussioncomment-14107657) so the commands would act "smarter". 

However, this "smart" behavior increases the user's cognitive load as users need to understand what non-linear navigation is and take that into account when using `back` and `forward`, and it's also hard to explain this concept to users. Therefore this PR just reverts that change to keep the behavior [simple and stupid](https://en.wikipedia.org/wiki/KISS_principle).

This PR also adds a new `stash` command that is [called by `cd` internally](https://github.com/sxyazi/yazi/blob/0b8aaba29bb7ceda912095746e1d64514a35b0a6/yazi-actor/src/mgr/cd.rs#L56) to record the path while `cd`ing, and exposes the corresponding `ind-stash` and `relay-stash` [DDS events](https://yazi-rs.github.io/docs/dds) so advanced users can customize `back` and `forward` behavior to their needs. For example:

```lua
function handler(args)
  if args.source == "tab" or args.source == "cd" then
    return args  -- Returns `args` as is to allow stashing this time
  end
end

ps.sub("ind-stash", handler)    -- Triggered when the `stash` command is called on `cd`
ps.sub("relay-stash", handler)  -- Triggered when the `stash` command is called on interactive `cd`
```

The snippet above will only allow paths to be added to the `back` and `forward` stacks when a tab is created (initial navigation) or when the user or a plugin runs the `cd` command *directly*. Other possible `source` values:

https://github.com/sxyazi/yazi/blob/8c2a6dba9060d3b4152f8a7d88467329a531bd3f/yazi-parser/src/mgr/cd.rs#L52-L62

### DIFF
--- a/yazi-actor/src/context.rs
+++ b/yazi-actor/src/context.rs
@@ -118,13 +118,5 @@ impl<'a> Ctx<'a> {
 	pub fn tasks(&self) -> &Tasks { &self.tasks }
 
 	#[inline]
-	pub fn source(&self) -> Source {
-		use Source::*;
-		match self.source {
-			Key if self.level != 1 => Ind,
-			Emit if self.level != 1 => EmitInd,
-			Relay if self.level != 1 => RelayInd,
-			s => s,
-		}
-	}
+	pub fn source(&self) -> Source { if self.level != 1 { Source::Ind } else { self.source } }
 }
--- a/yazi-actor/src/mgr/back.rs
+++ b/yazi-actor/src/mgr/back.rs
@@ -14,7 +14,7 @@ impl Actor for Back {
 
 	fn act(cx: &mut Ctx, _: Self::Options) -> Result<Data> {
 		if let Some(u) = cx.tab_mut().backstack.shift_backward().cloned() {
-			return act!(mgr:cd, cx, (u, CdSource::Back));
+			act!(mgr:cd, cx, (u, CdSource::Back))?;
 		}
 		succ!();
 	}
--- a/yazi-actor/src/mgr/cd.rs
+++ b/yazi-actor/src/mgr/cd.rs
@@ -9,7 +9,7 @@ use yazi_fs::{File, FilesOp, path::expand_url};
 use yazi_macro::{act, err, render, succ};
 use yazi_parser::mgr::CdOpt;
 use yazi_proxy::{CmpProxy, InputProxy, MgrProxy};
-use yazi_shared::{Debounce, data::Data, errors::InputError, url::{AsUrl, UrlBuf, UrlLike}};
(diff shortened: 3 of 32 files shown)


boydaihungst:
@sxyazi Could you also add Yank, Move, Paste, Copy, and Delete sources? It would be much more useful to stash items when the user performs those actions (Ranger users will be familiar with this). Also, what is the Displace source, and how do I trigger it?

sxyazi:
I don't quite understand what you mean - how do operations like yank, move, paste, copy, and delete trigger the `cd` command? They don't change the user's CWD, nor do they call `cd` internally.

Displace is a second `cd` that happens after an initial `cd`. For example, after the first `cd sftp://my-server`, Yazi needs to translate that into an absolute path, i.e., resolve the user's login home, to do a secondary jump to a path like `sftp://my-server//root`, similar to an HTTP 301 redirect, and that second `cd` is classified as a displace

boydaihungst:
Sorry. I mean there is any way to stash the cwd, after user use one of Yank, Move, Paste, Copy, and Delete commands/keybinds? 

For example: when user in location /home/user/Downloads/folder_A they yank some files, and use leave and enter commands to move to another location /home/user/Videos/folder_B. Can they move back to folder_A with `back` command after they pasted yanked files? I see some ppl using only 1 tab, so this could be easier for them.

sxyazi:
> after user use one of Yank, Move, Paste, Copy, and Delete commands/keybinds?

You don't need to stash the path *after* these operations, since the path has already been stashed once you change the CWD, and these operations are after the CWD changes right?

> when user in location /home/user/Downloads/folder_A they yank some files, and use leave and enter commands to move to another location /home/user/Videos/folder_B. Can they move back to folder_A with `back` command after they pasted yanked files?

Yes, this is the default behavior after this PR. Note that, in your case, the source would be `leave` or `enter` since that's what the command the user actually uses to change the CWD, and it has nothing to do with yank, move, paste, copy, and delete, etc. commands

Issue #2733, opened 2025-05-06
no manpage in the flake


teto:
### `yazi --debug` output

```Shell
Yazi
    Version: 25.4.8 (db60951 2025-05-05)
    Debug  : false
    Triple : x86_64-unknown-linux-gnu (linux-x86_64)
    Rustc  : 1.86.0 (05f9846f 2025-03-31)

Ya
    Version: 25.4.8 (db60951 2025-05-05)

Emulator
    TERM                : Some("xterm-kitty")
    TERM_PROGRAM        : None
    TERM_PROGRAM_VERSION: None
    Brand.from_env      : Some(Kitty)
    Emulator.detect     : Emulator { kind: Left(Kitty), light: false, csi_16t: (11, 24), force_16t: false }

Adapter
    Adapter.matches    : Kgp
    Dimension.available: Dimension { rows: 57, columns: 116, width: 1276, height: 1368 }

Desktop
    XDG_SESSION_TYPE           : Some("wayland")
    WAYLAND_DISPLAY            : Some("wayland-1")
    DISPLAY                    : None
    SWAYSOCK                   : Some("/run/user/1000/sway-ipc.1000.2821.sock")
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Variables
    SHELL           : Some("/run/current-system/sw/bin/zsh")
    EDITOR          : Some("nvim")
    VISUAL          : None
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
    FZF_DEFAULT_OPTS: Some("--bind=\'ctrl-o:execute(code {})+abort\' --bind=\'ctrl-o:execute(code {})+abort\'")

Text Opener
    default     : Some(OpenerRule { run: "xdg-open \"$1\"", block: true, orphan: false, desc: "xdg-open", for_: None, spread: false })
    block-create: Some(OpenerRule { run: "xdg-open \"$1\"", block: true, orphan: false, desc: "xdg-open", for_: None, spread: false })
    block-rename: Some(OpenerRule { run: "xdg-open \"$1\"", block: true, orphan: false, desc: "xdg-open", for_: None, spread: false })

Multiplexers
    TMUX               : false
    tmux version       : tmux 3.5a
    tmux build flags   : enable-sixel=Unknown
[...]

PR #3417, opened 2025-12-08
feat: install manpages from flake

teto:

When using the flake, there is no manpage available (contrary to the nixpgks stable package).
Fixes https://github.com/sxyazi/yazi/issues/2733

I think it would be easier to redefine the yazi package in the flake as an override of the nixpkgs one instead of duplicating everything.
That's what we do in https://github.com/nix-community/neovim-nightly-overlay/blob/master/flake/packages/neovim.nix

Resolves #

### DIFF
--- a/nix/yazi-unwrapped.nix
+++ b/nix/yazi-unwrapped.nix
@@ -6,7 +6,7 @@
   lib,
 
   installShellFiles,
-  stdenv,
+  fetchFromGitHub,
   rust-jemalloc-sys,
 
   imagemagick,
@@ -22,7 +22,7 @@ let
     ];
   };
 in
-rustPlatform.buildRustPackage rec {
+rustPlatform.buildRustPackage (finalAttrs: {
   pname = "yazi";
   inherit version src;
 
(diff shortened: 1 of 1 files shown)


sxyazi:
Thank you for the patch! I don't know Nix well, so I'd like someone else to review it some before merging.

Maybe @XYenon who is the maintainer of Yazi's Nix package can take a look at this PR if you have the time :)

nix/yazi-unwrapped.nix
```diff
@@ -60,14 +60,27 @@ rustPlatform.buildRustPackage rec {
...
     mkdir -p $out/share/applications
     install -m644 assets/yazi.desktop $out/share/applications/
   '';
 
+  passthru.srcs = {
+    man_src = fetchFromGitHub {
+      name = "manpages"; # needed to ensure name is unique
+      owner = "yazi-rs";
+      repo = "manpages";
+      rev = "8950e968f4a1ad0b83d5836ec54a070855068dbf";
```
XYenon, line 74: How do we keep this up-to-date?

XYenon: I found that manpages has only one commit and is currently not maintained. So, it's not a big problem to not consider updates here for now.

XYenon (approved):

sxyazi:
Thanks for the review @XYenon! Merging now

teto:
ty for merge . 
@XYenon how do you feel about defining the flake by overriding nixpkgs's yazi for code simplification ?

XYenon:
> @XYenon how do you feel about defining the flake by overriding nixpkgs's yazi for code simplification ?

I think it well be good

Issue #3325, opened 2025-11-12
Background configuration shows transparently on terminals with background/transparency


carlosedp:
### What system are you running Yazi on?

Windows WSL

### What terminal are you running Yazi in?

Windows Terminal 1.23.12811

### `yazi --debug` output

```Shell
❯ yazi --debug
Found existing alias for "yazi". You should use: "fm"

Yazi
    Version: 25.9.15 (Nixpkgs 2025-05-30)
    Debug  : false
    Triple : x86_64-unknown-linux-gnu (linux-x86_64)
    Rustc  : 1.91.1 (ed61e7d7 2025-11-07)

Ya
    Version: 25.9.15 (Nixpkgs 2025-05-30)

Emulator
    TERM                : Some("xterm-256color")
    TERM_PROGRAM        : None
    TERM_PROGRAM_VERSION: None
    Brand.from_env      : None
    Emulator.detect     : Emulator { kind: Right(Unknown { kgp: false, sixel: true }), light: false, csi_16t: (10, 20), force_16t: true }

Adapter
    Adapter.matches    : Sixel
    Dimension.available: Dimension { rows: 46, columns: 96, width: 0, height: 0 }

Desktop
    XDG_SESSION_TYPE           : None
    WAYLAND_DISPLAY            : Some("wayland-0")
    DISPLAY                    : Some(":0")
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Variables
    SHELL              : Some("/home/carlosedp/.nix-profile/bin/zsh")
    EDITOR             : Some("nvim")
    VISUAL             : None
    YAZI_FILE_ONE      : None
    YAZI_CONFIG_HOME   : None
    YAZI_ZOXIDE_OPTS   : None
[...]

PR #3505, opened 2026-01-04
feat: add flat background configuration options on theme section

carlosedp:

This PR fixes https://github.com/sxyazi/yazi/issues/3325

## Rationale of this PR

With the latest config for background, one issue when using Terminals which allow transparency is that it gets hard to read the text against it.

Inspired by FZF which renders the background with a flat color, I've implemented this PR which allows the following configuration overrides over the existing theme:

```toml
[flavor]
dark  = "dracula"

[app.overall]
bg = "#282a36"
# or
[app.parent]
bg = "#123456"
# and/or
[app.current]
bg = "#123456"
# and/or
[app.preview]
bg = "#123456"
```

To set the background color to the whole app, parent dir pane, current dir pane and preview pane. All 
can be set with different colors.

The end result looks like this with only preview pane color set:

image

And this with the overall color:

image

### DIFF
--- a/yazi-fm/src/bg_render.rs
+++ b/yazi-fm/src/bg_render.rs
@@ -32,12 +32,18 @@ pub fn apply_pane_bg_no_vertical_borders(buf: &mut Buffer, pane: Rect, bg_color:
 	}
 }
 
-/// Apply background color to entire area
+/// Apply background color to entire area, excluding header and status bar
 #[inline]
 pub fn apply_overall_bg(buf: &mut Buffer, area: Rect, bg_color: Color) {
-	for y in area.top()..area.bottom() {
-		for x in area.left()..area.right() {
-			buf[(x, y)].set_bg(bg_color);
+	// Skip the header (top row) and status bar (bottom row)
+	let start_y = area.top() + 1;
+	let end_y = if area.bottom() > 0 { area.bottom() - 1 } else { area.bottom() };
+
+	if start_y < end_y {
+		for y in start_y..end_y {
+			for x in area.left()..area.right() {
+				buf[(x, y)].set_bg(bg_color);
+			}
 		}
 	}
 }


sxyazi:
Thanks for your patch! Some change requests:

- Please remove the newly added `parent`, `current`, and `preview` properties - there's no plan to support them directly since the need to set a background color for the whole UI is small (only a few people have requested it so far, which is why it wasn't implemented till recent) and setting different background colors per pane is even more niche, and that should belong in the plugin system to keep core complexity down (I want to maintain Yazi long-term, and this kind of complexity would increase maintenance cost). If you need this feature, we can look into supporting it in a separate issue.
- Please remove the `\x1b]11;{}\x1b\\` sequence - once we use the "drawing a big box" method to render the background, it will no longer be needed.

carlosedp:
Removed the individual bg color config and the sequence as well. I'll submit a separate PR so we can discuss there and keep tracking it since I already have the code for it that works.

sxyazi:
Thank you for the update! Can you run `rustfmt +nightly **/*.rs` to format the code so that it's easier to see the difference?

carlosedp:
Done!

yazi-binding/src/elements/block.rs
```diff
@@ -0,0 +1,38 @@
...
+			}
+		}
+	}
+}
+
+impl UserData for Block {
+	fn add_methods<M: mlua::UserDataMethods<Self>>(methods: &mut M) {
+		crate::impl_area_method!(methods);
+		crate::impl_style_method!(methods, style);
+		crate::impl_style_shorthands!(methods, style);
```
sxyazi, line 36: Please remove impl_style_method and impl_style_shorthands, since it doesn't actually used anywhere

yazi-fm/src/bg_render.rs
```diff
@@ -0,0 +1,17 @@
+use ratatui::{buffer::Buffer, layout::Rect, style::Color};
```
sxyazi, line 1: Why is this file needed instead of just using `ui.Block` which was introduced by this PR?

yazi-fm/src/root.rs
```diff
@@ -68,5 +69,16 @@ impl Widget for Root<'_> {
...
+		// Fill background on all cells to create an opaque background (like fzf does).
+		// This is done AFTER all rendering so text/foreground colors are already set.
+		// The background is skipped on the header (top row) and status bar (bottom
+		// row).
+		let bg_color_str = THEME.app.bg_color();
+		if !bg_color_str.is_empty() {
+			if let Ok(bg_color) = bg_color_str.parse::<ratatui::style::Color>() {
+				apply_overall_bg(buf, area, bg_color);
+			}
+		}
```
sxyazi, line 129: This should be moved to `root.lua`, and pulled out into a new component called something like `Mask`

yazi-plugin/src/theme/theme.rs
```diff
@@ -35,7 +35,6 @@ fn app() -> Composer<ComposerGet, ComposerSet> {
 		let a = &THEME.app;
 		match key {
 			b"overall" => Style::from(a.overall).into_lua(lua),
-
```
sxyazi, line 38: Please keep the original style here

sxyazi (changes requested):

yazi-plugin/src/theme/theme.rs
carlosedp: This was added by rustfmt +nighly, not manually.

carlosedp:
Regarding the other comments, the `bg_render.rs` is needed and cannot be replaced by ui.Block. I understand that while `ui.Block` from `block.rs` is an element exposed to Lua plugins for rendering blocks with background colors, `bg_render.rs` contains the logic for applying the app's overall background color while specifically excluding the header and status bar. The `apply_overall_bg` function implements visual requirements that are needed by the app's appearance, while ui.Block is a generic API tool for plugin developers—they serve fundamentally different purposes and cannot replace each other.

To move the functionality to the Lua layer, we might need to expose more details to the API making it more complex instead.

sxyazi:
What's the point of adding `ui.Block` but not actually using it? Did you
encounter any problem of using `ui.Block`?

yazi-plugin/src/theme/theme.rs
sxyazi: Are you sure? Running `rustfmt +nightly yazi-plugin/src/theme/theme.rs` on https://github.com/sxyazi/yazi/commit/3c39a326abaacb37d9ff15af7668756d60624dfa which is this PR based on, didn't add or remove any newlines for me

carlosedp: It's the rustfmt from nix toolling:

```
❯ rustfmt --version
rustfmt 1.8.0-nightly (f04e3dfc87 2025-10-19)
❯ rustfmt yazi-plugin/src/theme/theme.rs
```

Nothing changes.

carlosedp:
> What's the point of adding `ui.Block` but not actually using it? Did you encounter any problem of using `ui.Block`?

Agree, I've removed the custom Block element since we don't need to expose it to Lua.
This simplifies the implementation.

sxyazi:
No, I mean, we should use `ui.Block`...

yazi-plugin/src/theme/theme.rs
sxyazi: If `rustfmt` doesn't do that change, and you were not doing it manually either, where does it come from? I'm a bit confused

carlosedp:
I'm sorry but I won't be able to keep these PRs anymore. I've been busy with other stuff and I've spent more time into maintaining the PRs than using Yazi itself.

I'd love if someone could keep up and carry this on but for the time being, I'll merge them into my own branch and built it myself.

Hope someday these feature (the flat background and exclude files) can be upstreamed.

sxyazi:
No problem, thanks for your efforts, @carlosedp! I'll take a look at how to support it when I get some time.

Issue #1440, opened 2024-08-08
Doesnt show compressed (tar.) files content


og900aero:
### What system are you running Yazi on?

Linux X11

### What terminal are you running Yazi in?

st 0.9.1

### `yazi --debug` output

```Shell
Yazi
    Version: 0.3.0 (d2ebadb 2024-08-03)
    Debug  : false
    OS     : linux-x86_64 (unix)

Ya
    Version: 0.3.0

Emulator
    Emulator.via_env: ("st-256color", "")
    Emulator.via_csi: Ok(Unknown([]))
    Emulator.detect : Unknown([])

Adapter
    Adapter.matches: X11

Desktop
    XDG_SESSION_TYPE: Some("tty")
    WAYLAND_DISPLAY : None
    DISPLAY         : Some(":0")

SSH
    shared.in_ssh_connection: false

WSL
    /proc/sys/fs/binfmt_misc/WSLInterop: false

Variables
    SHELL              : Some("/bin/bash")
    EDITOR             : Some("micro")
    YAZI_FILE_ONE      : None
    YAZI_CONFIG_HOME   : None
    ZELLIJ_SESSION_NAME: None

Text Opener
    default: Some(Opener { run: "$EDITOR \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block  : Some(Opener { run: "$EDITOR \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })

tmux
    TMUX   : false
    Version: tmux 3.3a
Dependencies
    file             : 5.44
    ueberzugpp       : 2.9.6
    ffmpegthumbnailer: 2.2.2
    magick           : No such file or directory (os error 2)
    fzf              : 0.50.0
    fd               : No such file or directory (os error 2)
    rg               : 13.0.0
    chafa            : 1.12.4
    zoxide           : 0.4.3
    7z               : No such file or directory (os error 2)
    7zz              : 22.01
    jq               : 1.6
```

### Did you try the latest nightly build to see if the problem got fixed?

No, and I'll explain why below

### Describe the bug

When I compress files with tar,gz, tar.xz, tar.zstd, tar.bz2, then doesnt show content, only show filename.tar

yazi.toml:

```
previewers = [
	{ mime = "application/x-tar",           run = "archive" },
	{ mime = "application/x-bzip",          run = "archive" },
	{ mime = "application/x-bzip2",         run = "archive" },
[...]

PR #3518, opened 2026-01-06
feat: support compressed tarballs (`.tar.gz`, `.tar.bz2`, etc.) in the preset archive previewer

### LINKED ISSUE #1548: Make the builtin `extract` and `archive` plugins handle `*.tar` formats better

itsvyle:

Resolves #1548 (the original fix only resolved the `extract` plugin, and not the `archive` plugin)

## Rationale of this PR

The built-in preview of `.tar.*` files just shows a `.tar` file inside - this is because 7zip just decompresses the first compression layer, without extracting the tar

I acknowledge #1440, but I feel like until a proper way to walk through archives is implemented, it is nice to have a way to preview .tar.* archive contents.

The bulk of my changes is that if the file name resembles a tar file, I use either `tar` (if supported, on unix systems, and if the actual program exists) or `7z/7zz` to preview the archive contents (I do so by extracting the .tar, and directly piping it back through 7-zip to list the contents). The reason I prefer to use "native" `tar` if possible is that `tar` supports extra formats, such as `.tar.zst` files, which 7-zip still doesn't support.

I also refactored the `archive.lua` file, to re-use code for the new tar support, as well as give more options for external plugins which do archive preview to use some of the functions in this plugin, such as the final display as a list of files in the preview window.

PS: I am quite new to Lua, and welcome any comments on how to make my code more idiomatic.

### DIFF
--- a/yazi-plugin/preset/plugins/archive.lua
+++ b/yazi-plugin/preset/plugins/archive.lua
@@ -214,7 +214,7 @@ function M.is_encrypted(s) return s:find(" Wrong password", 1, true) end
 function M.is_tar(path) return M.list_meta { "-p", tostring(path) } == "tar" end
 
 function M.should_decompress_tar(file)
-	return file.packed_size <= 1024 * 1024 * 1024 and file.path:find(".+%.tar$") ~= nil
+	return file.packed_size <= 1024 * 1024 * 1024 and file.path:lower():find(".+%.tar$") ~= nil
 end
 
 -- Parse the output of a "7z l -slt" command.


sxyazi:
Thank you for your patch!

This PR is quite large, 212 lines added and 15 lines deleted, while the entire archive plugin is only 173 lines, which makes it pretty hard to review.

Could you minimize the changes to only the necessary parts:

1. Revert the original style at line 30.
2. Revert the original style at line 43.
3. Remove comment blocks like `=========` to keep coding style consistent - there are no such comments elsewhere in the codebase.
4. Revert the relocation of the `M:seek` method at line 49.
5. Remove the `tar`-related code so this PR can focus on enhancing the existing 7zip functionality. Adding `tar` would require it to be listed as an optional dependency for Yazi, which increases the dependency surface and I'm not confident about its cross-platform support and I don't want to take on maintaining another tool in addition to 7zip (I want to maintain Yazi long-term and that would increase maintenance cost). That said, `tar` can be made as a good community plugin that interested users can use freely.

itsvyle:
Hi,

I have implemented the changes you requested.

For the `tar` functionality - the plugin would be less than 100 lines (thanks to using functionality which is available in the API of archive); do I publish the plugin myself? Or should I make a PR to <https://github.com/yazi-rs/plugins>?

yazi-plugin/preset/plugins/archive.lua
```diff
@@ -2,8 +2,26 @@ local M = {}
 
 function M:peek(job)
```
sxyazi, line 3: Can you revert the refactor to `M:peek`, for example the new `display_response` method? Since it's only called once, I don't see the need for it to be a separate method, and if it's for code cleanliness, it's better done in a separate refactor PR.

yazi-plugin/preset/plugins/archive.lua
```diff
@@ -66,23 +84,46 @@ function M.spawn_7z(args)
...
-	if not child then
-		return {}, 0, Err("Failed to start either `7zz` or `7z`. Do you have 7-zip installed?")
-	end
-
+function M.parse_7z_list(childL, skip, limit)
 	local i, files, err = 0, { { path = "", size = 0, attr = "" } }, nil
 	local key, value, stderr = "", "", {}
 	repeat
-		local next, event = child:read_line()
+		local next, event = childL:read_line()
```
sxyazi: Please keep the original name here

yazi-plugin/preset/plugins/archive.lua
```diff
@@ -170,4 +252,27 @@ function M.is_encrypted(s) return s:find(" Wrong password", 1, true) end
 
 function M.is_tar(path) return M.list_meta { "-p", tostring(path) } == "tar" end
 
+function M.is_nested_tar_path(path)
```
sxyazi: 7-Zip's output for `*.tar.*` contains only a single file named `*.tar` and it's better to check the file type from that output since it's more accurate than relying on the extension.

itsvyle: This would imply decompressing the file twice - once to check if it's tar, and a second to list the contents - which can be heavy if the file is large, or if the system isn't very powerful

sxyazi:
Thank you for the change!

> do I publish the plugin myself? Or should I make a PR to https://github.com/yazi-rs/plugins?

Either way is fine with me :)

yazi-plugin/preset/plugins/archive.lua
sxyazi: `7zz l` just lists the files contained in the archive and does not actually decompress them. 

In my test, it's pretty fast - for a 4.6G `.tar.gz` file, the elapsed time is 0.009:

```
time 7zz l ~/Downloads/test.tar.gz

7-Zip (z) 25.01 (arm64) : Copyright (c) 1999-2025 Igor Pavlov : 2025-08-03
 64-bit arm_v:8.5-A locale=en_US.UTF-8 Threads:8 OPEN_MAX:2560, ASM

Scanning the drive for archives:
1 file, 4956930267 bytes (4728 MiB)

Listing archive: /Users/ika/Downloads/test.tar.gz

--
Path = /Users/ika/Downloads/test.tar.gz
Type = gzip
Headers Size = 10

   Date      Time    Attr         Size   Compressed  Name
------------------- ----- ------------ ------------  ------------------------
2026-01-07 15:41:03 .....    686703616   4956930267  test.tar
------------------- ----- ------------ ------------  ------------------------
2026-01-07 15:41:03          686703616   4956930267  1 files
7zz l ~/Downloads/test.tar.gz  0.00s user 0.00s system 71% cpu 0.009 total
```

yazi-plugin/preset/plugins/archive.lua
itsvyle: I will open a PR - my idea is to have the `display_response` be a public API, for other archive plugins (such as the 'tar' plugin)

yazi-plugin/preset/plugins/archive.lua
```diff
@@ -170,4 +241,18 @@ function M.is_encrypted(s) return s:find(" Wrong password", 1, true) end
...
+	-- Warning: doing -slt will *not* print the .tar file, it will only print the .tar.* file
```
sxyazi: What's your 7-Zip version? It seems fine with me to use `-slt`:

```sh
$ 7zz l -ba -slt -sccUTF-8 /Users/ika/Downloads/test.tar.gz
Path = test.tar
Size = 686703616
Packed Size = 4956930267
Modified = 2026-01-07 15:41:03
Host OS = Unix
CRC = 3B03A64C
```

Here `Path = test.tar` is the filename contained in the `test.tar.gz`

sxyazi: Oh I see, certain tarballs may return a different output

itsvyle: I had taken a screenshot which I forgot to send; listing the .tar with -slt seems to only work on `.tar.gz` files
image

PR #3525, opened 2026-01-07
feat: tree view for the preset archive previewer

itsvyle:

Follow-up on #3518

## Rationale of this PR

Two changes:
- Moved the display of the list to the preview into a separate function. This is so that other previewers may call into the public API of the `archive` plugin to display their own file lists in a beautiful list with icons and colors
- Added a "tree" style display for folders in archives:

| Before | After |
| ------------- | ------------- |
| image | image  |
| image | image |
| image | image |

### DIFF
--- a/yazi-plugin/preset/plugins/archive.lua
+++ b/yazi-plugin/preset/plugins/archive.lua
@@ -110,7 +110,7 @@ end
 function M.list_archive(args, skip, limit)
 	local child = M.spawn_7z { "l", "-ba", "-slt", "-sccUTF-8", table.unpack(args) }
 	if not child then
-		return {}, 0, Err("Failed to start either `7zz` or `7z`. Do you have 7-zip installed?")
+		return {}, Err("Failed to start either `7zz` or `7z`. Do you have 7-zip installed?")
 	end
 
 	local files, err = M.parse_7z_slt(child, skip, limit)
@@ -131,7 +131,7 @@ function M.list_compressed_tar(args, skip, limit)
 		{ "l", "-ba", "-slt", "-ttar", "-sccUTF-8", "-si" }
 	)
 	if not dst then
-		return {}, 0, Err("Failed to start either `7zz` or `7z`. Do you have 7-zip installed?")
+		return {}, Err("Failed to start either `7zz` or `7z`. Do you have 7-zip installed?")
 	end
 
 	local files, err = M.parse_7z_slt(dst, skip, limit)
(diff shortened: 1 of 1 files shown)


sxyazi:
This looks fantastic! There's a problem though - when I seek down the preview (pressing the `J` key) there are some layout shifts that flicker between the list and the tree mode.

itsvyle:
> there are some layout shifts that flicker between the list and the tree mode

These should have been fixed by my latest change.

However, I noticed a new issue: some archive files don't actually contain the directories as "files" in their contents. As a result, my current way of detecting directories isn't working; example:

| Original archive | Decompressed then recompressed with 7z |
| -------- | -------- |
| image | image|
| image | image |

I'm wondering if auto-detecting a directory structure inside of the list is a good idea. Realistically, it's only at most processing on a terminal height's amount of files (50-ish?), but it could imply making some assumptions about the environment which can be pretty unknown.

sxyazi:
Yeah 7-Zip sometimes doesn't report directory status or even any attributes for a file. Do you think detect directories solely from the path? For example, for the path `a/b/c`, if `c` appears then that proves `a/b` exists and is a directory.

itsvyle:
I have added auto detection for directories.

Sadly, this only works if the files are listed in the correct lexical order, which is the order by default on a created zip, but isn't actually the order that 7z lists the files in - 7z just lists the files in an order decided by the compressor of the file.

This can lead to bad situations like this one (seems to only happen on zip files, although I'm not 100% sure it can't happen on tarballs)

image

And with the current architecture, I don't think there is much that can be done - because if we sort the files we have listed (which are limitted to the height of the terminal), scrolling through the archive will be very weird. We would have to get all the files in the archive and then sort them, but then that can add a lot of overhead.

I haven't really looked into a way to recreate an example zip file which has problematic order.

I however personally believe that the current behavior, where the archive is required to have been created with files in lexical order (which has been the default on 7z for years), is good enough, as it should regard most of the archive files out there.

sxyazi:
Hi, I did some refactoring to simplify the code, could you have a look to see if there are any errors?

During testing I found an issue: if an archive has a deep directory structure, scroll down to the end can cause some files to be lost, here's the steps to repruduce it:

1. Download [test.zip](https://github.com/user-attachments/files/24528791/test.zip)
2. Keep pressing `J` until you reach the bottom
3. The file `bbbbbbbbbb` (ten `b`) is missing, and only `bbbbbbbb` (eight `b`) is shown

UcUsBF0Q

Do you have any idea how to fix this?

itsvyle:
We created a bunch of rows for directories - these aren't listed inside the `7z l` output, but we scroll through that list, rather than the one with the directories. We would need to find a way to adjust "job.skip" for this shift; "job.skip" for the next page should be equal to "bound", not the number of lines which were outputted.

I'm not sure how job.skip is defined though, or if we can modify it...

sxyazi:
`skip` is changed inside the `M:seek()` method - basically it just increases by unit*screen_height/10 of units each time:

https://github.com/sxyazi/yazi/blob/d156508423981ea5f98d5456d71ca20e2bc2293d/yazi-plugin/preset/plugins/code.lua#L18-L24

But we can't get 7-Zip's output inside the `seek` method because [it's sync](https://yazi-rs.github.io/docs/plugins/overview#previewer) and not suitable for running time-consuming tasks.

sxyazi:
OK I think I've found a way to fix it in https://github.com/sxyazi/yazi/pull/3525/changes/dce7115850b6421a1408aa82ead1bf19b4f8be47, there are no blockers now

Issue #2500, opened 2025-03-20
Updates on NFS Server do not show up on Clients


CBeerta:
### What system are you running Yazi on?

Linux Wayland

### What terminal are you running Yazi in?

foot 1.12.0

### `yazi --debug` output

```Shell

Yazi
    Version: 25.3.7 (570c6d3 2025-03-19)
    Debug  : false
    Triple : x86_64-unknown-linux-gnu (linux-x86_64)
    Rustc  : 1.85.1 (4eb16125 2025-03-15)

Ya
    Version: 25.3.7 (570c6d3 2025-03-19)

Emulator
    TERM                : Some("tmux-256color")
    TERM_PROGRAM        : Some("tmux")
    TERM_PROGRAM_VERSION: Some("3.5a")
    Brand.from_env      : Some(Foot)
    Emulator.detect     : Emulator { kind: Left(Foot), light: false, cell_size: Some((7, 17)) }

Adapter
    Adapter.matches    : Sixel
    Dimension.available: WindowSize { rows: 80, columns: 365, width: 2555, height: 1360 }

Desktop
    XDG_SESSION_TYPE           : Some("wayland")
    WAYLAND_DISPLAY            : Some("wayland-1")
    DISPLAY                    : Some(":0")
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: Some("882f7ad7d2bbfc7440d0ccaef93b1cdd78e8e3ff_1741547704_1836036775")
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Variables
    SHELL           : Some("/usr/bin/bash")
    EDITOR          : Some("vim")
    VISUAL          : None
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
    FZF_DEFAULT_OPTS: Some("--bind ctrl-a:select-all,ctrl-d:deselect-all,ctrl-t:toggle-all --walker-skip .git,node_modules,target --walker=dir,follow --preview \'tree -C {}\'")

Text Opener
    default     : Some(Opener { run: "${EDITOR:=vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block-create: Some(Opener { run: "${EDITOR:=vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
[...]

PR #3526, opened 2026-01-07
refactor: make partition heuristic predicate a positive match

gwicke:
Make the partition "heuristic" predicate a positive match, in line with the "systemic" predicate, and document its expected semantics for future readers. Update the only call site to match.

Related to the NFS - specific investigation at #2500, but will not resolve the issue discussed there.

### DIFF
--- a/yazi-fs/src/mounts/partition.rs
+++ b/yazi-fs/src/mounts/partition.rs
@@ -14,9 +14,11 @@ pub struct Partition {
 }
 
 impl Partition {
+	// Match mount types that do not reliably emit change notifications, or update directory
+	// metadata on changes, and should be refreshed frequently / heuristically.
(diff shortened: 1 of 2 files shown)


sxyazi (approved):
Thank you!

gwicke:
@sxyazi, I noticed that this patch actually changed the non-linux/mac case at https://github.com/sxyazi/yazi/blob/main/yazi-fs/src/mounts/partitions.rs#L41 to default to "heuristic". Are other platforms generally reliable wrt update notifications? Happy to send a follow-up to flip the default to "false" if that is the case.

sxyazi:
Other platforms are unknown, but it's better to keep the original behavior (`false`) till someone someday reports that it actually needs to be changed to `true` :D

PR #3528, opened 2026-01-07
fix: trust directory stats outside of Linux and macOS

gwicke:
Restore the previous default of trusting directory stat updates to reflect directory changes on platforms other than linux & mac, without checking for mount types.

Also clarify wording in heuristic comment slightly.

#3526


sxyazi (approved):
Thank you

PR #3529, opened 2026-01-07
EXPERIMENT: Nfs update experiment for #2500

gwicke:
See if nfs4 update issues do indeed need to be addressed by heuristic polling.

NOTE: I am not positive this is needed, only putting this up for actual nfs users to test.
For testers: As a manual test, you can also compare `stat <directory>` output before / after directory changes (file creation / removal). If the stat times do not change, then heuristic polling will be needed. If they do change, then the issue will likely be elsewhere.


gwicke:
Per discusion in #2500, this does not seem to be needed since NFS updates stat data after cache expires (default up to 60 seconds).

PR #3530, opened 2026-01-07
feat: Add multi-file selection support to spot plugins with duration/size summation

salah0eldin:

Resolves #

## Rationale of this PR

Adds support for viewing aggregated information when multiple files are selected and Tab is pressed. Users can now see total duration for selected videos or total size for selected folders.

### Changes

**Rust core** (passes selected files to Lua plugins):
- `yazi-actor/src/mgr/spot.rs`: Collects `selected` URLs from tab state
- `yazi-core/src/spot/spot.rs`: Updates `go()` to accept `selected: Vec<UrlBuf>`
- `yazi-plugin/src/isolate/spot.rs`: Adds `selected` table to job passed to Lua

**video.lua**:
- Shows "# Videos (N)" with total duration when multiple videos selected
- Changed duration format to `H:MM:SS` (or `M:SS` if < 1 hour)
- Uses parallel ffprobe execution for performance

**folder.lua**:
- Shows "# Folders (N)" with total size when multiple folders selected
- Only counts directories, excludes files

The hovered file's detailed info still displays below the aggregate totals.

### DIFF
--- a/yazi-plugin/preset/plugins/folder.lua
+++ b/yazi-plugin/preset/plugins/folder.lua
@@ -51,8 +51,33 @@ end
 
 function M:spot(job)
 	self.size, self.last = 0, 0
+	self.selected_size = 0
+	self.folder_count = 0
+
+	-- Calculate selected folders sizes first if multiple are selected
+	if job.selected and #job.selected > 1 then
+		for _, file in ipairs(job.selected) do
+			local cha = fs.cha(file)
+			if cha and cha.is_dir then
+				local size = 0
+				local it = fs.calc_size(file)
+				while true do
+					local next = it:recv()
+					if next then
+						size = size + next
+					else
+						break
+					end
+				end
+				self.selected_size = self.selected_size + size
+				self.folder_count = self.folder_count + 1
+			end
+		end
+	end
+
 	self:spot_multi(job, false)
 
+	-- Calculate current folder size
 	local url = job.file.url
 	local it = fs.calc_size(url)
 	while true do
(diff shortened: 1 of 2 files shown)


sxyazi:
If you select multiple different types of files, say a video, an image, and a directory, which spotter should be used?

salah0eldin:
> If you select multiple different types of files, say a video, an image, and a directory, which spotter should be used?

Great point but handled already.
The answer is: the spotter of the one hovering on currently, and as you know we can switch hovering while spotting and this is handled automatically as next example shows:
First, we need multiple selection of same kind to see the difference from the ordinary so let's say
if we select two videos, two images, and two directories, 
and currently hovering on a video, it will show the spotter of videos showing the sum of their duration at top as well as the normal spotter info of the current hovered video underneath, if we switch using arrows to a folder it will show the spotter of folders showing the sum of their sizes at top as well as the normal spotter info, if we switch to any other type let's say the images here, it will only show the spotter info of the image as there is no custom selection for it.
  
 The idea could be far more expanded to show common things summed between unit type of elements, for example adding the sum of sizes of the hovered type in selections, or the sum of duration of audios, and for more to go.

sxyazi:
OK so you want to get the user's selection state in the spotter used for the hovered file right? That's already supported: you can access any context data you need via [`ya.sync()`](https://yazi-rs.github.io/docs/plugins/overview/#async-context). Also, there are no plans to build summary info into the preset spotters since that's already possible via custom spotters. In any case, thank you for the effort!

sxyazi:
This is now partly supported by https://github.com/sxyazi/yazi/pull/3733 - the new `multi` preset spotter shows the total size of the selected files as every file has that attribute, filetype-specific stuff like video duration still needs a custom spotter though.

Issue #1095, opened 2024-05-27
When grepping (`search rg`), preview could highlight the first match


mikavilpas:
### Please describe the problem you're trying to solve

When I search using S (`search rg`), and preview a file that matched my search, the file preview could
- scroll the preview to a section that shows the location of the first match, so that I can see the context of the match
- highlight the first match, so that I can quickly see where it matched

### Would you be willing to contribute this feature?

- [X] Yes, I'll give it a shot

### Describe the solution you'd like

This is the neovim + telescope implementation

<img width="1719" alt="image" src="https://github.com/sxyazi/yazi/assets/300791/8bcdf1ba-801a-4085-beea-c43c346db78e">

The matched files are displayed on the left, and the preview on the right.

Can it be done the same way? I think it's enough to just display each file once, even if it had multiple matches.

### Additional context

Not sure what the performance implications might be.

DreamMaoMao:
This task is want to make like [fg yazi] (https://github.com/DreamMaoMao/fg.yazi) to  a built-in function?

mikavilpas:
Hmm, conceptually they are very similar, but here are some differences:

- fg.yazi seems to use external programs - I actually meant simple highlighting in yazi's current search
- fg.yazi seems capable of showing each match as a separate result - yazi shows each matched file only once

soderluk:
+1 for this. It's a bit annoying to have to open the file, then search again, when I can't see where the matches are in the preview.

sxyazi:
Closing in favor of https://github.com/sxyazi/yazi/issues/3346

PR #3549, opened 2026-01-11
feat: make rg search more convenient

### LINKED ISSUE #3521: Url enhancement: support line and column addressing for more precise rg searches
### `yazi --debug` output

```Shell
Yazi
    Version: 26.1.4 (3c39a326 2026-01-06)
    Debug  : true
    Triple : x86_64-pc-windows-msvc (windows-x86_64)
    Rustc  : 1.92.0 (ded5c06c 2025-12-08)

Ya
    Version: 26.1.4 (3c39a326 2026-01-06)

Emulator
    TERM                : None
    TERM_PROGRAM        : Some("vscode")
    TERM_PROGRAM_VERSION: Some("1.107.1")
    Brand.from_env      : Some(VSCode)
    Emulator.detect     : Emulator { kind: Left(VSCode), version: "", light: false, csi_16t: (0, 0), force_16t: false }

Adapter
    Adapter.matches    : Iip
    Dimension.available: Dimension { rows: 31, columns: 150, width: 0, height: 0 }

Desktop

SSH

WSL
    WSL: false

Variables
    SHELL              : None
    EDITOR             : None
    VISUAL             : None
    YAZI_FILE_ONE      : None
    YAZI_CONFIG_HOME   : None
    YAZI_ZOXIDE_OPTS   : None
    FZF_DEFAULT_OPTS   : None
    FZF_DEFAULT_COMMAND: None

Text Opener
    default     : Some(OpenerRule { run: "code -g \"%s:%l\"", block: true, orphan: false, desc: "Edit with default editor", for: None, spread: true })
    block-create: Some(OpenerRule { run: "code -g \"%s:%l\"", block: true, orphan: false, desc: "Edit with default editor", for: None, spread: true })
    block-rename: Some(OpenerRule { run: "code -g \"%s:%l\"", block: true, orphan: false, desc: "Edit with default editor", for: None, spread: true })

Multiplexers
    TMUX               : false
    tmux version       : program not found
    ZELLIJ_SESSION_NAME: None
    Zellij version     : program not found

Dependencies
    file          : 5.45
    ueberzugpp    : program not found
    ffmpeg/ffprobe: 8.0.1 / 8.0.1
    pdftoppm      : 25.07.0
    magick        : 7.1.2-11
    fzf           : 0.67.0
    fd/fdfind     : 10.3.0 / program not found
    rg            : 15.1.0
    chafa         : program not found
    zoxide        : 0.9.8
    7zz/7z        : program not found / program not found
    resvg         : program not found
    jq            : 1.8.1

Clipboard
    wl-copy/paste: program not found / program not found
    xclip        : program not found
    xsel         : program not found

Routine
```

## Intention
I think Yazi's content searching experience is not quite ideal compared to VS Code currently.
I looked at the attempts of #3346 and #1095. and I summarized our needs as follows:
- Show the matched line in the preview so you can tell if it’s the file you want
- Place the cursor on the matched line (or the words) when opening the file in the default editor so you can start editing right away
And I find it impossible to solve this purely with a plugin, so we should tackle it step by step.

## Solution
I’d break the solution down like this:
1. Yazi will transmit line information to the previewer and the opener while in search mode (transparent to user, purely a system design detail).
2. The previewers will be responsible for highlighting logic.
3. The openers will accept those parameters and behave accordingly.

With that in mind, it makes sense to extend Url to support steps 1 and 3. I suggest adding line and column fields to Url.
```rs
pub enum UrlBuf {
	Regular(LocBuf),
	Search { loc: LocBuf, domain: Symbol<str>, line: Option<usize>, column: Option<usize> },
	Archive { loc: LocBuf, domain: Symbol<str> },
	Sftp { loc: LocBuf<typed_path::UnixPathBuf>, domain: Symbol<str> },
}

// same as UrlCow and Url
```
My implementation focuses on extending the Url module so previewers can highlight precisely and openers can jump to the exact line/column.

## Why this approach
1. Including line and column in Url just feels natural.
2. Url is a flexible part of Yazi; using it to carry search details to the previewer and opener keeps added system complexity minimal.
3. Minimal impact on other modules: Copy/Move/Delete/Tab/Folder can simply ignore the extra attributes.

I’m still unsure about the best way to handle step 2. The previewer could receive an appropriately offset position in the peek function, but we might want a new method to tell the previewer which line to highlight. Or we might want a single built-in highlighting mechanism.

>  We need to consider the worst-case scenario: if the offset is large, for example, if the first match is the last line of a 10,000-line file, highlighting will take a long time. How will we handle this?

This is a step 2 problem. I don't know if it is critical XD

I use `--no-heading` in rg.rs, so it shows matches inline instead of grouped by file.

### Additional context

https://github.com/user-attachments/assets/c232673d-24cb-4302-b471-0bbc753430c2

WIP
Still buggy. Path display is kinda broken, and a few larger bugs that cause other issues

### Checklist

Cachey4467:
Amazing tool! Try to improve the rg search experience.

Resolves #3521  

## Rationale of this PR
See the issue for more details.

- URL supports #line:column in domain
  - Added Url::line() and Url::column() methods to parse line/column from search URL fragments (e.g., keyword#123:45)
  - Updated scheme::encode::domain() to include # in the encode set
- Enhanced rg plugin
  - Changed rg args to --line-number --no-heading --column for detailed match positions
  - Added parse_rg_line_column() parser function
  - Search results now include line and column in the URL
- Preview improvements
  - When opening a file with line info, preview automatically scrolls to show the matched line at the top
- Search reveal support
  - Added Reveal::reveal_search() to handle revealing files within search results
- Splatter format support
  - Added %l/%L for line number
  - Added %c/%C for column number
- Lua bindings
  - Added file:line() method to get line number from URL

I changed the reveal logic to account for Search URL.

### DIFF
--- a/yazi-actor/src/mgr/peek.rs
+++ b/yazi-actor/src/mgr/peek.rs
@@ -2,7 +2,7 @@ use anyhow::Result;
 use yazi_macro::succ;
 use yazi_parser::mgr::PeekOpt;
 use yazi_proxy::HIDER;
-use yazi_shared::data::Data;
+use yazi_shared::{data::Data, url::AsUrl};
 
 use crate::{Actor, Ctx};
 
@@ -25,7 +25,8 @@ impl Actor for Peek {
 		let folder = cx.tab().hovered_folder().map(|f| (f.offset, f.cha));
 
 		if !cx.tab().preview.same_url(&hovered.url) {
-			cx.tab_mut().preview.skip = folder.map(|f| f.0).unwrap_or_default();
+			// set matched line on the top of preview window
+			cx.tab_mut().preview.skip = folder.map(|f| f.0).unwrap_or_else(|| hovered.url.as_url().line().map_or(0, |l| l.saturating_sub(1)));
 		}
 		if !cx.tab().preview.same_file(&hovered, &mime) {
 			cx.tab_mut().preview.reset();
--- a/yazi-actor/src/mgr/reveal.rs
+++ b/yazi-actor/src/mgr/reveal.rs
@@ -2,7 +2,7 @@ use anyhow::Result;
 use yazi_fs::{File, FilesOp};
 use yazi_macro::{act, render, succ};
 use yazi_parser::mgr::RevealOpt;
-use yazi_shared::{data::Data, url::UrlLike};
+use yazi_shared::{data::Data, url::{AsUrl, UrlLike}};
 
 use crate::{Actor, Ctx};
 
@@ -14,6 +14,15 @@ impl Actor for Reveal {
 	const NAME: &str = "reveal";
 
 	fn act(cx: &mut Ctx, opt: Self::Options) -> Result<Data> {
+		match cx.cwd().is_search() {
+			true => Self::reveal_search(cx, opt),
+			false => Self::reveal_regular(cx, opt),
+		}
+	}
+}
+
+impl Reveal {
+	fn reveal_regular(cx: &mut Ctx, opt: RevealOpt) -> Result<Data> {
 		let Some((parent, child)) = opt.target.pair() else { succ!() };
 
 		// Cd to the parent directory
@@ -33,6 +42,30 @@ impl Actor for Reveal {
 		// Now, we can safely hover over the target
 		act!(mgr:hover, cx, Some(child.into()))?;
 
+		act!(mgr:peek, cx)?;
+		act!(mgr:watch, cx)?;
+		succ!();
+	}
+
+	fn reveal_search(cx: &mut Ctx, opt: RevealOpt) -> Result<Data> {
+		let cwd = cx.cwd().clone();
+		let tab = cx.tab_mut();
+		let pos = tab.current.files.iter().position(|f| f.url == opt.target);
+
+		if let Some(pos) = pos {
+			let delta = pos as isize - tab.current.cursor as isize;
+			tab.current.arrow(delta);
+		} else if !opt.no_dummy {
+			let op = FilesOp::Creating(cwd, vec![File::from_dummy(&opt.target, None)]);
+			tab.current.update_pub(tab.id, op);
+
+			if let Some(pos) = tab.current.files.iter().position(|f| f.url == opt.target) {
+				let delta = pos as isize - tab.current.cursor as isize;
+				tab.current.arrow(delta);
+			}
+		}
+
+		act!(mgr:hover, cx, None)?;
 		act!(mgr:peek, cx)?;
 		act!(mgr:watch, cx)?;
 		succ!();
--- a/yazi-binding/src/macros.rs
+++ b/yazi-binding/src/macros.rs
@@ -238,5 +238,10 @@ macro_rules! impl_file_methods {
 			// TODO: use a cache
 			Ok(yazi_config::THEME.icon.matches(me).map(Icon::from))
 		});
+
+		$methods.add_method("line", |_, me, ()| {
+			use yazi_shared::url::AsUrl;
+			Ok(me.url.as_url().line().map(|v| v as i64))
+		});
 	};
 }
--- a/yazi-fs/src/splatter.rs
+++ b/yazi-fs/src/splatter.rs
@@ -2,7 +2,7 @@
 use std::os::unix::ffi::{OsStrExt, OsStringExt};
 #[cfg(windows)]
 use std::os::windows::ffi::{OsStrExt, OsStringExt};
-use std::{cell::Cell, ffi::{OsStr, OsString}, iter::{self, Peekable}, mem};
+use std::{cell::Cell, ffi::{OsStr, OsString}, fs::OpenOptions, io::Write, iter::{self, Peekable}, mem};
 
 use yazi_shared::url::{AsUrl, Url, UrlCow};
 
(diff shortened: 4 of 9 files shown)


yazi-actor/src/mgr/peek.rs
```diff
@@ -25,7 +25,8 @@ impl Actor for Peek {
 		let folder = cx.tab().hovered_folder().map(|f| (f.offset, f.cha));
 
 		if !cx.tab().preview.same_url(&hovered.url) {
-			cx.tab_mut().preview.skip = folder.map(|f| f.0).unwrap_or_default();
+			// set matched line on the top of preview window
+			cx.tab_mut().preview.skip = folder.map(|f| f.0).unwrap_or_else(|| hovered.url.as_url().line().map_or(0, |l| l.saturating_sub(1)));
```
sxyazi, line 29: This should be handled in the [`seek()` method](https://yazi-rs.github.io/docs/plugins/overview#previewer) of `code.lua`: when the URL contains line numbers, the `unit` for `seek()` should be `1`, meaning it should jump between matching lines, then, `peek()` should get the N-th line number by `skip`.

yazi-actor/src/mgr/reveal.rs
```diff
@@ -37,4 +46,28 @@ impl Actor for Reveal {
 		act!(mgr:watch, cx)?;
 		succ!();
 	}
+
+	fn reveal_search(cx: &mut Ctx, opt: RevealOpt) -> Result<Data> {
```
sxyazi, line 50: Why is a new method needed?

yazi-binding/src/macros.rs
```diff
@@ -238,5 +238,10 @@ macro_rules! impl_file_methods {
 			// TODO: use a cache
 			Ok(yazi_config::THEME.icon.matches(me).map(Icon::from))
 		});
+
+		$methods.add_method("line", |_, me, ()| {
```
sxyazi, line 242: This should be handled in the `peek()` method of `code.lua`

yazi-fs/src/splatter.rs
```diff
@@ -81,6 +81,8 @@ where
 			Some('d') | Some('D') => self.visit_dirname(it, buf),
 			Some('t') | Some('T') => self.visit_tab(it, buf),
 			Some('y') | Some('Y') => self.visit_yanked(it, buf),
+			Some('l') | Some('L') => self.visit_line_number(it, buf),
```
sxyazi, line 84: Any reason `l` and `c` are needed?

yazi-plugin/preset/components/entity.lua
```diff
@@ -43,6 +43,13 @@ end
 
 function Entity:highlights()
 	local name, p = self._file.name, ui.printable
+
+	local ok, line = pcall(function() return self._file:line() end)
```
sxyazi, line 47: What if the file has multiple lines matched?

yazi-plugin/src/utils/preview.rs
```diff
@@ -18,10 +18,7 @@ impl Utils {
 			let path = lock.url.as_url().unified_path();
 			let inner = match Highlighter::new(path).highlight(lock.skip, area.size()).await {
 				Ok(text) => text,
-				Err(e @ PeekError::Exceed(max)) => return (e.to_string(), max).into_lua_multi(&lua),
```
sxyazi: Why is this change made?

yazi-shared/src/scheme/encode.rs
```diff
@@ -14,7 +14,7 @@ impl<'a> From<crate::url::Encode<'a>> for Encode<'a> {
 impl<'a> Encode<'a> {
 	#[inline]
 	pub fn domain<'s>(s: &'s str) -> PercentEncode<'s> {
-		const SET: &AsciiSet = &CONTROLS.add(b'/').add(b':');
+		const SET: &AsciiSet = &CONTROLS.add(b'/').add(b':').add(b'#');
```
sxyazi, line 17: This is not right - `SET` here is used to encode the entire `domain`, not just the keyword part in the `domain`, so `.add(b'#')` will not work if you intend to use it as the separator between the keyword and the line/col numbers

yazi-shared/src/url/url.rs
```diff
@@ -86,6 +86,36 @@ impl<'a> Url<'a> {
...
+			return None
+		};
+
+		let frag = &domain[hash_pos + 1..];
+		if let Some(colon_pos) = frag.find(':') {
+			frag[colon_pos + 1..].parse::<usize>().ok()
+		} else {
+			None
+		}
+	}
```
sxyazi, line 117: This should be handled in the respective previewer so that different previewers can use different formats for different search metadata

sxyazi (changes requested):

yazi-plugin/preset/components/entity.lua
Cachey4467: This implementation is designed to achieve a #1095 like rg search experience, where each match in the same file is treated as an individual entry.

image

So 1 match = 1 "file"

sxyazi: Why do you want to show the same file as multiple entries? Is there any advantage to this compared to showing it as a single entry?

Cachey4467: Well... It feels natural to me (that's how nvim + telescope handles it as seen in #1095 ). I haven't dug into the project too much yet, but would the line and column information be accessible in the context for opening files in the Openers? And I understand if you think it's bit complex or disruptive to the module design.

sxyazi: That's possible, but it's better to discuss it in a separate issue on how to attach line numbers to the file path and keep this PR focused on preview matched lines

yazi-actor/src/mgr/reveal.rs
Cachey4467: If we use a search URL to create a File in rg.rs, `to_search()` ends up zeroing out the URN and URI, which results in the placeholder file issue in the reveal scenario. We’re using search URLs in the first place because regular URLs don't have a domain field.

``` rust
let Some((fp, l, c)) = parse_rg_line_column(&line) else { continue };

let url = cwd.try_join(fp)
	.ok()
	.and_then(|u| u.to_search(&format!("{}#{}:{}", subject, l, c)).ok()); // URI set to 0 causing the placeholder file issue

let Some(url) = url else { continue };

if let Ok(file) = File::new(url).await {
	tx.send(file).ok();
}
```

PR #3555, opened 2026-01-12
feat: renaming empty files into directories

### LINKED ISSUE #2853: Ability to rename empty files and convert them to a directory
### `yazi --debug` output

```Shell
Yazi
    Version: 25.4.8 (Arch Linux 2025-04-09)
    Debug  : false
    Rustc  : 1.86.0 (05f9846f 2025-03-31)

Ya
    Version: 25.4.8 (Arch Linux 2025-04-09)

Emulator
    TERM                : Some("alacritty")
    TERM_PROGRAM        : None
    TERM_PROGRAM_VERSION: None
    Brand.from_env      : None
    Emulator.detect     : Emulator { kind: Right(Unknown { kgp: false, sixel: false }), light: false, cell_size: None }

Adapter
    Adapter.matches    : Wayland
    Dimension.available: WindowSize { rows: 57, columns: 211, width: 1899, height: 1026 }

Desktop
    WAYLAND_DISPLAY            : Some("wayland-1")
    DISPLAY                    : Some(":1")
    HYPRLAND_INSTANCE_SIGNATURE: Some("185c96849ef59da3e101116662d942dea16b468a_1749395703_1562938571")

SSH

WSL
    WSL: false

Variables
    SHELL           : Some("/bin/zsh")
    EDITOR          : Some("nvim")
    VISUAL          : Some("neovide")
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
    FZF_DEFAULT_OPTS: None

Text Opener
    default     : Some(OpenerRule { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block-create: Some(OpenerRule { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block-rename: Some(OpenerRule { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })

Multiplexers
    TMUX               : false
    ZELLIJ_SESSION_NAME: None

Dependencies
    file          : 5.46
    ueberzugpp    : 2.9.6
    ffmpeg/ffprobe: 7.1.1 / 7.1.1
    pdftoppm      : 25.05.0
    magick        : 7.1.1-47
    fzf           : 0.62.0
    fd/fdfind     : 10.2.0 / No such file or directory (os error 2)
    rg            : 14.1.1
    chafa         : 1.14.5
    zoxide        : 0.9.7
    jq            : 1.7.1

Clipboard
    wl-copy/paste: 2.2.1 / 2.2.1

Routine
```

Often times I try to create a directory but forget to suffix it with '/' and I end up deleting the new created file and creating it again by adding the '/' would be neat If when renaming the file if I just add the '/' it would just convert it to a directory. I think it's intuitive feature and I bet a lot of people would appreciate this feature

I think it would be as easy as checking if the file you are trying to rename is a completely empty file (0 bytes in size) then you are allowed convert it to a directory by just adding the '/' at the end.

### Additional context

_No response_

### Checklist

jlledo:

Resolves #2853

---

I'm not familiar with the codebase so I've based this implementation on the "create" actor.

Tests performed:

- Rename empty file to dir with same name -> converted to dir
- Rename empty file to dir with different name -> converted to dir
- Rename non-empty file to dir -> aborted and warning shown
- Rename file to file with different name -> renamed
- Rename dir to dir with a different name -> renamed

### DIFF
--- a/yazi-actor/src/mgr/rename.rs
+++ b/yazi-actor/src/mgr/rename.rs
@@ -67,18 +67,18 @@ impl Actor for Rename {
 				return;
 			};
 
-			if opt.force || !maybe_exists(&new).await || provider::must_identical(&old, &new).await {
-				if conversion_to_dir_requested && target_is_empty {
-					Self::replace_file_with_dir(old, new).await.ok();
-				} else {
-					Self::r#do(tab, old, new).await.ok();
-				}
-			} else if ConfirmProxy::show(ConfirmCfg::overwrite(&new)).await {
-				if conversion_to_dir_requested && target_is_empty {
-					Self::replace_file_with_dir(old, new).await.ok();
-				} else {
-					Self::r#do(tab, old, new).await.ok();
-				}
+			if !opt.force
+				&& maybe_exists(&new).await
+				&& !provider::must_identical(&old, &new).await
+				&& !ConfirmProxy::show(ConfirmCfg::overwrite(&new)).await
+			{
+				return;
+			}
+
+			if conversion_to_dir_requested && target_is_empty {
+				Self::replace_file_with_dir(old, new).await.ok();
+			} else {
+				Self::r#do(tab, old, new).await.ok();
 			}
 		});
 		succ!();
--- a/yazi-actor/src/mgr/rename.rs
+++ b/yazi-actor/src/mgr/rename.rs
@@ -76,9 +76,9 @@ impl Actor for Rename {
 			}
 
 			if conversion_to_dir_requested && target_is_empty {
-				Self::replace_file_with_dir(old, new).await.ok();
+				_ = Self::replace_file_with_dir(old, new).await;
 			} else {
-				Self::r#do(tab, old, new).await.ok();
+				_ = Self::r#do(tab, old, new).await;
 			}
 		});
 		succ!();
--- a/yazi-actor/src/mgr/rename.rs
+++ b/yazi-actor/src/mgr/rename.rs
@@ -4,8 +4,12 @@ use yazi_dds::Pubsub;
 use yazi_fs::{File, FilesOp};
 use yazi_macro::{act, err, ok_or_not_found, succ};
 use yazi_parser::mgr::RenameOpt;
-use yazi_proxy::{ConfirmProxy, InputProxy, MgrProxy};
-use yazi_shared::{Id, data::Data, url::{UrlBuf, UrlLike}};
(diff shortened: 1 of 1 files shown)


jlledo:
Something I'm not too sure about is the DDS event pub. I've removed the rename event for directory conversions because I think it doesn't really fit, but we should probably still publish some event(s?) - even if not the existing one.

Thoughts?

sxyazi:
Thank you for your patch!

I'm not quite sure if I want to support this feature:

- It checks the file length equal to 0 to decide if the file is empty, but I'm not certain if all filesystems have the same behavior, such as FUSE or remote filesystems.
- The definition of an empty file is unclear. If a file contains nothing but a trailing newline, should it be considered empty? This is [common on *nix](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206), and text editors may insert a newline automatically.
- By appending a `/`, you can rename an empty file to a directory, but I'm not sure how the reverse should be achieved, namely, how to turn an empty directory into a file? This feels incomplete.
- The OS doesn't provide an API to rename a file as a directory, so this has to be implemented as `stat` + `remove` + `mkdir`, which is not an atomic operation and introduces a [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) race condition. That is, if another process modifies the filesystem between the `stat` and `remove`, it may lead to data loss. On remote filesystems, this problem could be exacerbated by network instability, which is unsafe.

jlledo:
No worries, I figured the PR might not get accepted since you didn't comment on the issue, as well as the non-atomicity problem. And that's completely reasonable.

Just for the sake of discussion:

- I would be surprised if any mainstream filesystem considered anything other than files with a content size of 0 as empty.
- For this use case I would define empty as a file with no content whatsoever, whitespace or not. The feature provides a convenient way to correct files that get created by mistake when a directory was intended, by forgetting the trailing slash - and those are truncated afaict.
- Fair. However, I imagine the reverse functionality wouldn't be that useful haha.
- Again, fair. Unlikely given the use case but I understand if this level of data loss is unacceptable.

sxyazi:
I decided not to build it in because of said security and completeness concerns, but maybe it's worth making it as a simple plugin. In any case, thanks for your effort!

jlledo:
That’s understandable. Thank you for your time.

PR #3582, opened 2026-01-17
fix: `sort extension` excludes directories since only files have extensions

### LINKED ISSUE #3580: directories should not be sorted by extension
### `yazi --debug` output

```Shell
not applicable
```

when sorting is done by extension directories are also sorted if they contain a dot (`.`)

directories have dots (`.`) for reasons other that indicating extension, for example `folder v2.1.6`

dont apply sorting to directories

patch i came up with:
```diff
diff --git a/yazi-fs/src/sorter.rs b/yazi-fs/src/sorter.rs
index 2d745dc7..cd4de41a 100644
--- a/yazi-fs/src/sorter.rs
+++ b/yazi-fs/src/sorter.rs
@@ -40,12 +40,14 @@ impl FilesSorter {
 				if ord == Ordering::Equal { by_alphabetical(a, b) } else { ord }
 			}),
 			SortBy::Extension => items.sort_unstable_by(|a, b| {
+				let aa = if a.is_dir() && self.dir_first { a.name() } else { a.url.ext() };
+				let bb = if b.is_dir() && self.dir_first { b.name() } else { b.url.ext() };
 				let ord = if self.sensitive {
-					self.cmp(a.url.ext(), b.url.ext(), self.promote(a, b))
+					self.cmp(aa, bb, self.promote(a, b))
 				} else {
 					self.cmp_insensitive(
-						a.url.ext().map_or(&[], |s| s.encoded_bytes()),
-						b.url.ext().map_or(&[], |s| s.encoded_bytes()),
+						aa.map_or(&[], |s| s.encoded_bytes()),
+						bb.map_or(&[], |s| s.encoded_bytes()),
 						self.promote(a, b),
 					)
 				};
```

current behaviour vs fixed behaviour:
Image

### Additional context

_No response_

### Checklist

AminurAlam:

Resolves #3580

## Rationale of this PR

when sorting is done by extension directories are also sorted if they contain a dot (`.`)

directories have dots (`.`) for reasons other that indicating extension, for example folder v2.1.6

Image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,8 +24,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/):
 
 - Smart-case in interactive `cd` broken due to a typo ([#3540])
 - Fix shell formatting for non-spread opener rules ([#3532])
-- Account for URL covariance in `Url:join()` ([#3514])
 - `sort extension` excludes directories since only files have extensions ([#3582])
+- Account for URL covariance in `Url:join()` ([#3514])
 
 ## [v26.1.4]
 
@@ -341,7 +341,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/):
 
 ### Fixed
 
-- `before_ext` should exclude directories since only files have extensions ([#2132])
+- `before_ext` excludes directories since only files have extensions ([#2132])
 - Element style of `ui.Text` was not applied to the entire area ([#2093])
 - Incorrect monorepo sub-plugin path resolution ([#2186])
 - Use `u32` for parsing Linux partition blocks ([#2234])
@@ -1606,3 +1606,4 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/):
 [#3541]: https://github.com/sxyazi/yazi/pull/3541
 [#3561]: https://github.com/sxyazi/yazi/pull/3561
 [#3566]: https://github.com/sxyazi/yazi/pull/3566
+[#3582]: https://github.com/sxyazi/yazi/pull/3582


AminurAlam:
does this also need a changelog update?

yazi-fs/src/sorter.rs
```diff
@@ -40,12 +40,14 @@ impl FilesSorter {
 				if ord == Ordering::Equal { by_alphabetical(a, b) } else { ord }
 			}),
 			SortBy::Extension => items.sort_unstable_by(|a, b| {
+				let aa = if a.is_dir() && self.dir_first { a.name() } else { a.url.ext() };
```
sxyazi:
```suggestion
				let aa = if a.is_dir() { a.name() } else { a.url.ext() };
```

Is `self.dir_first` needed?

sxyazi:
> does this also need a changelog update?

Yep

yazi-fs/src/sorter.rs
AminurAlam: no not really, either one works for me

left is without the extra condition and right is with the `&& self.dir_first`
Screenshot_20260117_213352_niri

sxyazi: Do you have `sort_dir_first = true` set in your `yazi.toml`? `&& self.dir_first` should only make a change with that option enabled

AminurAlam: yes i usually have that enabled but disabled it for the above screenshot

sxyazi:
Does this patch work for you:

```diff
diff --git a/yazi-fs/src/sorter.rs b/yazi-fs/src/sorter.rs
index 2d745dc7..5f189d2f 100644
--- a/yazi-fs/src/sorter.rs
+++ b/yazi-fs/src/sorter.rs
@@ -40,12 +40,14 @@ impl FilesSorter {
 				if ord == Ordering::Equal { by_alphabetical(a, b) } else { ord }
 			}),
 			SortBy::Extension => items.sort_unstable_by(|a, b| {
+				let aa = a.url.ext().filter(|_| a.is_file());
+				let bb = b.url.ext().filter(|_| b.is_file());
 				let ord = if self.sensitive {
-					self.cmp(a.url.ext(), b.url.ext(), self.promote(a, b))
+					self.cmp(aa, bb, self.promote(a, b))
 				} else {
 					self.cmp_insensitive(
-						a.url.ext().map_or(&[], |s| s.encoded_bytes()),
-						b.url.ext().map_or(&[], |s| s.encoded_bytes()),
+						aa.map_or(&[], |s| s.encoded_bytes()),
+						bb.map_or(&[], |s| s.encoded_bytes()),
 						self.promote(a, b),
 					)
 				};
```

it seems more natural to me:

| `sort_dir_first = false` | `sort_dir_first = true` |
|--------|--------|
| 3Gj3h645 | bQWZfzAT |

AminurAlam:
this is also looks good, i'll go with this one then

PR #3602, opened 2026-01-23
fix: remove duplicate categories in the `.desktop` file

### LINKED ISSUE #3601: Categories field: remove non-standard/redundant entries in yazi.desktop
### `yazi --debug` output

```Shell
N/A
```

The yazi.desktop file has redundant and non-standard entries in its Categories field, such as Core (which is not in the freedesktop spec) and System (which is redundant for yazi's use as a file manager/utility). These categories can interfere with proper desktop environment categorization and compliance.

Clean up the Categories entry in yazi.desktop to only include standard and necessary entries: Utility, FileManager, FileTools, and ConsoleOnly.

Example:
Categories=Utility;FileManager;FileTools;ConsoleOnly;

### Additional context

Reference: [specifications.freedesktop.org/menu-spec/latest/category-registry.html](https://specifications.freedesktop.org/menu/latest/category-registry.html)

This change would improve standards-compliance and integration with Linux desktop environments, so yazi appears correctly categorized.

### Checklist

Jaredy899:

Resolves #3601 

## Rationale of this PR

Avoids duplicate of yazi in Desktop categories.

### DIFF
--- a/assets/yazi.desktop
+++ b/assets/yazi.desktop
@@ -7,5 +7,5 @@ TryExec=yazi
 Exec=yazi %u
 Type=Application
 MimeType=inode/directory
-Categories=Utility;Core;System;FileTools;FileManager;ConsoleOnly
+Categories=System;FileManager;FileTools;ConsoleOnly
 Keywords=File;Manager;Explorer;Browser;Launcher


sxyazi:
Thanks for your patch!

I'm not sure if we should remove "System" - I looked at the `.desktop` files of other terminal file managers and they all seem to have it:

- `lf`: https://github.com/gokcehan/lf/blob/master/lf.desktop
- `ranger`: https://github.com/ranger/ranger/blob/master/doc/ranger.desktop
- `nnn`: https://github.com/jarun/nnn/blob/master/misc/desktop/nnn.desktop

Would keeping it cause any problems? Also, is there any reason it ends with a `;`?

Jaredy899:
Hi! In a DE like KDE plasma, keeping both utility and system means yazi shows up in both categories since plasma has both a utility and system menu. In your examples, none have utility so I could switch it to system instead and get rid of utility. Whichever. And no reason it ends in the semicolon. I can remove that too.

sxyazi:
Thanks for the explanation.

So you want the app not to show up in both "System" and "Utility" categories, but don't other categories have the same problem here e.g. the app shows up in both the "FileManager" and "ConsoleOnly". Is there any reason to treat "System" and "Utility" specially?

Jaredy899:
FileManager is an Additional category whose related main category is System, and FileTools is an Additional category whose related main category is “Utility or System”.

ConsoleOnly is also an Additional category that just marks “this only runs in a terminal”, not a user‑visible menu section.

So a terminal file manager described as ```FileManager;FileTools;ConsoleOnly``` is already saying “system-ish, file‑management utility that runs in a terminal” in spec terms.

FileManager and ConsoleOnly are not competing main categories; they describe different axes: “what does this do?” (file manager/file tool) vs “how does it run?” (only in a terminal).

A DE will typically use FileManager/FileTools/System/Utility to decide where it appears in the menu, and use ConsoleOnly only to decide things like “launch in a terminal” or whether to show it in purely GUI launchers at all.

sxyazi:
I see, that makes a lot of sense! 

I'm fine to accept the patch with the trailing `;` removed

Jaredy899:
Thanks! I also think it makes probably more sense to change to ```System``` so I'll do that too.

sxyazi:
Don't "FileManager" already mean "System"?

Jaredy899:
Main categories like System (for system tools) and Utility (for accessories) define top-level menu sections that desktops must support.

Additional categories like FileManager describe app functions with "Related Categories" suggestions—FileManager suggests Development, FileTools suggests Utility or System—but do not replace main categories.

Main categories like System (for "System Tools" such as log viewers or network monitors) and Utility (for small accessories) create top-level menu sections that every conforming desktop must support.

sxyazi:
Hmm I'm a bit dizzy but will merge it now anyway since matching all other terminal file managers seems like the right move (`vifm` is the only exception I know of) unless someone else points out it's actually wrong in the future ;)

Jaredy899:
Thanks! Looking at ```vifm```, it also shows up in both Utilities and System. 
Screenshot 2026-01-23 at 8 18 14 AM
Screenshot 2026-01-23 at 8 18 26 AM

I packaged yazi for Solus, so when I packaged it, I removed the Utility so it would only show up in one place. But I figured it would be better to fix it upstream.

sxyazi:
TIL about Solus, it looks pretty out-of-the-box which I appreciate, thanks for packaging Yazi for Solus!

If you'd like, you can add a Solus installation command at https://yazi-rs.github.io/docs/installation

Jaredy899:
I already did! Thanks!

sxyazi:
Ah indeed, I didn't even realize it was already there, that's awesome!

PR #3663, opened 2026-02-07
docs: update documentation in readme

ram-xv:
Enhanced the "Documentation" section and added helpful resources.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,15 +41,36 @@ Public beta, can be used as a daily driver.
 
 Yazi is currently in heavy development, expect breaking changes.
 
-## Documentation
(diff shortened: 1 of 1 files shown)


sxyazi:
I prefer the original format as this feels a bit vibe-coded, but in any case, thanks for your contribution!

ram-xv:
@sxyazi    def did not "vibe-code", I tried to match the original format:
image

LOL 

I guess we are at a point in time where we have to go out of our way 
to make it seem like PR changes have not been generated by AI.

but in any case, I love this tool so much and will try to contribute as much as I can.

Thank you for your work :)

Issue #3382, opened 2025-11-28
Custom sorting


grimandgreedy:
### `yazi --debug` output

```Shell

Yazi
    Version: 25.5.31 (Homebrew 2025-05-30)
    Debug  : false
    Triple : aarch64-apple-darwin (macos-aarch64)
    Rustc  : 1.87.0 (17067e9a 2025-05-09)

Ya
    Version: 25.5.31 (Homebrew 2025-05-30)

Emulator
    TERM                : Some("xterm-256color")
    TERM_PROGRAM        : Some("iTerm.app")
    TERM_PROGRAM_VERSION: Some("3.6.6")
    Brand.from_env      : Some(Iterm2)
    Emulator.detect     : Emulator { kind: Left(Iterm2), light: false, csi_16t: (0, 0), force_16t: false }

Adapter
    Adapter.matches    : Iip
    Dimension.available: Dimension { rows: 32, columns: 122, width: 1952, height: 1088 }

Desktop
    XDG_SESSION_TYPE           : None
    WAYLAND_DISPLAY            : None
    DISPLAY                    : Some("/private/tmp/com.apple.launchd.ybDwC1ilFo/org.xquartz:0")
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Variables
    SHELL           : Some("/bin/zsh")
    EDITOR          : Some("nvim")
    VISUAL          : None
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
    FZF_DEFAULT_OPTS: None

Text Opener
    default     : Some(OpenerRule { run: "nvim \"$@\"", block: true, orphan: false, desc: "NeoVim", for: None, spread: true })
    block-create: Some(OpenerRule { run: "nvim \"$@\"", block: true, orphan: false, desc: "NeoVim", for: None, spread: true })
    block-rename: Some(OpenerRule { run: "nvim \"$@\"", block: true, orphan: false, desc: "NeoVim", for: None, spread: true })

Multiplexers
    TMUX               : false
    tmux version       : No such file or directory (os error 2)
    tmux build flags   : enable-sixel=Unknown
    ZELLIJ_SESSION_NAME: None
    Zellij version     : No such file or directory (os error 2)

Dependencies
    file          : 5.41
[...]

PR #3669, opened 2026-02-09
feat: add `sort_dir_by` option for separate directory sorting

pilnyak:
I want to be able to sort directories in alphabetical order while sorting files by modification time - sorting folders by the latter is confusing, and I frequently want to mess with the last file I changed in a given directory. Right now `sort_by` applies to both, so I added a `sort_dir_by` option that lets directories have their own sort key.

## What it does

When set (and when `sort_dir_first = true`), directories sort by that key independently from files.

```toml
[mgr]
sort_by        = "mtime"
sort_reverse   = true
sort_dir_first = true
sort_dir_by    = "natural"   # dirs alphabetical, files by mtime
```

Also added a `,d` keybinding; a sort mode that combines alphabetical directory sorting with mtime file sorting. All other sort keys (`,m`, `,n`, etc.) clear the dir override and work as before.

## Changes

- `FilesSorter::sort()` partitions into dirs/files when `dir_by` is active, sorts each group separately
- Changing `sort_by` (via `,m` etc.) automatically clears `dir_by` so other modes aren't affected

### DIFF
--- a/yazi-actor/src/lives/preference.rs
+++ b/yazi-actor/src/lives/preference.rs
@@ -11,7 +11,8 @@ pub(super) struct Preference {
 	v_name:     Option<Value>,
 	v_linemode: Option<Value>,
 
-	v_sort_by: Option<Value>,
+	v_sort_by:     Option<Value>,
+	v_sort_dir_by: Option<Value>,
 }
 
 impl Deref for Preference {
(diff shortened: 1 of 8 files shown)


sxyazi:
There's no plan to build this in:

- `sort_dir_by` only works when `sort_dir_first = true`, and it's not clear what it should do when `sort_dir_first = false`.
- Adding `sort_dir_by` (to pair with `sort_by`) would mean we'd also need `sort_dir_sensitive`, `sort_dir_reverse`, `sort_dir_translit` to match `sort_sensitive`, `sort_reverse`, `sort_translit`, which would make sorting pretty complex
- [Custom sorting](https://github.com/sxyazi/yazi/issues/3382) would be a better way to handle use cases like this.

In any case, thank you for your effort!

Issue #3685, opened 2026-02-14
fd search showing path problem


og900aero:
### What system are you running Yazi on?

Linux Wayland

### What terminal are you running Yazi in?

Ghostty 1.2.3

### `yazi --debug` output

```Shell
Yazi
    Version: 26.2.2 (13bdb16 2026-02-13)
    Debug  : false
    Triple : x86_64-unknown-linux-gnu (linux-x86_64)
    Rustc  : 1.93.1 (01f6ddf7 2026-02-11)

Ya
    Version: 26.1.22 (4e0acf8 2026-01-22)

Config
    Init             : /home/shyciii/.config/yazi/init.lua (4592 chars)
    Yazi             : /home/shyciii/.config/yazi/yazi.toml (4592 chars)
    Keymap           : /home/shyciii/.config/yazi/keymap.toml (10914 chars)
    Theme            : /home/shyciii/.config/yazi/theme.toml (37348 chars)
    VFS              : /home/shyciii/.config/yazi/vfs.toml (No such file or directory (os error 2))
    Package          : /home/shyciii/.config/yazi/package.toml (1071 chars)
    Dark/light flavor: "" / ""

Emulator
    TERM                : Some("xterm-ghostty")
    TERM_PROGRAM        : Some("ghostty")
    TERM_PROGRAM_VERSION: Some("1.2.3")
    Brand.from_env      : Some(Ghostty)
    Emulator.detect     : Emulator { kind: Left(Ghostty), version: "ghostty 1.2.3", light: false, csi_16t: (9, 22), force_16t: false }

Adapter
    Adapter.matches    : Kgp
    Dimension.available: Dimension { rows: 53, columns: 212, width: 1916, height: 1166 }

Desktop
    XDG_SESSION_TYPE           : Some("wayland")
    WAYLAND_DISPLAY            : Some("wayland-1")
    DISPLAY                    : Some(":0")
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Variables
    SHELL              : Some("/bin/bash")
    EDITOR             : Some("micro")
    VISUAL             : Some("micro")
    YAZI_FILE_ONE      : None
    YAZI_CONFIG_HOME   : None
    YAZI_ZOXIDE_OPTS   : None
    SSH_AUTH_SOCK      : None
[...]

PR #3696, opened 2026-02-18
feat: new `--in` for `search` action to set search directory

sxyazi:
Fixes https://github.com/sxyazi/yazi/issues/3685

The `search` action uses the CWD (current working directory) as its search base, so if you have:

```sh
{ on = "<C-f>", run = 'search --via=fd --args="--search-path=/"' },
```

And hit `<C-f>` while in the `/root` directory, it won't work, because search results from `/root` can't include its parent directory `/`, but only files in itself or its subdirectories.

This PR adds a new `--in` option, allowing you set which directory to search in:

```sh
{ on = "<C-f>", run = 'search --in=/ --via=fd' },
```

This will open the search view of `/` and run the search under `/`, ignoring whatever your current working directory is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/):
 ### Added
 
 - Custom tab name ([#3666])
+- New `--in` for `search` action to set search directory ([#3696])
 - Allow using `ps.sub()` in `init.lua` directly without a plugin ([#3638])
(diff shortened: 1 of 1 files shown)


Goku-San:
Hi, just a question :) 
Will you push a new version with this feature and when ? 
Thanks for the option...

sxyazi:
Hi, we provide a [nightly build](https://github.com/sxyazi/yazi/releases/tag/nightly) that includes this feature, feel free to use it for now until the next release :)

Issue #1337, opened 2024-07-24
Cursor starts blinking after launching and exiting yazi


CyberShadow:
### What system are you running Yazi on?

Linux X11

### What terminal are you running Yazi in?

rxvt-unicode (urxvt) v9.31 - released: 2023-01-02

### `yazi --debug` output

<details>

```sh

Yazi
    Version: 0.2.5 (4a05226 1980-01-01)
    Debug  : true
    OS     : linux-x86_64 (unix)

Ya
    Version: No such file or directory (os error 2)

Emulator
    Emulator.via_env: ("rxvt-unicode-256color", "")
    Emulator.via_csi: Ok(Unknown([]))
    Emulator.detect : Urxvt

Adapter
    Adapter.matches: X11

Desktop
    XDG_SESSION_TYPE: Some("tty")
    WAYLAND_DISPLAY : None
    DISPLAY         : Some("localhost:18.0")

SSH
    shared.in_ssh_connection: true

WSL
    /proc/sys/fs/binfmt_misc/WSLInterop: false

Variables
    SHELL              : Some("/bin/zsh")
    EDITOR             : Some("/home/vladimir/bin/editor")
    ZELLIJ_SESSION_NAME: None
    YAZI_FILE_ONE      : None
    YAZI_CONFIG_HOME   : None

Text Opener
    default: Some(Opener { run: "${EDITOR:=vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })
    block  : Some(Opener { run: "${EDITOR:=vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for_: None, spread: true })

tmux
    TMUX   : false
    Version: 3.4

Dependencies
    file             : 5.45
    ueberzugpp       : No such file or directory (os error 2)
    ffmpegthumbnailer: No such file or directory (os error 2)
    magick           : 7.1.1-35
    fzf              : 0.54.0
    fd               : 10.1.0
    rg               : 14.1.0
    chafa            : No such file or directory (os error 2)
    zoxide           : No such file or directory (os error 2)
    unar             : No such file or directory (os error 2)
    jq               : 1.7.1


--------------------------------------------------
When reporting a bug, please also upload the `yazi.log` log file - only upload the most recent content by time.
You can find it in the "/home/vladimir/.local/state/yazi" directory.

```

</details>

### Did you try the latest nightly build to see if the problem got fixed?

[...]

PR #3698, opened 2026-02-18
fix: reset cursor to default in fallback

QuarticCat:

On Konsole, yazi incorrectly set my cursor to "block" at exit, while my previous cursor is "vertical bar + blink".

## Rationale of this PR

It seems like Konsole doesn't support both `"\x1bP$q q\x1b\\"` (request cursor shape) and `"\x1b[?12$p"` (request cursor blink status). But it supports `"\x1b[0 q"` (reset cursor to default). In this case, use `"\x1b[0 q"` as fallback seems to be more reasonable.

### DIFF
--- a/yazi-shim/src/crossterm/restore_cursor.rs
+++ b/yazi-shim/src/crossterm/restore_cursor.rs
@@ -18,9 +18,7 @@ impl crossterm::Command for RestoreCursor {
 			2 if !blink => SetCursorStyle::SteadyUnderScore.write_ansi(f)?,
 			3 if blink => SetCursorStyle::BlinkingBar.write_ansi(f)?,
 			3 if !blink => SetCursorStyle::SteadyBar.write_ansi(f)?,
-			_ if blink => SetCursorStyle::DefaultUserShape.write_ansi(f)?,
-			_ if !blink => SetCursorStyle::SteadyBlock.write_ansi(f)?,
-			_ => unreachable!(),
+			_ => SetCursorStyle::DefaultUserShape.write_ansi(f)?,
 		})
 	}
 


sxyazi:
Thanks for the patch!

`\x1b[0 q` means [setting the cursor to a blink block](https://vt100.net/docs/vt510-rm/DECSCUSR.html), not setting it to the default - crossterm's `DefaultUserShape` name is kinda confusing here, so we can't use it as a fallback since that would break other cases like steady block.

For terminals that don't support `DECSCUSR` and `DECSET 12`, a workaround is https://github.com/sxyazi/yazi/discussions/1612#discussioncomment-10575624

QuarticCat:
I'm not an expert in those escape sequences, but both Konsole and Ghostty say it's reset to default:

- https://invent.kde.org/utilities/konsole/-/blob/master/src/Vt102Emulation.cpp#L2323
- https://ghostty.org/docs/vt/csi/decscusr

QuarticCat:
Also, what's the rationale of using `SteadyBlock` when we don't know if it's blinking or not?

sxyazi:
Looks like Konsole and Ghostty each do it in their way, I'm not sure Konsole, but Ghostty [says](https://ghostty.org/docs/features#terminal-(vt)-features) it's xterm-compatible, so it should behave like xterm, i.e. a blinking block, `\x1b[0 q` seems to be an exception though, it might be worth filing as a bug.

I think we *should* follow xterm's behavior here since it's the de facto standard, rather than relying on terminal-specific behavior. Ideally terminals would support `DECSCUSR` and `DECSET 12` so Yazi can explicitly restore the previous cursor shape instead of reverting to the terminal-defined "default" cursor which depends on the terminal e.g. treating the default as the user's config instead of the cursor that was last used while entering Yazi.

> Also, what's the rationale of using `SteadyBlock` when we don't know if it's blinking or not?

IIRC that follows Vim's implementation - older Yazi used `DefaultUserShape`, but that [broke cursor behavior in `rxvt-unicode`](https://github.com/sxyazi/yazi/issues/1337), so I checked how Vim handled it and did something similar. But if there's a better way, I'm open to changing this behavior :)

QuarticCat:
> IIRC that follows Vim's implementation

That's pretty weird since I don't have this issue for Vim. I think Vim doesn't even try to change my cursor shape. Can we avoid that in Yazi as well?

sxyazi:
That would require:

- Disabling the input component's cursor changes that indicate the current mode - we'd need some other way to show users if they're in normal/insert/visual mode.
- Not restoring the user's cursor when exiting Yazi.

I'm not sure it's worth the effort - it's kinda niche yet there's already a workaround, and only happens if the user changed the default cursor *and* the terminal doesn't support `DECSCUSR` or `DECSET 12`. For the latter, I'd suggest adding support in Konsole directly (even `tmux` now [supports it](https://github.com/tmux/tmux/pull/4614)!), so other apps besides Yazi would benefit too.

That said, I'm not against giving it a shot, and if it doesn't add too much complexity I'd be happy to review a PR.

Issue #3395, opened 2025-12-02
Known issues with remote file management


sxyazi:
Known issues:

- [ ] Doesn't work well with interactive `cd` (<kbd>g</kbd> => <kbd>Space</kbd> by default)

They will be addressed in future releases.

boydaihungst:
I got this error when tried to download a remote file: `Cannot download non-remote file`. 

```toml
[providers.home]
  host     = "home"
  key_file = "~/.ssh/id_ecdsa"
  port     = 22
  type     = "sftp"
  user     = "huyhoang"
```


```
  download = [
    { run = 'ya emit download --open %s', desc = "Download and open" },
    { run = 'ya emit download %s', desc = "Download" },
  ]
```
Start command: `YAZI_LOG=debug yazi sftp://home`
Log file doesn't show any error.

Zevan770:
> Allow using an external `ssh` binary to support `~/.ssh/config` and more SSH options.

 FWIW, there is [russh-config](https://crates.io/crates/russh-config) for `~/.ssh/config`

CaptaiNiveau:
+1 for the ability to parse `~/.ssh/config`. That'd be insanely useful, especially with a fzf search across available connections. Is there a way to help this feature get released? I might be able to contribute with some guidance, if that's welcomed.

CaptaiNiveau:
Also, this would then hopefully allow proxy jumps, which would enable more usage scenarios, too.

sxyazi:
@CaptaiNiveau  Contributions are always welcome, thanks for your interest!

I haven't gotten around to digging deep into it yet, but the rough idea should probably be something like using [UnixSocket](https://docs.rs/tokio/latest/tokio/net/struct.UnixSocket.html) (does Windows have an equivalent?) to establish a connection with `ssh` instead of connecting directly to the remote host, so that `ssh` can parse its config file, you might want to check out how the [`--sftp-ssh`](https://rclone.org/sftp/#sftp-ssh) works in `rclone`.

[...]

PR #3716, opened 2026-02-25
feat: certificate authentication for SFTP VFS provider

itsvyle:

It adds yazi support for the ssh `CertificateFile` option, which allows users to authenticate using openssh certificates.

## Rationale of this PR

This PR is pretty simple - it simply uses the `russh` functionality for certificates. Most of the code is just a copy which follows the same pattern as the public key option.

I added a `cert_file` option to the vfs configuration, and if it is set, it will load the certificate and use it for authentication.

There is also a `no_cert_verify` option, which can be used to skip certificate validation. For now, certificate validation isn't really implemented (verification would just be checking the certificate dates, and maybe the signature) as per the `russh` documentation, but it can be added. I just haven't really looked into how to get the CAs, etc.

## Testing steps

<details>
<summary>Testing steps to make a mock certificate ssh server (generated by AI - they look fine to me, and worked when I tested them)</summary>

### Keys
1. Generate a Test Certificate Authority (CA) 
   ```bash
    ssh-keygen -t ed25519 -f ./test_ca -N ""
    ```   
1. Generate a User Key
    ```bash
    ssh-keygen -t ed25519 -f ./yazi_test_key -N ""
    ```
1. Sign the User Key with the CA to create a Certificate; the `-V +1h` can be changed if you want to test expired certificates, etc.
    ```bash
    ssh-keygen -s ./test_ca -I testing-yazi -n $USER -V +1h ./yazi_test_key.pub
    ```
1. Verify the Certificate
    ```bash
    ssh-keygen -L -f ./yazi_test_key-cert.pub
    ```
1. Configure vfs 
    - Set `cert_file` to the path of the generated certificate (e.g., `/tmp/ssh_server_test/yazi_test_key-cert.pub`)
    - Set `key_file` to the path of the generated user key (e.g., `/tmp/ssh_server_test/yazi_test_key`)
    - If using the mock server, set `host` to `127.0.0.1` and `port` to `2222`
    

### Mock ssh server
1. `mkdir -p /tmp/ssh_server_test && cd /tmp/ssh_server_test`
1. 
    Create a `sshd_config_test` file with the following content:
    ```bash
    Port 2222
    Protocol 2
    HostKey /tmp/ssh_server_test/test_ca
    TrustedUserCAKeys /tmp/ssh_server_test/test_ca.pub
    AuthorizedKeysFile /dev/null
    PasswordAuthentication no
    ChallengeResponseAuthentication no
    UsePAM no
    PidFile /tmp/ssh_server_test/ssh_server_test/sshd.pid
    Subsystem sftp internal-sftp
    ```
1. Start the ssh server
    ```bash
    sudo $(which sshd) -f /tmp/ssh_server_test/sshd_config_test -D -e
    ```

### vfs.toml file for mock server
```toml
[services.mock-ssh]
type = "sftp"
host = "127.0.0.1"
user = "<your-username>"
port = 2222
key_file =  "/tmp/ssh_server_test/yazi_test_key"
cert_file = "/tmp/ssh_server_test/yazi_test_key-cert.pub"
```

</details>

## Further development options (not necessarily in this PR)
- Add support for plugins which could be used to generate the certificate on the fly, or to fetch it from a vault, etc.
- Maybe at some point add support for `ProxyCommand` - this could be done via plugins maybe


itsvyle:
I noticed that you mentionned in #3395 that you were considering adding the option of using a custom ssh binary to support this kind of option.

But I still believe that this change here is small enough for it to be integrated as of now, before we manage to use the "default" openssh binary

sxyazi:
Thanks for the patch! 

Yeah, for fancier features I plan to support them by allowing the use of an external `ssh` binary, but bundling certificate authentication in sounds reasonable since it's easy to implement and shouldn't add much maintenance overhead - will review it.

yazi-config/src/vfs/service.rs
```diff
@@ -39,6 +39,10 @@ pub struct ServiceSftp {
 	pub key_file:       PathBuf,
 	pub key_passphrase: Option<String>,
 	#[serde(default)]
+	pub cert_file: PathBuf,
+	#[serde(default)]
+	pub skip_cert_validation: bool,
```
sxyazi:
```suggestion
	pub no_cert_verify: bool,
```

to make the name shorter

yazi-vfs/src/provider/sftp/conn.rs
```diff
@@ -146,6 +150,61 @@ impl Conn {
...
+			// TODO: Also validate the certificate cryptographically: https://docs.rs/russh/latest/russh/keys/struct.Certificate.html#method.validate
+			// however, this would require fetching CAs, which could require additional config and complexity
+
+			cert.verify_signature().map_err(|e| {
+				russh::Error::InvalidConfig(format!("Certificate signature verification failed: {e}"))
+			})?;
+			let unix_timestamp = std::time::UNIX_EPOCH.elapsed().unwrap_or_default().as_secs();
+			if !(cert.valid_after() <= unix_timestamp && unix_timestamp <= cert.valid_before()) {
+				return Err(russh::Error::InvalidConfig(
+					"Certificate is not valid at this time".to_owned(),
```
sxyazi: It would be great to include the valid date range of the cert

sxyazi (changes requested):

yazi-vfs/src/provider/sftp/conn.rs
itsvyle: I added a way to show in how long/how long ago the certificate will be/was valid - I haven't found in the yazi codebase a way to transform a timestamp into a formatted date

sxyazi: Maybe we can add [`chrono`](https://docs.rs/chrono/latest/chrono/) crate and use that instead to simplify the code some since it's already in the dep tree anyway

itsvyle: I added chrono

The error now looks like this:
image

The shown date should be in the system's timezone

sxyazi (approved):
Thank you

PR #3748, opened 2026-03-08
feat: new `ind-hidden` and `key-hidden` DDS events to change hidden status in Lua

IcEarthlight:

This PR introduces preflight hooks for the `hidden` command, similar to the existing `ind-sort` implementation.

## Rationale of this PR

Currently, dynamic visibility rules (e.g., "hide hidden files only in the Home directory") must rely on subscribing to the `cd` event and then reactively emitting a `hidden` command.

```lua
-- ~/.config/yazi/plugins/folder-rules.yazi/main.lua
local function setup()
	local home = os.getenv("HOME") or os.getenv("USERPROFILE")
	ps.sub("cd", function(opt)
		local cwd = cx.active.current.cwd
		if tostring(cwd.path) == home then
			ya.emit("hidden", { "hide" })
		else
			ya.emit("hidden", { "show" })
		end
	end)
end

return { setup = setup }
```

This approach is reactive rather than proactive. Yazi first renders the directory with the old hidden state and then re-renders it after the Lua script emits the new command, leading to a visible "flicker" of hidden files.

By adding `ind-hidden` and `key-hidden` preflight hooks, Lua plugins are allowed to intercept and modify the hidden command's parameters before execution. This enables atomic rendering, where the directory is displayed with the correct visibility state from the very first frame.

```lua
-- ~/.config/yazi/plugins/folder-rules.yazi/main.lua
local function setup()
	local home = os.getenv("HOME") or os.getenv("USERPROFILE")
	ps.sub("ind-hidden", function(opt)
		local cwd = cx.active.current.cwd
		if tostring(cwd.path) == home then
			opt.state = "hide"
		else
			opt.state = "show"
		end
		return opt
	end)
end

return { setup = setup }
```

## Changes
- `yazi-dds`: Added `KeyHidden` and `IndHidden` variants to `SparkKind`.
- `yazi-actor`: Implemented the `hook` method for the `Hidden` actor and ensured `cd` sets the source to `Ind` before calling `hidden`.
- `yazi-parser`: Added `Serialize`/`Deserialize` and Lua conversion support for `HiddenOpt` to allow parameter passing between Rust and Lua.

### DIFF
--- a/yazi-actor/src/mgr/cd.rs
+++ b/yazi-actor/src/mgr/cd.rs
@@ -50,7 +50,6 @@ impl Actor for Cd {
 
 		err!(Pubsub::pub_after_cd(tab.id, tab.cwd()));
 		act!(mgr:displace, cx)?;
-		cx.source = yazi_shared::Source::Ind;
 		act!(mgr:hidden, cx)?;
 		act!(mgr:sort, cx).ok();
 		act!(mgr:hover, cx)?;
--- a/yazi-actor/src/mgr/cd.rs
+++ b/yazi-actor/src/mgr/cd.rs
@@ -50,6 +50,7 @@ impl Actor for Cd {
 
 		err!(Pubsub::pub_after_cd(tab.id, tab.cwd()));
 		act!(mgr:displace, cx)?;
+		cx.source = yazi_shared::Source::Ind;
 		act!(mgr:hidden, cx)?;
 		act!(mgr:sort, cx).ok();
 		act!(mgr:hover, cx)?;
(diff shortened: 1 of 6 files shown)


yazi-actor/src/mgr/cd.rs
```diff
@@ -50,6 +50,7 @@ impl Actor for Cd {
 
 		err!(Pubsub::pub_after_cd(tab.id, tab.cwd()));
 		act!(mgr:displace, cx)?;
+		cx.source = yazi_shared::Source::Ind;
```
sxyazi: Why is this needed?

sxyazi (commented):
Thanks for the patch!

Overall looks good to me, some comments below:

yazi-actor/src/mgr/cd.rs
IcEarthlight: Oops, that's some debuging code ensuring the `ind-hidden` been triggered, which was accidentally left in. I've removed it in the new commit.

sxyazi (approved):
Thank you!

PR #3777, opened 2026-03-18
feat: add toggle support for sort boolean flags

### LINKED ISSUE #3157: Feature to toggle "sort_dir_first" by shortcut key
### `yazi --debug` output

```Shell
Yazi
    Version: 25.5.31 (Arch Linux 2025-07-23)
    Debug  : false
    Rustc  : 1.88.0 (6b00bc38 2025-06-23)

Ya
    Version: 25.5.31 (Arch Linux 2025-07-23)

Emulator
    TERM                : Some("xterm-256color")
    TERM_PROGRAM        : Some("terminology")
    TERM_PROGRAM_VERSION: Some("1.14.0")
    Brand.from_env      : None
    Emulator.detect     : Emulator { kind: Right(Unknown { kgp: false, sixel: false }), light: false, csi_16t: (0, 0), force_16t: false }

Adapter
    Adapter.matches    : X11
    Dimension.available: Dimension { rows: 40, columns: 147, width: 0, height: 0 }

Desktop
    XDG_SESSION_TYPE           : Some("x11")

SSH

WSL
    WSL: false

Variables
    EDITOR          : None
    VISUAL          : None
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
    FZF_DEFAULT_OPTS: None

Text Opener
    default     : Some(OpenerRule { run: "${EDITOR:-vim} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for: None, spread: true })
    block-create: Some(OpenerRule { run: "${EDITOR:-vim} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for: None, spread: true })
    block-rename: Some(OpenerRule { run: "${EDITOR:-vim} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for: None, spread: true })

Multiplexers
    TMUX               : false
    ZELLIJ_SESSION_NAME: None

Dependencies
    file          : 5.46
    ueberzugpp    : 2.9.7
    ffmpeg/ffprobe: 7.1.1 / 7.1.1
    pdftoppm      : 25.09.1
    magick        : 7.1.2-3
    fzf           : 0.65.2
    fd/fdfind     : 10.3.0 / No such file or directory (os error 2)
    rg            : 14.1.1
    chafa         : 1.16.2
    zoxide        : 0.9.8
    7zz/7z        : No such file or directory (os error 2) / 25.01
    jq            : 1.8.1

Clipboard
    xclip        : 0.13

Routine


```

Currently I added two shortcut key in `keymap.toml` to set the "dir_fist" behavior:
```
[mgr]
keymap = [
	{ on = [ ",", "f" ], run = "sort --dir-first=no",                           desc = "Unset dir_first sort" },
	{ on = [ ",", "F" ], run = "sort --dir-first",                              desc = "Set dir_first sort" },
]
```

May I know if it is possible to toggle this behavior by one shortcut key?

Add a `--dir-first=toggle` option to invert the status between `--dir-first=yes` and `--dir-first=no`. Also recommend same change for `--reverse` / `--translit`

### Additional context

_No response_

### Checklist

alastairsounds:
## Rationale
`sort`'s boolean flags (`--reverse`, `--dir-first`, `--translit`) had no toggle support. This adds `toggle` as a valid value for each, mirroring the existing `hidden toggle` command.

```
sort --reverse=toggle
sort --dir-first=toggle
sort --translit=toggle
```

## Changes
- **`yazi-parser`**: Adds `SortBoolState` enum (`none` / `on` / `off` / `toggle`), replacing `bool` for the three flags. Modelled on `HiddenOptState`.
- **`yazi-actor`**: Sort actor now calls `.bool(old)` so `none` preserves and `toggle` inverts current state.

### DIFF
--- a/yazi-parser/src/mgr/sort.rs
+++ b/yazi-parser/src/mgr/sort.rs
@@ -1,16 +1,19 @@
+use std::str::FromStr;
+
+use anyhow::bail;
 use mlua::{FromLua, IntoLua, Lua, LuaSerdeExt, Value};
 use serde::{Deserialize, Serialize};
 use yazi_binding::SER_OPT;
 use yazi_fs::{SortBy, SortFallback};
-use yazi_shared::event::ActionCow;
+use yazi_shared::{data::Data, event::ActionCow};
 
 #[derive(Debug, Default, Deserialize, Serialize)]
 pub struct SortOpt {
 	pub by:        Option<SortBy>,
-	pub reverse:   Option<bool>,
-	pub dir_first: Option<bool>,
+	pub reverse:   Option<SortBoolState>,
+	pub dir_first: Option<SortBoolState>,
 	pub sensitive: Option<bool>,
-	pub translit:  Option<bool>,
+	pub translit:  Option<SortBoolState>,
 	pub fallback:  Option<SortFallback>,
 }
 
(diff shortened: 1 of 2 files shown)


alastairsounds:
Also wondering if I should implement this for the `sensitive` sort? But figure there's a little less utility in toggling case sensitivity 🤷🏻‍♂️

yazi-parser/src/mgr/sort.rs
```diff
@@ -36,3 +39,46 @@ impl FromLua for SortOpt {
...
+}
+
+impl TryFrom<&Data> for SortBoolState {
+	type Error = anyhow::Error;
+
+	fn try_from(value: &Data) -> Result<Self, Self::Error> {
+		match value {
+			Data::Boolean(true) => Ok(Self::On),
+			Data::Boolean(false) => Ok(Self::Off),
+			Data::String(s) => s.parse().map_err(Into::into),
```
sxyazi: This treats both boolean and string as `SortBoolState`, but it isn't clear whether `SortBoolState` should be serialized as a boolean or a string in Lua, which was in use by https://yazi-rs.github.io/docs/tips#folder-rules

alastairsounds:
I see what you're saying. I think it should be serialized as a boolean and not handle strings. I don't see a scenario in which a user would need to set `"toggle"` via a Lua script. Since it's not really a "state" per-se, I've updated `Sort::hook()` to account for that. Now, when any of `reverse`, `dir_first`, or `translit` contains `Toggle`, it returns `None` to skip preflight entirely.

https://github.com/sxyazi/yazi/pull/3777/changes#diff-608cad97a235673889398f55a252b08703746ca6bce53ee32ebaabe78294f9cbR62-R70

With this revision, Lua shouldn't be able to see `"toggle"`. Sorts that occur from plugins should work normally since they're only handling booleans and not strings.

Toggle being a CLI feature, not Lua, seems like a clean implementation. Or might there be a better route to follow?

PR #3843, opened 2026-04-02
feat(s3): add S3 bucket browsing as a VFS provider

fecet:
## Summary

- Implement a read-only S3 provider built on the `object_store` crate, allowing users to browse S3-compatible buckets natively within yazi
- Add `ServiceS3` config with support for region, endpoint, credentials, `force_path_style`, and `allow_http`
- Extend `Url` enum and `SchemeKind` with an `S3` variant; implement `read_dir` (paginated listing), `metadata`, and streaming read

## Screenshot
image

## Details

This builds on the VFS provider infrastructure from #3821. The S3 provider is currently **read-only** — write operations (`copy`, `create`, `create_dir`, etc.) return `Unsupported`.

Configuration example in `yazi.toml`:

```toml
[[vfs.services]]
name = "my-s3"
type = "s3"
region = "us-east-1"
endpoint = "https://s3.amazonaws.com"
# access_key_id = "..."
# secret_access_key = "..."
# force_path_style = false
# allow_http = false
```

Then browse via `s3://my-s3/bucket-name/path/`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -226,6 +226,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-trait"
+version = "0.1.89"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9035ad2d096bed7955a320ee7e2230574d28fd3c3a0f186cbea1ff3c7eed5dbb"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.117",
+]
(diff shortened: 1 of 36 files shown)


sxyazi:
Closing in favor of the integration of `rclong` (tracked in https://github.com/sxyazi/yazi/issues/3395) as we discussed on Telegram, in any case, thanks for your effort!

PR #3855, opened 2026-04-05
fix: avoid ffmpeg no-frame error when seeking with nokey

leson207:
## Problem
Using -ss with -skip_frame nokey sometimes causes:
"No filtered frames for output stream"

## Cause
Seeking may land on non-keyframe → nokey skips all frames ("No more keyframes available" err)

## Fix
- disable skip_frame, nokey when seeking

## Result
Stable frame extraction without empty output

### DIFF
--- a/yazi-plugin/preset/plugins/video.lua
+++ b/yazi-plugin/preset/plugins/video.lua
@@ -46,7 +46,7 @@ function M:preload(job)
 	end
 
 	local pic = M.has_pic(meta)
-	local percent = (pic and 0 or 5) + job.skip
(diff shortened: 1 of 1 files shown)


sxyazi:
How can I reproduce the problem? 

It would be great to first discuss it in a bug report that lists the exact steps of the reproduction.

leson207:
i just create an issue

https://github.com/sxyazi/yazi/issues/3858

sxyazi:
Closing with https://github.com/sxyazi/yazi/issues/3858 due to no follow-up

PR #3875, opened 2026-04-11
use kitty image protocol for Rio terminal

raphamorim:
Screenshot 2026-04-11 at 09 30 52

### DIFF
--- a/yazi-adapter/src/adapters.rs
+++ b/yazi-adapter/src/adapters.rs
@@ -33,7 +33,7 @@ impl From<yazi_emulator::Brand> for Adapters {
 			B::Ghostty => vec![A::Kgp],
 			B::Microsoft => vec![A::Sixel],
 			B::Warp => vec![A::Iip, A::KgpOld],
-			B::Rio => vec![A::Iip, A::Sixel],
+			B::Rio => vec![A::Kgp],
 			B::BlackBox => vec![A::Sixel],
 			B::VSCode => vec![A::Iip, A::Sixel],
 			B::Tabby => vec![A::Iip, A::Sixel],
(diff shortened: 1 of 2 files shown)


yazi-adapter/src/adapters.rs
```diff
@@ -33,7 +33,7 @@ impl From<yazi_emulator::Brand> for Adapters {
 			B::Ghostty => vec![A::Kgp],
 			B::Microsoft => vec![A::Sixel],
 			B::Warp => vec![A::Iip, A::KgpOld],
-			B::Rio => vec![A::Iip, A::Sixel],
+			B::Rio => vec![A::Kgp],
```
sxyazi, line 36: Has Rio dropped the support of the inline image protocol and Sixel? If not, it would be nice to keep them in the list of supported protocols.

sxyazi (changes requested):
Thanks for the patch!

yazi-adapter/src/adapters.rs
raphamorim: Sixel is unstable, I plan to rework on it. Then I will add it back 🙏

raphamorim: I mean we can keep sixel as well, but might gonna have that bug with dimensions for few versions

sxyazi: 1. Is IIP also unstable and buggy?
2. Was Kgp supported in a recent version? If so it would be great to note that on the README about the required minimum version of Rio.

raphamorim: I will readd both. And do a proper fix for sixel. I just have been pushing it 😭 😂 

lemme update the pr, also release a fix on rio before we merge

sxyazi:
I've updated the code and README in the latest commit

raphamorim:
Thank you!

PR #3890, opened 2026-04-15
feat: drag-resize panes with mouse

winlaic:
Implements **tmux-style mouse dragging** of the panel separators to dynamically resize the parent, current, and preview panels (see the gif below).

example

### `yazi-config/preset/yazi-default.toml`
- Added `"drag"` to the default `mouse_events` list so drag handling is enabled by default.

### `yazi-actor/src/app/mouse.rs`
- Added `render!()` after the `Root:drag()` Lua call so the UI repaints immediately on every drag event.
- Without this, dragging updated the layout state but the resize was not visible until some unrelated event triggered a redraw.

### `yazi-plugin/preset/components/root.lua`
- Added a `_drag_which` field (`nil` | `"left"` | `"right"`) to track which separator is currently being dragged.
- Extended `Root:click()` to detect mouse-down on either separator (with a ±1 column tolerance), set `_drag_which`, and clear it again on mouse-up.
- Implemented `Root:drag()` to read the current panel widths from `tab._chunks`, compute the new widths from the cursor position, and write the updated all-pixel-width array back to `rt.mgr.ratio`.

## How it works

```text
mouse down near separator  ->  Root:click()  sets _drag_which
mouse drag                 ->  Root:drag()   updates rt.mgr.ratio + render!()
mouse up                   ->  Root:click()  clears _drag_which
```

The separator hit zones are derived from tab._chunks, which contains the three column rects produced by Tab:layout()
Each separator uses a ±1 column tolerance and is only active when the corresponding side panel is visible (width > 0).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/):
 - Custom tab name ([#3666])
 - New `--in` for `search` action to set search directory ([#3696])
 - Hover cursor over the new file after copying/cutting/linking/hardlinking/extracting ([#3846], [#3854])
+- Drag-resize panes with mouse ([#3890])
 - Multi-file spotter ([#3733])
 - New `app:theme` action that hot-reload user themes/flavors ([#3906])
 - Dynamic open/opener Lua API ([#3901])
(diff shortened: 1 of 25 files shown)


sxyazi (approved):
Thank you!

winlaic:
> Thank you!

感谢merge😄，这个鼠标拖动的功能一直想加了～
另外yazi确实相当好用👍，一直喜欢折腾各种工具，好多都是折腾完就吃灰了😂，但是yazi一直用了两年了还是碰到新环境上来就装，已经离不开了

sxyazi:
Thanks for your kind words, it's huge!

PR #3906, opened 2026-04-21
feat: new `app:theme` action that hot-reload user themes/flavors

sxyazi:
## Demo

```toml
# ~/.config/yazi/keymap.toml
[[mgr.prepend_keymap]]
on   = "<C-t>"
run  = "app:theme"
desc = "Reload theme"
```

https://github.com/user-attachments/assets/abcc0950-e9ff-43dd-8c50-7efdf27b0320


hankertrix:
This doesn't seem to work with light and dark transitions. Is that intended?

sxyazi:
What do you mean by “light and dark transitions“?

hankertrix:
When the system theme changes from light mode to dark mode. Or in Yazi's case, when the terminal theme changes.

The idea behind my use case was to use `ya emit-to` to emit the `app:theme` command to all instances to update their themes when the system theme changes, like so:
```sh
ya emit-to 0 'app:theme'
```

The flavour used stays the same, I just want the theme to update so it doesn't look out of place with the rest of my system.

sxyazi:
Right now we only support querying the terminal's instantaneous background at startup via `OSC 11` and using that to pick the dark or light flavor. 

To follow system/terminal dark/light mode switches we'd need [`OSC ? 2031`](https://contour-terminal.org/vt-extensions/color-palette-update-notifications/#request-unsolicited-dsr-on-color-palette-updates), but Crossterm (the terminal lib driving Yazi) [doesn't support it](https://github.com/crossterm-rs/crossterm/pull/1045). I've been looking for Crossterm alternatives lately since it's a bit unmaintained and we've hit a bunch of issues with it in the past, but that would be quite a bit of work to get there.

hankertrix:
Ohhh I see. No worries, just wondering if terminal theme changes are supported. Seems like it would take quite a bit to get there.

PR #3912, opened 2026-04-25
fix: use cspell package in devShell

yadokani389:
## Rationale of this PR

`nix develop` is currently broken with the nixpkgs revision pinned in `flake.lock` because the devShell still uses `nodePackages.cspell`.
This follows upstream nixpkgs changes: `nodePackages` was dropped in NixOS/nixpkgs#496365, and `cspell` was migrated out of `nodePackages` in NixOS/nixpkgs@f3fa5ff106d9b6b7a48a5bdb0348155a17e3f088. Switching to the top-level `cspell` package restores the devShell.

### DIFF
--- a/nix/shell.nix
+++ b/nix/shell.nix
@@ -2,7 +2,7 @@
   mkShell,
   yazi,
   toolchain,
-  nodePackages,
+  cspell,
   yazi-unwrapped,
 }:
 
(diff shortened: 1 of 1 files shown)


sxyazi (approved):
LGTM, thank you!

PR #3946, opened 2026-05-08
fix: keep repeated `!` negations in a condition right-associative

immanuwell:

No direct issue found. Searched issues/PRs for `!!dir`, `condition parser`, and `theme condition`.

## Rationale of this PR

`!` is unary, so repeated negation should stack right-to-left. On main, parsing `!!dir` errors with `Invalid condition: !!dir`, while `!dir` works. a bit goofy for theme conditions, ngl.

Repro:
- run the included `!!dir` test
- before: `cargo +nightly test -p yazi-shared condition::tests::test_condition_not` fails
- after: it passes

This keeps `&` and `|` left-assoc, and only lets `!` stack properly. tiny fix, no big refactor.

### DIFF
--- a/yazi-shared/src/condition.rs
+++ b/yazi-shared/src/condition.rs
@@ -1,4 +1,4 @@
-use std::str::FromStr;
+use std::{cmp::Ordering, str::FromStr};
 
 use anyhow::bail;
 use serde_with::DeserializeFromStr;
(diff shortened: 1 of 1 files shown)


sxyazi:
Thanks for the patch!

Is there a use case of using `!!dir` instead of just `dir`?

sxyazi (approved):
Thank you!

Issue #3619, opened 2026-01-27
Impossible to install if a `CARGO_TARGET_DIR` is configured


Zykino:
### What system are you running Yazi on?

Linux Wayland

### What terminal are you running Yazi in?

alacritty

### `yazi --debug` output

```Shell
Yazi
    Version: 26.1.22 (Crates.io 2026-01-25)
    Debug  : false
    Triple : x86_64-unknown-linux-gnu (linux-x86_64)
    Rustc  : 1.92.0 (ded5c06c 2025-12-08)

Ya
    Version: 26.1.4 (Crates.io 2026-01-25)

Config
    Init             : /home/zykino/.config/yazi/init.lua (185 chars)
    Yazi             : /home/zykino/.config/yazi/yazi.toml (126 chars)
    Keymap           : /home/zykino/.config/yazi/keymap.toml (397 chars)
    Theme            : /home/zykino/.config/yazi/theme.toml (62 chars)
    VFS              : /home/zykino/.config/yazi/vfs.toml (No such file or directory (os error 2))
    Package          : /home/zykino/.config/yazi/package.toml (655 chars)
    Dark/light flavor: "catppuccin-mocha" / "catppuccin-latte"

Emulator
    TERM                : Some("xterm-256color")
    TERM_PROGRAM        : None
    TERM_PROGRAM_VERSION: None
    Brand.from_env      : None
    Emulator.detect     : Emulator { kind: Right(Unknown { kgp: false, sixel: true }), version: "Zellij(4301)", light: false, csi_16t: (0, 0), force_16t: false }

Adapter
    Adapter.matches    : Sixel
    Dimension.available: Dimension { rows: 21, columns: 94, width: 0, height: 0 }

Desktop
    XDG_SESSION_TYPE           : Some("x11")
    WAYLAND_DISPLAY            : None
    DISPLAY                    : Some(":0")
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Variables
    SHELL              : Some("/bin/bash")
    EDITOR             : Some("hx")
    VISUAL             : None
    YAZI_FILE_ONE      : None
    YAZI_CONFIG_HOME   : None
    YAZI_ZOXIDE_OPTS   : None
    FZF_DEFAULT_OPTS   : None
    FZF_DEFAULT_COMMAND: None

Text Opener
[...]

Issue #2934, opened 2025-06-30
Yazi Doesn't Work on aarch64 16K Page


sallyhaj:
### What system are you running Yazi on?

Linux Wayland

### What terminal are you running Yazi in?

alacritty

### `yazi --debug` output

```Shell
yazi --debug
<jemalloc>: Unsupported system page size                                                                          <jemalloc>: Unsupported system page size                                                                    memory allocation of 16 bytes failed
Aborted
```

### Describe the bug

Hello developers,

I have installed `yazi` via `cargo install --locked yazi-fm yazi-cli` , but when try to run it, it refuses and complains about the memory allocation. I think this issue related to one other dependency called `jemalloc`. 
When I try to run `yazi` I get this error:
```
yazi 
<jemalloc>: Unsupported system page size
<jemalloc>: Unsupported system page size
memory allocation of 16 bytes failed
Aborted

```
Then, I have tried the pre-build package `yazi-aarch64-unknown-linux-gnu.zip` which works just fine. 
Does that mean, the issue within the `yazi` inside `crates.io`? And, do I have to report this issue there in crates.io?

Thank you.

### Minimal reproducer

I think the `jemalloc` is a retired project. Any plan to replace it with other tool?

### Checklist

- [ ] I tried the [latest nightly build](https://yazi-rs.github.io/docs/installation#binaries), and the issue is still reproducible
- [x] I updated the debug information (`yazi --debug`) input box to the nightly that I tried
- [x] I can reproduce it after disabling all custom configs/plugins (`mv ~/.config/yazi ~/.config/yazi-backup`)

Issue #2898, opened 2025-06-20
Cargo install failed


Aiontropy:
### What system are you running Yazi on?

Windows

### What terminal are you running Yazi in?

Nushell

### `yazi --debug` output

```Shell
cargo install --locked --profile release-windows yazi-fm yazi-cli
    Updating `aliyun` index
  Installing yazi-fm v25.5.31
    Updating `aliyun` index
    Updating `aliyun` index
  Installing yazi-cli v25.5.31
    Updating `aliyun` index
    Updating `aliyun` index
error: failed to compile `yazi-fm v25.5.31`, intermediate artifacts can be found at `C:\Users\Aiont\AppData\Local\Temp\cargo-installXCgawF`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.

Caused by:
  profile `release-windows` is not defined
error: failed to compile `yazi-cli v25.5.31`, intermediate artifacts can be found at `C:\Users\Aiont\AppData\Local\Temp\cargo-installmCCN7t`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.

Caused by:
  profile `release-windows` is not defined
     Summary Failed to install yazi-fm, yazi-cli (see error(s) above).
error: some crates failed to install
```

### Describe the bug

I want to install Yazi by cargo.

 cargo --version
cargo 1.87.0 (99624be96 2025-05-06)
 rustc --version
rustc 1.87.0 (17067e9ac 2025-05-09)
rustup --version
rustup 1.28.2 (e4f3ad6f8 2025-04-28)

And I'm not sure that's the problem, it is my rust installed by MinGW

### Minimal reproducer

 cargo install --locked --profile release-windows yazi-fm yazi-cli

### Checklist

- [x] I tried the [latest nightly build](https://yazi-rs.github.io/docs/installation#binaries), and the issue is still reproducible
- [x] I updated the debug information (`yazi --debug`) input box to the nightly that I tried
- [x] I can reproduce it after disabling all custom configs/plugins (`mv ~/.config/yazi ~/.config/yazi-backup`)

Aiontropy:
> Hey [@Aiontropy](https://github.com/Aiontropy), I noticed that you did not correctly follow the issue template. Please ensure that:
> 
>     * The bug can still be reproduced on the [newest nightly build](https://yazi-rs.github.io/docs/installation/#binaries).
> 
[...]

PR #3952, opened 2026-05-09
fix: get rid of the Cargo curse

sxyazi:
I've run into a bunch of problems caused by Cargo when installing the Yazi package from crates.io:

- Cargo ignores the project's `.cargo/config.toml` - https://github.com/sxyazi/yazi/issues/2934
- Cargo ignores workspace profile settings - https://github.com/sxyazi/yazi/issues/2898

So I created a new `yazi-build` crate and used a custom `build.rs` to work around those issues, but that introduced other problems:

- If a user sets `CARGO_TARGET_DIR` it can deadlock - https://github.com/sxyazi/yazi/issues/3619

I've spent a lot of time constantly working around these issues, yet other problems popped up:

- `include_bytes!()` won't work for files from other workspace crates - https://github.com/sxyazi/yazi/discussions/3936

It's like I've fallen down an endless rabbit hole.

Let's break the curse: until Cargo gives proper workspace support for crates.io, just have our little `build.rs` run `git clone` and `cargo build` to manually install the binary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1069,7 +1069,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/):
 [v25.12.29]: https://github.com/sxyazi/yazi/compare/v25.5.31...v25.12.29
 [v26.1.4]: https://github.com/sxyazi/yazi/compare/v25.12.29...v26.1.4
 [v26.1.22]: https://github.com/sxyazi/yazi/compare/v26.1.4...v26.1.22
-[v26.5.6]: https://github.com/sxyazi/yazi/compare/v26.5.6...v26.5.6
+[v26.5.6]: https://github.com/sxyazi/yazi/compare/v26.1.22...v26.5.6
 [#4]: https://github.com/sxyazi/yazi/pull/4
 [#5]: https://github.com/sxyazi/yazi/pull/5
 [#6]: https://github.com/sxyazi/yazi/pull/6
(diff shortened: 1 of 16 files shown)


tessus:
> It's like I've fallen down an endless rabbit hole.

I hear you. I've opened https://github.com/rust-lang/cargo/issues/15467 a while back, but instead of creating a build workaround like `yazi-build` I've accepted the fact that I have to build from a source tree. Since `cargo install` requires connectivity anyway, it doesn't matter whether I use `cargo install` or a short bash script to build (and check whether a new version is available and a build is required).

My solution (adaptable for any rust project really):

```sh
function build_yazi {
	YWD=~/build/yazi
	[[ $(command -v yazi) ]] && YCV=$(yazi --version |awk '{print $2}') || YCV=0.0.0
	YNV=$(curl -Ls https://api.github.com/repos/sxyazi/yazi/releases/latest |jq -r .name |cut -c2-)
	if [[ "$YCV" != "$YNV" ]]; then
		[[ -d $YWD ]] && rm -rf $YWD
		cd ~/work
		git clone https://github.com/sxyazi/yazi.git
		cd $YWD
		git checkout v$YNV
		cargo +stable build --release --locked
		if [ "$?" == "0" ]; then
			cp -f ${YWD}/target/release/{ya,yazi} ~/bin
		else
			notify "Error: yazi build failed." 1
		fi
	else
		echo "yazi ($YNV) already installed."
	fi
}
```

Issue #3969, opened 2026-05-17
Allow arrow movements relative to current visible window


SpiritCroc:
### `yazi --debug` output

```Shell
Yazi
    Version  : 26.5.6 (Arch Linux 2026-05-10)
    Debug    : false
    Triple   : x86_64-unknown-linux-gnu (linux-x86_64)
    Rustc    : 1.95.0 (59807616 2026-04-14)
    Backtrace: None

Ya
    Version: 26.5.6 (Arch Linux 2026-05-10)

[...]
```

### Please describe the problem you're trying to solve

I got used to moving my cursor with "H"/"M"/"L" key bindings in applications supporting vim-like movements. These would go to the top/center/bottom of the currently visible frame, rather than relative to the open file. It would be nice to have these available in yazi too.

### Would you be willing to contribute this feature?

- [x] Yes, I'll give it a shot

### Describe the solution you'd like

Relative movements could e.g. be configured as follows, once the feature exists:
```
{ on = "<S-H>", run = "arrow 0%-window", desc = "High: Move cursor to visible window top without scrolling" },
{ on = "<S-M>", run = "arrow 50%-window", desc = "Middle: Move cursor to visible window center" },
{ on = "<S-L>", run = "arrow 100%-window", desc = "Low: Move cursor to visible window bottom without scrolling" },
```

### Checklist

- [x] I have searched the existing issues/discussions
- [x] The [latest nightly build](https://yazi-rs.github.io/docs/installation/#binaries) doesn't already have this feature

sxyazi:
Done in https://github.com/sxyazi/yazi/pull/3970, ty!

PR #3970, opened 2026-05-17
feat: H/M/L Vim-like motion for moving cursor relative to viewport

SpiritCroc:
Introduce a new `vp` step unit, that allows you to move the cursor to relative position based on the items currently rendered on viewport, similar as H/M/L ("high", "middle", "low") would do in Vim.

Not added to default keybinding config since there are some conflicts, but you can put following to override:

```sh
{ on = "H", run = "arrow 0vp",   desc = "High: Move cursor to visible viewport top" },
{ on = "M", run = "arrow 50vp",  desc = "Middle: Move cursor to visible viewport center" },
{ on = "L", run = "arrow 100vp", desc = "Low: Move cursor to visible viewport bottom" },
```

https://github.com/sxyazi/yazi/issues/3969

## Rationale of this PR

I got used to moving my cursor with "H"/"M"/"L" key bindings in applications supporting Vim-like motions. Since switching to yazi recently it has been constantly tripping me, that it is not possible to do.

## AI disclaimer

I did use AI to write this for me. This means while I did read the code, question some AI decisions and iterated on them, and of course tested this myself, it's fair to point out I'm not an expert in this codebase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/):
 - Drag and drop ([#4005])
 - Bulk create ([#3793])
 - Make help menu a command palette ([#4074])
+- H/M/L Vim-like motion for moving cursor relative to viewport ([#3970])
 - Dynamic keymap Lua API ([#4031])
 - New `ui.Input` element ([#4040])
 - Image preview with Überzug++ on Niri ([#3990])
@@ -1749,6 +1750,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/):
 [#3906]: https://github.com/sxyazi/yazi/pull/3906
 [#3934]: https://github.com/sxyazi/yazi/pull/3934
 [#3943]: https://github.com/sxyazi/yazi/pull/3943
+[#3970]: https://github.com/sxyazi/yazi/pull/3970
 [#3989]: https://github.com/sxyazi/yazi/pull/3989
 [#3990]: https://github.com/sxyazi/yazi/pull/3990
 [#4005]: https://github.com/sxyazi/yazi/pull/4005
--- a/yazi-actor/src/confirm/arrow.rs
+++ b/yazi-actor/src/confirm/arrow.rs
@@ -19,7 +19,7 @@ impl Actor for Arrow {
 		let len = confirm.list.line_count(area.width);
 
 		let old = confirm.offset;
-		confirm.offset = form.step.add(confirm.offset, len, area.height as _);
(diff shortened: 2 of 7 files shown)


yazi-actor/src/mgr/arrow.rs
```diff
@@ -27,7 +29,11 @@ impl Actor for Arrow {
 			*items = (start.min(end)..=end.max(start)).collect();
 		}
 
-		act!(mgr:hover, cx)?;
+		if window_relative {
+			err!(Pubsub::pub_after_hover(tab.id, tab.hovered().map(|h| &h.url)));
+		} else {
+			act!(mgr:hover, cx)?;
+		}
```
sxyazi: Why is this change needed?

yazi-core/src/tab/folder.rs
```diff
@@ -93,13 +93,18 @@ impl Folder {
...
 		let mut b = if self.files.is_empty() {
 			(mem::take(&mut self.cursor), mem::take(&mut self.offset)) != (0, 0)
 		} else {
 			self.scroll(step)
 		};
 
-		b |= self.squeeze_offset();
+		if !window_relative {
+			b |= self.squeeze_offset();
+		}
```
sxyazi: Why is this change needed?

sxyazi:
This implementation looks wrong as it doesn't respect [`scrolloff`](https://yazi-rs.github.io/docs/configuration/yazi#mgr.scrolloff), e.g. with `scrolloff=5`:

- `H` should jump to the 6th line within the window, not to the first line
- similarly, `L` should jump to the 6th-from-last line, not the last line.

Here's how Neovim behaves for `H` and `L` when you set `:set scrolloff=5`:


https://github.com/user-attachments/assets/35faf9f1-80d5-41ee-b8c6-0438a2cee9f2

The same behavior can be reproduced in Vim too.

SpiritCroc:
Ohhhh that makes sense, I never used scrolloff on (neo)vim, so it was off-putting for me and I worked around it for this change, sounds like I can simplify the patch then :+1:

yazi-actor/src/mgr/arrow.rs
SpiritCroc: To work around scrolloff, see my misconception below, my bad

yazi-core/src/tab/folder.rs
SpiritCroc: I don't think it is :pray:

SpiritCroc:
Hope I got the behaviour correct now.
BTW Thank you for the quick feedback, and for creating/maintaining yazi in the first place, greatly appreciated!

sxyazi (approved):
ty!

PR #3986, opened 2026-05-21
Open file on right arrow

apaz-cli:
Ranger opens files on right arrow, which is the primary way I navigate. With Yazi you have to press enter.

If you don't want this or think it should be behind an option, feel free to reject.


sxyazi:
That's by design not to do so, see https://yazi-rs.github.io/docs/faq#why-separate-open-enter, but thanks for your effort!

apaz-cli:
Ah, gotcha, ty :)

PR #3988, opened 2026-05-22
Double Click Support

Emrebener:

// no existing GitHub issue

Resolves #

## Rationale of this PR

Yazi supports clicking a file to select/reveal it, so the mouse support is there. However, running/opening a file or navigating to a directory still requires a follow-up Enter keypress (or a right-click, which is only for directories and not for files).

In my opinion, if Yazi supports mouse clicking to select files, it might as well support double clicks as well. I see no reason to not go a step further.

In fact, it's confusing. "I can click to select a file, but I can't double-click to open it??"

Double-click is the conventional gesture that every file manager supports, and it's natural for people to expect this behaviour. I think adding this feature can help GUI users migrate to and adopt a terminal-based workflow and help increase Yazi's popularity.

--- technical explanations below ---

## Behavior

On a left-button mouse-down in the current pane:

If the same file URL was clicked within the configured threshold → emit open (the same command Enter runs). Files open in the default opener; directories are entered.

Otherwise → existing behavior (reveal the entry). The click is recorded as the new "last click".

Right-click, middle-click, and the parent pane are intentionally untouched. The parent pane already has its own single-click semantics (empty-area click emits leave), so adding double-click there would conflict.

## Configuration

New optional key under [mgr] in yazi.toml:

```
mouse_double_click_delay = 300  # milliseconds; 0 disables
```

Default 300 matches common OS thresholds. Setting 0 disables the feature.

## Implementation notes

Detection lives in Current:click (Lua), not in the Rust mouse actor — a module-local last_click = { time, url } table persists across redraws.

The config field (SyncCell) is exposed read-only to Lua via the existing rt.mgr.* composer, alongside scrolloff, mouse_events, etc.

Wall-clock timing uses ya.time(), matching the pattern in yazi-plugin/preset/plugins/mime-local.lua.

After firing open, last_click.time is reset so a third quick click doesn't chain into another open.

~25-line diff across three commits.

## Tested

It works nicely in my experience.

Double-click on file opens it; double-click on dir enters it; slow clicks don't open; clicks on different files don't open; triple-click opens once; right-click and parent-pane behavior unchanged; delay = 0 disables; delay = 600 honors the wider window.

### DIFF
--- a/yazi-config/preset/yazi-default.toml
+++ b/yazi-config/preset/yazi-default.toml
@@ -11,11 +11,12 @@ sort_reverse 	 = false
 sort_dir_first = true
 sort_translit  = false
 sort_fallback  = "alphabetical"
-linemode       = "none"
-show_hidden    = false
-show_symlink   = true
-scrolloff      = 5
-mouse_events   = [ "click", "scroll", "drag" ]
+linemode                    = "none"
+show_hidden                 = false
+show_symlink                = true
+scrolloff                   = 5
+mouse_events                = [ "click", "scroll", "drag" ]
+mouse_double_click_delay    = 300
 
 [preview]
 wrap            = "no"
--- a/yazi-config/src/mgr/mgr.rs
+++ b/yazi-config/src/mgr/mgr.rs
@@ -22,10 +22,11 @@ pub struct Mgr {
 	// Display
 	#[serde(deserialize_with = "deserialize_linemode")]
 	pub linemode:     ArcSwap<String>,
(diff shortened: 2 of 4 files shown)


sxyazi:
There's no plan to support double-click to open files:

- Yazi already supports right-click to open files, which seems redundant here
- This can already be done through custom click events, see https://github.com/sxyazi/yazi/pull/2925
- Not all terminals support double-click events, and interval-based hacks are fragile on slow SSH networks, plus it might not match users' system click intervals which could cause an inconsistent experience or even misclicks

PR #3996, opened 2026-05-24
fix: drain delayed tmux terminal replies before events

xarthurx:

Related: [#3997](https://github.com/sxyazi/yazi/issues/3997)

## Rationale of this PR

This fixes a startup input race seen in slow tmux passthrough setups, reproduced on WSL2 + tmux 3.6b + WezTerm.

Yazi sends terminal capability queries during TUI startup, then creates the terminal `EventStream`. In this environment, some terminal replies can arrive after the DA1/DSR waits have completed. Those delayed replies contain printable payload bytes, for example:

```text
ESC P > | WezTerm 20240203-110809-5046fc22 ESC \
ESC P 1 $ r  q 0  q ESC \
ESC [ ? 12 ; 2 $ y
```

If those bytes are still pending when `EventStream` starts, they can be interpreted as user keypresses. In practice this caused Yazi to start with a random action already active, such as filter, rename, or an fzf plugin binding.

The fix adds a small tty helper to drain pending bytes until input has been quiet, and calls it under tmux immediately before constructing `EventStream`. This keeps delayed terminal replies out of the user input stream.

Verified locally:

```sh
cargo build --release --bin yazi
```

Smoke-tested the rebuilt binary in fresh tmux panes under WSL2 + WezTerm. Startup opened normal Yazi UI without the previous auto-triggered prompt.

### DIFF
--- a/yazi-tty/src/tty.rs
+++ b/yazi-tty/src/tty.rs
@@ -58,4 +58,32 @@ impl Tty {
 		let result = read();
 		(buf, result)
 	}
+
+	pub fn drain_until_quiet(&self, timeout: Duration, quiet: Duration) -> std::io::Result<usize> {
+		let mut drained = 0;
+		let until = Instant::now() + timeout;
+		let mut stdin = self.stdin.lock();
+
+		while Instant::now() < until {
+			match stdin.poll(quiet) {
+				Ok(true) => {}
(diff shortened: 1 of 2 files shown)


sxyazi:
Closing as https://github.com/sxyazi/yazi/issues/3997

Issue #3707, opened 2026-02-22
Font Previewing does not work on Windows


NSPC911:
### What system are you running Yazi on?

Windows

### What terminal are you running Yazi in?

wezterm 20260117-154428-05343b38

### `yazi --debug` output

```Shell
Yazi
    Version: 26.2.2 (d84f0b1 2026-02-22)
    Debug  : false
    Triple : x86_64-pc-windows-msvc (windows-x86_64)
    Rustc  : 1.93.1 (01f6ddf7 2026-02-11)

Ya
    Version: 26.2.2 (d84f0b1 2026-02-22)

Config
    Init             : C:\Users\notso\AppData\Roaming\yazi\config\init.lua (The system cannot find the file specified. (os error 2))
    Yazi             : C:\Users\notso\AppData\Roaming\yazi\config\yazi.toml (120 chars)
    Keymap           : C:\Users\notso\AppData\Roaming\yazi\config\keymap.toml (The system cannot find the file specified. (os error 2))
    Theme            : C:\Users\notso\AppData\Roaming\yazi\config\theme.toml (The system cannot find the file specified. (os error 2))
    VFS              : C:\Users\notso\AppData\Roaming\yazi\config\vfs.toml (The system cannot find the file specified. (os error 2))
    Package          : C:\Users\notso\AppData\Roaming\yazi\config\package.toml (The system cannot find the file specified. (os error 2))
    Dark/light flavor: "" / ""

Emulator
    TERM                : Some("xterm-256color")
    TERM_PROGRAM        : Some("WezTerm")
    TERM_PROGRAM_VERSION: Some("20260117-154428-05343b38")
    Brand.from_env      : Some(WezTerm)
    Emulator.detect     : Emulator { kind: Left(WezTerm), version: "WezTerm 20260117-154428-05343b38", light: false, csi_16t: (13, 27), force_16t: true }

Adapter
    Adapter.matches    : Iip
    Dimension.available: Dimension { rows: 37, columns: 143, width: 0, height: 0 }

Desktop
    XDG_SESSION_TYPE           : None
    WAYLAND_DISPLAY            : None
    DISPLAY                    : None
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Variables
[...]

PR #4024, opened 2026-06-07
fix: magick preview of multiframe files

Omenabaka:
### What changed

If the `magick` previewer is used without the `--flatten` argument, it now processes only the first frame of the file. For example, instead of running `magick` with `input.jpg`, it will run it with `input.jpg[0]` (which discards all frames but first and for files with only single frame it changes nothing).

### Why do you need this?

Two reasons.

**1st reason.** If you set the yazi previewer to `magick` and try to preview a file with several frames (layered image or animation), it generates preview JPEGs for each frame and names them `<hash>-0`, `<hash>-1`, etc. As a result, preview still won't be displayed (because no file named just `<hash>` was created). It seems that this logic has no use case and simply punishes a user who forgot to add `--flatten` argument to the `magick` previewer, by creating a bunch of unnecessary files and not showing preview anyway. 

**2nd reason.** Moreover sometimes for a useful preview you need the first frame instead of a flattened image.

_Example 1._  
Animated `.avif` with transparent background.  
On the left: preview generated after this PR with the previewer `magick --bg=gray`.  
On the right: preview generated with `magick --flatten --bg=gray` (you forced to use `--flatten` before this PR, otherwise a valid preview file would not be generated at all as described above).

| fish.avif[0]  --bg=gray | fish.avif --flatten --bg=gray |
| --- | --- |
| 6d13eb6fd8a9b31329844a65bccced74(avif 0  --bg=gray) | 6d13eb6fd8a9b31329844a65bccced74(avif --flatten --bg=gray) |

So before this PR you could only get a preview of all frames mingled together.  
(Furthermore, the `--bg` argument is not respected. This is likely because in `magick.lua` the background color is set after flattening, but that's another issue that I don't touch here.)

_Example 2._  
Layered `.psd` file with usage of layer masks.

| lorem.psd[0] | lorem.psd --flatten |
| --- | --- |
| 90705e64c14be5eb549defd27c09913e | 90705e64c14be5eb549defd27c09913e flatten |

ImageMagick's flatten does not respect layer masks (which should hide parts of a layer). But you don't really even need to flatten `.psd` files because they already have a preview which is stored as the first frame.

Files used for examples: [examples.zip](https://github.com/user-attachments/files/28689417/examples.zip)

## Checklist

### DIFF
--- a/yazi-plugin/preset/plugins/magick.lua
+++ b/yazi-plugin/preset/plugins/magick.lua
@@ -25,11 +25,11 @@ function M:preload(job)
 		return true
 	end
 
-	local cmd = M.with_limit()
+	local cmd = M.with_limit():arg(tostring(job.file.path))
 	if job.args.flatten then
-		cmd:arg(tostring(job.file.path)):arg("-flatten")
+		cmd:arg("-flatten")
 	else
-		cmd:arg(tostring(job.file.path) .. "[0]")
+		cmd:arg { "-delete", "1--1" }
 	end
 	cmd:arg { "-auto-orient", "-strip" }
 
--- a/yazi-plugin/preset/plugins/magick.lua
+++ b/yazi-plugin/preset/plugins/magick.lua
@@ -25,11 +25,13 @@ function M:preload(job)
 		return true
 	end
 
-	local cmd = M.with_limit():arg(tostring(job.file.path))
+	local cmd = M.with_limit()
+	cmd:arg { "-define", "filename:literal=true" }
+	local path = tostring(job.file.path)
 	if job.args.flatten then
-		cmd:arg("-flatten")
+		cmd:arg { path, "-flatten" }
 	else
-		cmd:arg { "-delete", "1--1" }
+		cmd:arg { "-define", "image:frames=0", path }
 	end
 	cmd:arg { "-auto-orient", "-strip" }
 
--- a/yazi-plugin/preset/plugins/magick.lua
+++ b/yazi-plugin/preset/plugins/magick.lua
@@ -25,7 +25,7 @@ function M:preload(job)
 		return true
 	end
 
-	local cmd = M.with_limit():arg({ "-define", "filename:literal=true" })
+	local cmd = M.with_limit():arg { "-define", "filename:literal=true" }
 	if job.args.flatten then
 		cmd:arg { tostring(job.file.path), "-flatten" }
 	else
--- a/yazi-plugin/preset/plugins/magick.lua
+++ b/yazi-plugin/preset/plugins/magick.lua
@@ -25,13 +25,11 @@ function M:preload(job)
 		return true
 	end
 
-	local cmd = M.with_limit()
-	cmd:arg { "-define", "filename:literal=true" }
-	local path = tostring(job.file.path)
+	local cmd = M.with_limit():arg({ "-define", "filename:literal=true" })
 	if job.args.flatten then
-		cmd:arg { path, "-flatten" }
+		cmd:arg { tostring(job.file.path), "-flatten" }
 	else
-		cmd:arg { "-define", "image:frames=0", path }
+		cmd:arg { "-define", "image:frames=0", tostring(job.file.path) }
 	end
 	cmd:arg { "-auto-orient", "-strip" }
 
--- a/yazi-plugin/preset/plugins/magick.lua
+++ b/yazi-plugin/preset/plugins/magick.lua
@@ -25,9 +25,11 @@ function M:preload(job)
 		return true
 	end
 
-	local cmd = M.with_limit():arg(tostring(job.file.path))
+	local cmd = M.with_limit()
 	if job.args.flatten then
-		cmd:arg("-flatten")
+		cmd:arg(tostring(job.file.path)):arg("-flatten")
+	else
+		cmd:arg(tostring(job.file.path) .. "[0]")
 	end
 	cmd:arg { "-auto-orient", "-strip" }
 


yazi-plugin/preset/plugins/magick.lua
```diff
@@ -25,9 +25,11 @@ function M:preload(job)
 		return true
 	end
 
-	local cmd = M.with_limit():arg(tostring(job.file.path))
+	local cmd = M.with_limit()
 	if job.args.flatten then
-		cmd:arg("-flatten")
+		cmd:arg(tostring(job.file.path)):arg("-flatten")
+	else
+		cmd:arg(tostring(job.file.path) .. "[0]")
```
sxyazi: Thanks for the patch! 

Appending `[0]` after the file path causes an error when image filenames already contain `[0]`, like `a.jpg[0]` or `b.jpg[0][0]`:

```sh
`magick` exited with error code: 1
```

without manually appending `[0]`, they can be previewed normally with the `magick` previewer. 

Perhaps some kind of escaping is needed for the file path here?

sxyazi (changes requested):

yazi-plugin/preset/plugins/magick.lua
Omenabaka: This turned out to be more difficult than I thought. There a combination of two problems.

1. Apparently, `magick` does not provide a way to escape individual characters in a file path. Moreover, you cannot even pass an input path to `magick` in such a way that it does not internally process the characters in it at all. For the output path, there is a setting that allows this (`-define filename:literal=true`), but looks like its implementation for the input path was not added: https://github.com/ImageMagick/ImageMagick/issues/8035#issuecomment-2727571243
2. You can only select a frame by adding square brackets to the end of the path. There is no alternative argument for this operation.

So, as I understand it, `magick` works with filenames as follows:

- If a file exists with exactly the name that was passed as an argument, it uses that file. It doesn't matter whether it contains special characters or not, everything is interpreted as part of the path. Therefore, files like `a.jpg[0]` are now handled correctly.
- If a file with exactly that name does not exist, `magick` tries to parse the options passed in `[]`, but it also parse the `[]` that were part of the filename, which leads to an error.

In other words, you can avoid using `magick`'s functions that activated by special characters in the filename, and everything will be fine with any filenames. But if you use at least one of these functions (and we need Frame Selection, which is only available by adding `[]` to the filename), then ALL special characters in the filename start being treated as special. There is no option in between, like escaping some characters and having special use for others.

At first, I thought about checking path for a trailing `]` (that was the only source of issues with square brackets). But then I found that `magick` also have special meaning for `:`, `@`, and `%`. And having these in a filename while trying to use Frame Selection (`[0]`) trips up `magick` too, because it now tries to interpret them. It becomes too complicated to deal with, so maybe it's better not to use Frame Selection at all (maybe sometime in the future if `magick` adds a proper argument for it).

### Alternative way to get the first frame

`-delete 1--1` argument discards all frames from the second (`1`) to (`-`) the last (`-1`).

The only concern is performance. I'm not good at benchmarking (perhaps someone else should run more tests) but here are my observations:

**For a multiframe file:**
- `[0]` is the fastest. As I understand it, it does not depend on the number of frames, because it doesn't even load them for processing, only the first one. But we cannot use this.
- `-delete 1--1` is significantly slower, because it loads all frames and then deletes the extras.
- *no arguments* is even slower than the previous option (but not for much). That's how it works now, creates one file per frame — not what we need.
- `-flatten` takes about the same time as the previous option.

**For a single-frame file:**
- *no arguments* is how it works now, nothing is done with frames.
- `[0]` takes the same time as above.
- `-delete 1--1` takes the same time as above.
- `-flatten` is slightly slower.

So if we use `-delete 1--1`, performance for single-frame files should not change. For multi-frame files, performance will not be the maximum that magick could achieve, but still it will be the same/faster than now (and working previews will be generated instead of a bunch of numbered frames).

sxyazi: Thanks for digging into this so thoroughly, the path rule of ImageMagick is really a pain, which reminds me of https://github.com/sxyazi/yazi/issues/3707 :(

Would it be possible to check whether the file path contains `[]`, and if so, fall back to `-delete`; otherwise, still use `[0]` for better performance for multi-frame files?

Omenabaka: I've looked more into it, and it seems to be ImageMagick's bug rather than intended behavior. Because `:` or `@` in filenames (for which IM has a special meaning too) doesn't cause the same problems (as I assumed in previous reply). The problem only occurs with filenames containing a trailing `]`.

It looks like it will be fixed soon: https://github.com/ImageMagick/ImageMagick/issues/8813

I have in mind some other edge cases that could still hinder the use of the `[0]` syntax. But first I want to see how this will work after IM's fix for the trailing `]` is implemented, and then I'll come back here.

Omenabaka:
ImageMagick 7.1.2-27 released

## Trailing `]`

Bug with trailing `]` in filenames has been fixed.

## Filenames ambiguity

Another potential issue that I mentioned: https://github.com/ImageMagick/ImageMagick/issues/8823
In short: if you have files named `test` and `test[0]` in the same directory and try to get the 1st frame of `test` by adding `[0]` to the filename you instead pass the file named `test[0]` to IM.

IM 7.1.2-27 changes this and now all special characters are always interpreted first. In this version, using`test[0]` as filename always tries to get the 1st frame of `test` even if the file `test[0]` exist.

This leads to new problem with Yazi previewer that already affects anyone who updated IM (it's not tied to this PR). Previewing files with names like `test[0]` will result in error (or a wrong preview if the file `test` exists). To avoid interpretation of special characters in filenames, you should use `-define filename:literal=true` (which starting from this IM version works for input names too).

## Frame Selection without `[0]`

But after this IM patch new edge-case bugs were found: with filenames containing `%d` and `[0]` syntax at the same time https://github.com/ImageMagick/ImageMagick/issues/8823#issuecomment-5004898647

So instead of further dealing with the filename parser, IM added an alternative way to select frames: `-define image:frames=`

It should be released in the next version (7.1.2-28), but I tested the latest build and it works just fine.

## Old IM versions

Even if you add a `-define` setting that not exists in your version of IM, it is just ignored. So after this PR for users with older IM versions Yazi behaviour shouldn't change and after they updates to IM 7.1.2-28 they get fully fixed previews.

sxyazi (approved):
LGTM!

Thanks for following up on this issue. Glad ImageMagick can provide a new `-define` value so we don't need to add hacks and can keep the code as clean as it is!

PR #4035, opened 2026-06-10
feat: OSC 5522

UnnaturalTwilight:
## Rationale of this PR
Adds support for using OSC 5522 (Kitty Clipboard Extension) to read and write all mime types from the system clipboard.

The main protocol can be found here: https://sw.kovidgoyal.net/kitty/clipboard/#copying-all-data-types-to-the-clipboard
The specification for paste events can be found here: https://rockorager.dev/misc/bracketed-paste-mime/
In cases where Kitty's implementation differs from the above specifications I have followed Kitty's implementation.

This pr implements support for copying and pasting `text/uri-list` as well as `text/plain`. The paste events have been exposed to lua similarly to drag and drop allowing supporting more formats easily. I have left the existing clipboard functions as is so `ya.clipboard()` and similar continue to use external clipboard tools. In the case of the copy keybindings support for OSC 5522 is detected and OSC 5522 is used if available, otherwise the original clipboard commands are used.

For now I have left the same 1MB size limit as the drag and drop however this will need to be revisited to support directly pasting files such as images from the clipboard.

I have tested this with Kitty 0.47.1

## Checklist


UnnaturalTwilight:
Due to the significant discrepancy between Kitty's actual behavior and what is described at https://rockorager.dev/misc/bracketed-paste-mime/ I wrote up a Gist describing how the spec is actually implemented in Kitty.
https://gist.github.com/UnnaturalTwilight/2617336353fb5f8056413e3c759aa6d3
I believe Kitty's behavior follows the description and intent of the original spec, however the examples provided on that page do not match Kitty's behavior or my implementation. As Kitty is the only terminal I could find that currently implements the paste events mode I think it makes sense to treat its implementation as correct.

yazi-plugin/preset/components/root.lua
```diff
@@ -99,3 +99,31 @@ function Root:drop(event)
...
+		require("clipboard").copy_uri_list(list)
+	end
+	-- TODO !!5522!! Suport non text formats
+	-- if event.data["image/png"] ~= nil then
+	-- 	local type = "image/png"
+	-- 	local data = event.data["image/png"]
+	-- 	ya.dbg("Pasting image/png:")
+	-- 	require("clipboard").paste_image(type, data)
+	-- end
+end
```
sxyazi: These two should be merged into one to reduce the API surface. See how the DnD did that.

yazi-plugin/src/runtime/term.rs
```diff
@@ -7,6 +7,7 @@ pub(super) fn term() -> Composer<ComposerGet, ComposerSet> {
 		match key {
 			b"light" => EMULATOR.light.into_lua(lua),
 			b"cell_size" => cell_size(lua)?.into_lua(lua),
+			b"osc5522_clipboard" => EMULATOR.osc_5522.into_lua(lua),
```
sxyazi: Why expose it since it doesn't seem to be used anywhere?

yazi-core/src/tasks/file.rs
```diff
@@ -23,7 +23,7 @@ impl Tasks {
 		}
 	}
 
-	pub fn file_copy(&self, src: &Yanked, dest: &UrlBuf, force: bool, follow: bool) {
+	pub fn file_copy(&self, src: &Yanked, dest: &UrlBuf, force: bool, follow: Option<bool>) {
```
sxyazi: Just remove the `follow` parameter and handle it `FileInCopy::new()` internally, see how the `file_cut()` did that.

yazi-cli/src/env/env.rs
```diff
@@ -112,6 +112,7 @@ impl Env {
 		writeln!(s, "    wl-copy/paste: {} / {}", Self::dep_version("wl-copy", "--version"), Self::dep_version("wl-paste", "--version"))?;
 		writeln!(s, "    xclip        : {}", Self::dep_version("xclip", "-version"))?;
 		writeln!(s, "    xsel         : {}", Self::dep_version("xsel", "--version"))?;
+		writeln!(s, "    OSC 5522     : {:?}", yazi_emulator::EMULATOR.osc_5522)?;
```
sxyazi: I believe https://github.com/sxyazi/yazi/blob/07a5deb1095321917e172770fc1630a57dedce52/yazi-cli/src/env/env.rs#L34 already includes this info, so no need here

yazi-actor/Cargo.toml
```diff
@@ -52,6 +52,7 @@ scopeguard   = { workspace = true }
 tokio        = { workspace = true }
 tokio-stream = { workspace = true }
 tracing      = { workspace = true }
+percent-encoding = { workspace = true }
```
sxyazi: Please align the `=` with others

sxyazi (changes requested):

yazi-plugin/src/runtime/term.rs
UnnaturalTwilight: The idea was for plugins to be able to see if calling `rt.tty:queue("WriteClipboard", { })` would work for setting the clipboard.

sxyazi: That should be hidden in the `ya.clipboard()` as an internal detail, and we probably need to find a way to harmonize them together, a little design is needed here.

UnnaturalTwilight: I think OSC5522 support needs to be exposed to plugins somehow as features of OSC5522 like aliases and separate payloads per mime types aren't available with other ways of accessing the clipboard.
I think everything but OSC52 support setting the MIME type when copying so I suppose `ya.clipboard()` could do a "close enough" fallback.

I'm good with removing the field for now and leave it for a future implementation to solve.

sxyazi: I'm not able to access X11 right now but on Wayland `wl-copy` seems to support specifying a MIME type with its `--type` option. But in general I agree leaving it for the future to resolve and keep the PR as simple if there's still something that hasn't been figured out.

yazi-plugin/preset/components/root.lua
```diff
@@ -99,3 +99,19 @@ function Root:drop(event)
...
+		rt.tty:queue("ReadClipboard", { mimes = mimetypes, pw = event.pw, name = "Paste Event", primary = event.primary })
+		rt.tty:flush()
+	elseif event.data["text/uri-list"] ~= nil then
+		local list = event.data["text/uri-list"]
+		ya.dbg("Pasting URI list:", list)
+		require("clipboard").copy_uri_list(list)
+	end
+end
+
+function Root:write_result(event) end
```
sxyazi: This should be removed and leave only one API for clipboard events

yazi-actor/src/mgr/copy.rs
```diff
@@ -53,7 +64,32 @@ impl Actor for Copy {
...
+				"uri_list" => {
+					data.push(ClipboardData {
+						mime:    b"text/uri-list".to_vec(),
+						payload: s.clone(),
+						alias:   b"text/plain".to_vec(),
+					});
+					#[cfg(target_os = "linux")]
+					// Because Thunar (and likely others) won't reconize `text/uri-list`
+					data.push(ClipboardData {
+						mime:    b"x-special/gnome-copied-files".to_vec(),
```
sxyazi: I think this should be implemented as a plugin for better flexibility so users can customize it since there might be a couple of filemanager-specific custom MIME types and we won't want to build them all in.

yazi-actor/src/app/clipboard.rs
```diff
@@ -0,0 +1,45 @@
...
+
+		let result = Lives::scope(cx.core, move || {
+			runtime_scope!(LUA, "root", {
+				let root = LUA.globals().raw_get::<Table>("Root")?.call_method::<Table>("new", area)?;
+
+				if event.is_read() {
+					root.call_method::<()>("paste_data", event)?;
+				} else {
+					root.call_method::<()>("write_result", event)?;
+				}
```
sxyazi: Ditto. These two should be merged into one.

sxyazi (changes requested):

yazi-emulator/src/emulator.rs
```diff
@@ -65,12 +68,17 @@ impl Emulator {
 		};
 
 		let csi_16t = Self::csi_16t(&resp).unwrap_or_default();
+
+		let osc_5522 =
+			["\x1b[?5522;1$y", "\x1b[?5522;2$y", "\x1b[?5522;3$y"].iter().any(|s| resp.contains(s));
```
sxyazi: Please pull it off into a new method called `osc_5522` like the existing `csi_16t`

yazi-binding/src/clipboard.rs
```diff
@@ -0,0 +1,48 @@
...
+				Ok(Value::Nil)
+			}
+		});
+
+		fields.add_cached_field_mut("data", |lua, me| match &mut me.inner {
+			Inner::ReadData(ClipboardRead { data, .. }) => lua
+				.create_table_from(
+					data
+						.iter()
+						.map(|d| (lua.create_string(&d.mime).ok(), lua.create_external_string(&*d.data).ok())),
```
sxyazi: Creating string failures should be a fatal error and shouldn't be silently ignored

sxyazi (changes requested):

yazi-actor/src/app/mod.rs
```diff
@@ -16,4 +16,5 @@ yazi_macro::mod_flat!(
 	theme
 	title
 	update_progress
+	clipboard
```
sxyazi: Need a reorder to make sure it's in alphabetical order with the other actions

sxyazi (changes requested):

yazi-parser/src/spark/spark.rs
```diff
@@ -25,6 +25,7 @@ pub enum Spark<'a> {
 	AppTheme(crate::VoidForm),
 	AppTitle(crate::app::TitleForm),
 	AppUpdateProgress(crate::app::UpdateProgressForm),
+	AppClipboard(crate::app::ClipboardForm),
```
sxyazi: Ditto. New variants should be added in alphabetical order

yazi-binding/src/lib.rs
```diff
@@ -2,4 +2,4 @@ mod macros;
 
 yazi_macro::mod_pub!(config elements event keymap process theme);
 
-yazi_macro::mod_flat!(access calculator cha chan composer dnd error fd file handle icon id image input iter layer mouse path permit range runtime scheme selector sendable stage style tty url utils);
+yazi_macro::mod_flat!(access calculator cha chan composer dnd error fd file handle icon id image input iter layer mouse path permit range runtime scheme selector sendable stage style tty url utils clipboard);
```
sxyazi: Ditto. New module should be added in alphabetical order

yazi-term/src/sequence/clipboard.rs
```diff
@@ -18,3 +19,100 @@ impl Display for SetClipboard {
...
+	pub mime:    &'a [u8],
+	pub pw:      &'a [u8],
+	pub name:    &'a [u8],
+	pub primary: bool,
+}
+
+impl Display for ReadClipboard<'_> {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		let b64_mime = BASE64_PAD.encode(self.mime).into_bytes();
+		let mime_str = unsafe { String::from_utf8_unchecked(b64_mime) };
```
sxyazi: `BASE64_PAD.encode` returns a `String`, you don't need to turn it into `Vec<u8>` and then turn it back with an `unsafe`

yazi-term/src/sequence/clipboard.rs
```diff
@@ -18,3 +19,100 @@ impl Display for SetClipboard {
...
+		let mut metadata = String::new();
+		if self.pw.len() > 0 {
+			let b64_pw = BASE64_PAD.encode(self.pw).into_bytes();
+			let pw_str = unsafe { String::from_utf8_unchecked(b64_pw) };
```
sxyazi: Ditto.

yazi-term/src/sequence/clipboard.rs
```diff
@@ -18,3 +19,100 @@ impl Display for SetClipboard {
...
+			let b64_name = BASE64_PAD.encode(self.name).into_bytes();
+			let name_str = unsafe { String::from_utf8_unchecked(b64_name) };
```
sxyazi: Ditto.

sxyazi (changes requested):

yazi-term/src/event/clipboard.rs
```diff
@@ -0,0 +1,156 @@
...
+#[derive(Clone, Copy, Debug, Eq, FromRepr, IntoStaticStr, PartialEq)]
+#[repr(u8)]
+pub enum ClipboardType {
+	Read  = 1,
+	Write = 2,
+}
+
+// --- MIME list
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct ClipboardMimeList(String);
```
sxyazi: It'd be better to rename the existing `DndMimeList` to `MimeList` and reuse it instead.

yazi-term/src/event/clipboard.rs
```diff
@@ -0,0 +1,156 @@
...
+}
+
+impl ClipboardEvent {
+	pub fn r#type(&self) -> &'static str {
+		match self {
+			Self::ReadMimetypes(_) => "read_mimetypes",
+			Self::ReadData(_) => "read_data",
+			Self::ReadError(_) => "read_error",
+			Self::WriteSuccess => "write_success",
+			Self::WriteError(_) => "write_error",
```
sxyazi: It's better to have them in one word like the DnD's

yazi-term/src/sequence/clipboard.rs
```diff
@@ -18,3 +19,100 @@ impl Display for SetClipboard {
...
+pub struct WriteClipboard<'a> {
+	pub data: Vec<WriteClipboardData<'a>>,
+}
+
+impl Display for WriteClipboard<'_> {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		write!(f, "\x1b]5522;type=write\x1b\\")?;
+		for item in &self.data {
+			let b64_mime = BASE64_PAD.encode(item.mime).into_bytes();
+			let mime_str = unsafe { String::from_utf8_unchecked(b64_mime) };
```
sxyazi: Ditto.

yazi-term/src/sequence/clipboard.rs
```diff
@@ -18,3 +19,100 @@ impl Display for SetClipboard {
...
+			let data = item.payload;
+
+			for (_, chunk) in data.chunks(4096).enumerate() {
+				let b64_chunk = BASE64_PAD.encode(chunk).into_bytes();
+				let s = unsafe { String::from_utf8_unchecked(b64_chunk) };
```
sxyazi: Ditto.

yazi-term/src/sequence/clipboard.rs
```diff
@@ -18,3 +19,100 @@ impl Display for SetClipboard {
...
+				write!(f, "\x1b]5522;type=wdata:mime={};{s}\x1b\\", mime_str)?;
+			}
+
+			if item.alias.len() > 0 {
+				let b64_alias = BASE64_PAD.encode(item.alias).into_bytes();
+				let s = unsafe { String::from_utf8_unchecked(b64_alias) };
```
sxyazi: Ditto.

sxyazi (changes requested):

yazi-actor/src/mgr/copy.rs
```diff
@@ -51,6 +52,19 @@ impl Actor for Copy {
 				"name_without_ext" => {
 					s.extend_from_slice(&form.separator.transform(&f.stem().unwrap_or_default()));
 				}
+				"uri_list" => {
```
sxyazi: Can you tell me what the behavior of `copy uri_list` is on Linux? 

I tested it on macOS and it seems like it just copies the `file://` paths as plain text to the clipboard - is that the expected behavior?

UnnaturalTwilight: It should copy the `file://` paths as both `text/uri-list` and `text/plain`, any app that knows how to handle `text/uri-list` should paste the files where as a app that only handles plain text will paste the `file://` links insted. If macos has a different way of copying files to the clipboard it may be best to break this off into a plugin.

sxyazi: I dug into it a bit, and this probably needs the terminal to map `text/uri-list` to [`NSFilenamesPboardType`](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/DragandDrop/Tasks/DraggingFiles.html) on macOS, kinda like what kitty does in its DnD protocol.

For now I'll just leave it out for plugins until we find a cross-platform way to support it.

sxyazi (approved):
Thanks for the groundwork you did on this! 

It's a great starting point, looking forward to finding places where OSC 5522 can be put to even better use!

Issue #404, opened 2023-11-27
Add undo feature to file operations


evilakela:
### Will you be willing to contribute this feature?

- [ ] Yes, I'll give it a shot

### Describe the solution you'd like

Ability to undo following actions by pressing u (vim like)
- Paste the yanked files (p)
- Move the files to the trash (d)
- Create a file or directory (a)
- Rename a file or directory (r)
- Create a symbolic link to the yanked files (- and _)

I think the feature does not need additional explanation and I am surprised that there is no such option. It might be difficult to implement, idk, but that would be nice.

PR #4039, opened 2026-06-11
feat: allow cycle rename

### LINKED ISSUE #4034: Allow cycle rename during bulk renaming
### `ya env` output

```Shell
Yazi
    Version  : 26.5.6 (Homebrew 2026-05-05)
    Debug    : false
    Triple   : aarch64-apple-darwin (macos-aarch64)
    Rustc    : 1.95.0 (59807616 2026-04-14)
    Backtrace: None

Ya
    Version: 26.5.6 (Homebrew 2026-05-05)

Emulator
    TERM                : Some("xterm-kitty")
    TERM_PROGRAM        : None
    TERM_PROGRAM_VERSION: None
    Brand.from_env      : Some(Kitty)
    Emulator.detect     : Emulator { kind: Left(Kitty), version: "kitty(0.46.2)", light: false, csi_16t: (15, 31), force_16t: false }

Adapter
    Adapter.matches    : Kgp
    Dimension.available: Dimension { rows: 50, columns: 193, width: 2895, height: 1550 }

Desktop

SSH

WSL
    WSL: false

Variables
    VISUAL             : Some("nvim")
    YAZI_FILE_ONE      : None
    YAZI_CONFIG_HOME   : None
    YAZI_ZOXIDE_OPTS   : None
    SSH_AUTH_SOCK      : Some("/private/tmp/com.apple.launchd.MyJnc2iptR/Listeners")
    FZF_DEFAULT_OPTS   : Some("--bind=tab:up,shift-tab:down --style=default --preview=\"[[ -f {} ]] && bat --color=always --style=numbers --line-range=:500 {} || [[ -d {} ]] && eza -1 --color=always {} || [[ {} == http* ]] && curl -sL --max-time 3 {} | w3m -dump -T text/html | head -80 || echo {}\"")
    FZF_DEFAULT_COMMAND: Some("fd")

Text Opener
    default     : Some(OpenerRule { id: Id(24), run: NonEmptyString("${EDITOR:-nvim} %s"), block: true, orphan: false, desc: "$EDITOR", for: Unix, spread: true })
    block-create: Some(OpenerRule { id: Id(24), run: NonEmptyString("${EDITOR:-nvim} %s"), block: true, orphan: false, desc: "$EDITOR", for: Unix, spread: true })
    block-rename: Some(OpenerRule { id: Id(24), run: NonEmptyString("${EDITOR:-nvim} %s"), block: true, orphan: false, desc: "$EDITOR", for: Unix, spread: true })

Multiplexers
    TMUX               : false
    tmux version       : tmux 3.6b
    tmux build flags   : enable-sixel=Supported
    ZELLIJ_SESSION_NAME: None

Dependencies
    file          : 5.41
    ffmpeg/ffprobe: 8.1.1 / 8.1.1
    pdftoppm      : No such file or directory (os error 2)
    magick        : 7.1.2-24
    fzf           : 0.73.1
    fd/fdfind     : 10.4.2 / No such file or directory (os error 2)
    rg            : 15.1.0
    zoxide        : 0.9.9
    7zz/7z        : 26.01 / No such file or directory (os error 2)
    jq            : 1.7.1

Clipboard

Routine


```

Currently, bulk rename fails when the rename operations contain a cycle. For example, I want to rename:

- `file1` → `file2`
- `file2` → `file3`
- `file3` → `file1`

The goal is to support cyclic renames during bulk rename.

My idea is to modify this function so that it returns both normal rename chains and cycle rename groups:

https://github.com/sxyazi/yazi/blob/07a5deb1095321917e172770fc1630a57dedce52/yazi-actor/src/mgr/bulk_rename.rs#L200-L210

Normal chains can be processed as they are. For cycle groups, add a temporary rename to break the cycle. For example, the rename order will be:

- `file1` → `.tmp`
- `file3` → `file1`
- `file2` → `file3`
- `.tmp` → `file2`

This function can be used to produce the cycle groups, and it reuses the `todos` and `orders` in `prioritized_paths`.
```rust
fn partition_cycles(
	mut todos: HashMap<&Tuple, Tuple>,
	orders: &HashMap<&Tuple, usize>,
) -> Vec<Renames> {
	let mut starts = todos.keys().copied().collect();
	starts.sort_unstable_by(|a, b| orders[a].cmp(&orders[b]));

	let mut cycles = Vec::new();
	for start in starts {
		if !todos.contains_key(start) {
			continue;
		}

		let mut cycle = Vec::new();
		let mut cur = start;
		while let Some(next) = todos.remove(cur) {
			cycle.push((cur.clone(), next.clone()));
			let Some((&owner, _)) = todos.get_key_value(&next) else { break };
			cur = owner;
		}
		cycles.push(cycle);
	}
	cycles
}
```

### Additional context

I saw this issue but it was closed #3718.

### Checklist

likelikeslike:

Resolves #4034 

## Rationale of this PR

This PR is a follow of issue #4034, which introduces the cycle rename groups to solve the cycle rename during bulk renaming. By adding a temp filename for each cycle group, it breaks and can be renamed as normal chain renaming.

demo

## Checklist

### DIFF
--- a/yazi-actor/src/mgr/bulk_rename.rs
+++ b/yazi-actor/src/mgr/bulk_rename.rs
@@ -386,7 +386,7 @@ impl Eq for Tuple {}
 
 impl Hash for Tuple {
 	fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
-		self.1.hash(state);
+		self.1.as_strand().encoded_bytes().hash(state);
 	}
 }
 
--- a/yazi-actor/src/mgr/bulk_rename.rs
+++ b/yazi-actor/src/mgr/bulk_rename.rs
@@ -409,44 +409,70 @@ mod tests {
 
 	#[test]
 	fn test_sort() {
-		fn cmp(input: &[(&str, &str)], expected: &[(&str, &str)]) {
-			let sorted = BulkRename::prioritized_paths(
+		fn cmp(input: &[(&str, &str)], chain: &[(&str, &str)], cycles: &[&[(&str, &str)]]) {
+			let (res_chain, res_cycles) = BulkRename::prioritized_paths(
 				input.iter().map(|&(o, _)| Tuple::new(0, o)).collect(),
 				input.iter().map(|&(_, n)| Tuple::new(0, n)).collect(),
 			);
-			let sorted: Vec<_> =
-				sorted.iter().map(|(o, n)| (o.to_str().unwrap(), n.to_str().unwrap())).collect();
-			assert_eq!(sorted, expected);
+			let flat = |v: &[(Tuple, Tuple)]| -> Vec<(String, String)> {
+				v.iter().map(|(o, n)| (o.to_str().unwrap().into(), n.to_str().unwrap().into())).collect()
+			};
+			let want = |v: &[(&str, &str)]| -> Vec<(String, String)> {
+				v.iter().map(|&(o, n)| (o.into(), n.into())).collect()
+			};
+			assert_eq!(flat(&res_chain), want(chain));
+			assert_eq!(
+				res_cycles.iter().map(|c| flat(c)).collect::<Vec<_>>(),
+				cycles.iter().map(|c| want(c)).collect::<Vec<_>>()
+			);
 		}
 
 		#[rustfmt::skip]
 		cmp(
 			&[("2", "3"), ("1", "2"), ("3", "4")],
-			&[("3", "4"), ("2", "3"), ("1", "2")]
+			&[("3", "4"), ("2", "3"), ("1", "2")],
+			&[]
 		);
 
 		#[rustfmt::skip]
 		cmp(
 			&[("1", "3"), ("2", "3"), ("3", "4")],
-			&[("3", "4"), ("1", "3"), ("2", "3")]
+			&[("3", "4"), ("1", "3"), ("2", "3")],
+			&[]
+		);
+		#[rustfmt::skip]
+		cmp(
+			&[("b", "b_"), ("a", "a_"), ("c", "c_")],
+			&[("b", "b_"), ("a", "a_"), ("c", "c_")],
+			&[]
 		);
 
 		#[rustfmt::skip]
 		cmp(
-			&[("2", "1"), ("1", "2")],
-			&[("2", "1"), ("1", "2")]
+			&[("1", "2"), ("2", "1")],
+			&[],
+			&[&[("1", "2"), ("2", "1")]]
+		);
+
+		#[rustfmt::skip]
+		cmp(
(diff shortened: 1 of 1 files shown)


likelikeslike:
Hi @sxyazi, would you mind reviewing this PR when you have some time? I'd really appreciate it. Thanks.

dw-xdy:
@likelikeslike great feature! I ran into this issue myself some time ago — I wanted to swap the names of two images, and it would have been great if that were supported.

However, I do see a problem here: Yazi currently doesn't have an `undo` feature (except within the `input` context). Although Yazi has a final confirmation step in place, I think the vast majority of people won't bother to double-check it. So my concern is that — without an undo mechanism — performing bulk cyclic rename operations on a large number of files is a fairly risky move.

Unfortunately, in #404, Yazi's maintainer sxyazi has already made it clear that undo support will not be implemented.

I'd like to hear your thoughts on this concern.

likelikeslike:
> @likelikeslike great feature! I ran into this issue myself some time ago — I wanted to swap the names of two images, and it would have been great if that were supported.
> 
> However, I do see a problem here: Yazi currently doesn't have an `undo` feature (except within the `input` context). Although Yazi has a final confirmation step in place, I think the vast majority of people won't bother to double-check it. So my concern is that — without an undo mechanism — performing bulk cyclic rename operations on a large number of files is a fairly risky move.
> 
> Unfortunately, in #404, Yazi's maintainer sxyazi has already made it clear that undo support will not be implemented.
> 
> I'd like to hear your thoughts on this concern.

The only change here is that, for each cyclic rename group, a temporary filename is introduced. It then serves as the normal bulk rename operation. So the concern you mentioned can also occur in a normal bulk rename, not just during cyclic renames.

dw-xdy:
@likelikeslike Thanks for the quick reply. :)

Yes, you're absolutely right. But I think it's worth distinguishing between ordinary batch renaming and cyclic renaming — because the risks are quite different.

With ordinary batch renaming, the changes are typically applied within each line independently — e.g., adding a prefix, replacing text, or changing file extensions. There's no dependency between lines, so users generally have a clear idea of what they're doing, and the chance of errors is relatively low.

Cyclic renaming, on the other hand, involves swapping names between lines, and the filenames often follow a certain pattern or sequence. This interdependency makes mistakes much more likely — and if something goes wrong, cleaning up the mess afterward can be quite painful.

That's really the core of my concern.

likelikeslike:
@dw-xdy I do agree. But since yazi has no global undo mechanism, building a localized undo specifically for cyclic renames would feel out of place.

So I'd rather not add an undo onto this PR. The cyclic-rename support itself only makes up for a correctness gap (a valid intent like a → b, b → a currently fails), and the safety/undo question is really a broader, app-wide decision that's best left to @sxyazi rather than solved here.

PR #4040, opened 2026-06-13
feat: new `ui.Input` element

sxyazi:
This PR adds a new `ui.Input` element that lets you use and interact with input components in the layout system:

```lua
-- ~/.config/yazi/init.lua
local M = {
  _id = "my-input",
  _input = ui.Input { value = "default value" },
}

function M:new(area)
  self:layout(area)
  return self
end

function M:layout(area)
  self._outer = ui.Rect {
    x = area.x,
    y = area.y,
    w = math.min(30, area.w),
    h = math.min(3, area.h),
  }
  self._inner = self._outer:pad(ui.Pad.xy(1))
end

function M:reflow() return { self } end

function M:redraw()
  return {
    ui.Clear(self._outer),
    ui.Border(ui.Edge.ALL):area(self._outer):style(ui.Style():fg("blue")):title("Input:"),
    self._input:area(self._inner):focus(true),
  }
end

local id = Modal:children_add(M, 10)
ps.sub("key-input-close", function(args)
  Modal:children_remove(id)
  return args
end)
```

https://github.com/user-attachments/assets/43b79c5a-7bd0-49f8-8b80-1cba5cd93395

There are still some APIs missing that will be filled in through follow-up PRs.


neelspoc32:
Does focusing on the input field freeze the main Yazi UI, or does it allow for simultaneous interaction with both the input element and the rest of the interface?

sxyazi:
When you focus a `ui.Input`, it takes over the user input, you can subscribe to DDS events that fire when you submit or cancel the input, or even just while typing in it. If you need to, you can revoke the focus on the `ui.Input` from inside the event handler to stop it from taking over user input.

neelspoc32:
Thank you for responding. 
Are you perhaps talking about this:


```
function M:redraw()
  return {
    ...
    ....
    self._input:area(self._inner):focus(true),  <--- setting this to false?
  }
```

sxyazi:
Yup! Passing a false or not returning (i.e. not rendering) at all both ways unfocus it.

neelspoc32:
Thanks!! I will try it!

PR #4041, opened 2026-06-13
adapter: allow opt-in Kitty graphics passthrough under Zellij

luccahuguet:
## What

Yazi currently forces the Sixel adapter whenever `ZELLIJ_SESSION_NAME` is set.

This adds an explicit opt-in environment variable:

```sh
YAZI_ZELLIJ_KITTY_PASSTHROUGH=1
```

When set, Yazi keeps Kitty graphics adapters available under Zellij while preserving the existing safe default for normal Zellij sessions.

## Why

The default Sixel fallback is still the right behavior for upstream Zellij today.

However, downstream integrations can provide a Zellij build that passes Kitty graphics through to the outer terminal. Yazelix currently carries a small Zellij fork for that path:

https://github.com/luccahuguet/yazelix-zellij

That fork lets Yazi image previews work through Zellij in Kitty-graphics-capable terminals. Without this opt-in, Yazi still forces Sixel just because `ZELLIJ_SESSION_NAME` is present, even when the surrounding Zellij/terminal stack can handle Kitty graphics correctly.

This should also make the eventual upstream Zellij transition easier for Yazi. When Zellij itself starts supporting Kitty graphics passthrough, Yazi maintainers can switch from “always force Sixel under Zellij” to a capability-aware path incrementally, with this narrow opt-in already isolating the adapter-filtering behavior that needs to change.

## Details

- `ZELLIJ_SESSION_NAME` still triggers Zellij-specific adapter filtering
- without `YAZI_ZELLIJ_KITTY_PASSTHROUGH`, only Sixel remains allowed
- with `YAZI_ZELLIJ_KITTY_PASSTHROUGH`, Kitty graphics adapters are also allowed
- non-Kitty protocols such as IIP remain filtered under Zellij
- `ya env` reports the new variable for diagnostics

## Checks

- `cargo test -p yazi-adapter zellij -- --test-threads=1`
- `cargo check -p yazi-adapter -p yazi-cli`

### DIFF
--- a/yazi-adapter/src/adapter.rs
+++ b/yazi-adapter/src/adapter.rs
@@ -9,6 +9,9 @@ use yazi_shared::env_exists;
 
 use crate::{Adapters, SHOWN, drivers};
 
+const ZELLIJ_SESSION_NAME_ENV: &str = "ZELLIJ_SESSION_NAME";
+pub const ZELLIJ_KITTY_PASSTHROUGH_ENV: &str = "YAZI_ZELLIJ_KITTY_PASSTHROUGH";
+
 #[derive(Clone, Copy, Debug, Display, Eq, IntoStaticStr, PartialEq)]
 #[strum(serialize_all = "kebab-case")]
 pub enum Adapter {
@@ -75,8 +78,8 @@ impl Adapter {
 impl Adapter {
 	pub fn matches(emulator: &Emulator) -> Self {
 		let mut adapters: Adapters = emulator.into();
-		if env_exists("ZELLIJ_SESSION_NAME") {
-			adapters.retain(|p| *p == Self::Sixel);
+		if env_exists(ZELLIJ_SESSION_NAME_ENV) {
+			adapters.retain(|p| p.supported_in_zellij());
 		} else if TMUX.get() {
 			adapters.retain(|p| *p != Self::KgpOld);
 		}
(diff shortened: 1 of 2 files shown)


sxyazi:
You don't need a new variable to do that, just run Yazi with:

```sh
ZELLIJ_SESSION_NAME= KITTY_WINDOW_ID=1 yazi
```

should do the trick, which forces Yazi to use Kgp under Zellij. You can also make it an alias if that's desired:

```sh
alias y="ZELLIJ_SESSION_NAME= KITTY_WINDOW_ID=1 yazi"
```

Also, just a mild heads up, any AI-generated code must be explicitly disclosed, and this project doesn't allow PR descriptions written by AI, see https://github.com/sxyazi/yazi/blob/main/CONTRIBUTING.md#ai-policy

luccahuguet:
> Also, just a mild heads up, any AI-generated code must be explicitly disclosed, and this project doesn't allow PR descriptions written by AI, see https://github.com/sxyazi/yazi/blob/main/CONTRIBUTING.md#ai-policy

oh, i was not aware! 

> ZELLIJ_SESSION_NAME= KITTY_WINDOW_ID=1 yazi

thanks, that will do

luccahuguet:
> In any case, thanks for your effort!

no, thank you

luccahuguet:
quick note: your proposed fix worked well

it did break some zellij processes that yazelix starts from yazi, 

but after fixing that its all good

PR #4080, opened 2026-06-24
feat: context-aware icons for inputs

sxyazi:
This PR adds icons to the input prompts for `create`, `rename`, `shell`, and interactive `cd` actions:

- Better distinguish file types and avoid typos
- For `create` and `cd`, better indicate whether you're creating/cd-ing to a file or a directory
- Respects user-configured file icons instead of hardcoding. If the user has icons disabled, the input won't show any icons either

https://github.com/user-attachments/assets/2dad4071-ce09-4a45-9add-16c2a05a1ee8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/):
 - Bulk create ([#3793])
 - Make help menu a command palette ([#4074])
 - H/M/L Vim-like motion for moving cursor relative to viewport ([#3970])
+- Context-aware icons for inputs ([#4080])
 - Dynamic keymap Lua API ([#4031])
 - New `ui.Input` element ([#4040])
 - Image preview with Überzug++ on Niri ([#3990])
@@ -1763,3 +1764,4 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/):
 [#4067]: https://github.com/sxyazi/yazi/pull/4067
 [#4068]: https://github.com/sxyazi/yazi/pull/4068
 [#4074]: https://github.com/sxyazi/yazi/pull/4074
+[#4080]: https://github.com/sxyazi/yazi/pull/4080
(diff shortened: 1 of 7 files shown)


hankertrix:
Would you be open to adding a flag to customise the create behaviour?

In [my plugin](https://github.com/hankertrix/augment-command.yazi#create-create), I reimplement the create command to match the default command as much as possible, but it's not perfect, and with this change not being exposed in the `ya.input` function, there is no way to match this.

What my plugin's version of the create function does is change the default behaviour to create directories by default, unless there is a file extension given, then a file is created. If there is a trailing slash or backslash, it will still be a directory. I'm happy to open a PR to add this behaviour as a flag passed to the create function, maybe something like `--default-dirs` or something (not very sure what to call the flag currently).

sxyazi:
I plan to make it available in some way for `ya.input` to let plugins customize their icon behavior, but not currently sure how to design the API generically enough yet. Do you have any ideas?

hankertrix:
Ah I see. I think the most generic way to implement such a thing is a function call, while providing the default function to use if none is given, or having it be explicit and expose the default function.

But it can also get pretty annoying if plugin developers just want to make a tiny change to the default function, or just add another icon, so I think having that function take a mapping of Lua patterns to the icon, and exposing the default mapping for merging additional icons might also work.

I'm not sure if this is generic enough for most use cases, but this is what I came up with.

sxyazi:
Would you like to open a feature request for adding icon support to `ya.input` where you can propose your solution and we can discuss it further there.

hankertrix:
Okay, sure thing!

neelspoc32:
how do you enable this, i don't see it ? 
I'm on the new version.

image

dw-xdy:
@neelspoc32 Hey, this feature is only available in the nightly version, not in 26.5.6. First, check your Yazi version — could you share the output of yazi --debug (or ya env if you're on nightly)?

neelspoc32:
@dw-xdy ok i see. I don't think i'm on nightly version. Yazi
    Version  : 26.5.6 (Arch Linux 2026-06-15)
    Debug    : false
    Triple   : x86_64-unknown-linux-gnu (linux-x86_64)

Issue #4082, opened 2026-06-26
Add new icons support to `ya.input`


hankertrix:
### `ya env` output

```Shell
Yazi
    Version: 26.5.6 (ef654f44 2026-06-24)
    Debug  : false
    Triple : x86_64-unknown-linux-gnu (linux-x86_64)
    Rustc  : 1.96.0 (ac68faa2 2026-05-25)
    Backtrace: None

Ya
    Version: 26.5.6 (ef654f44 2026-06-24)
    Debug  : false
    Triple : x86_64-unknown-linux-gnu (linux-x86_64)
    Rustc  : 1.96.0 (ac68faa2 2026-05-25)

Config
    Init             : /home/hanker/.config/yazi/init.lua (236 chars)
    Yazi             : /home/hanker/.config/yazi/yazi.toml (1632 chars)
    Keymap           : /home/hanker/.config/yazi/keymap.toml (9885 chars)
    Theme            : /home/hanker/.config/yazi/theme.toml (739 chars)
    VFS              : /home/hanker/.config/yazi/vfs.toml (No such file or directory (os error 2))
    Package          : /home/hanker/.config/yazi/package.toml (445 chars)
    Dark/light flavor: ArcSwapAny("bluloco-dark") / ArcSwapAny("bluloco-light")

Emulator
    TERM                : Some("xterm-kitty")
    TERM_PROGRAM        : None
    TERM_PROGRAM_VERSION: None
    Brand.from_env      : Some(Kitty)
    Emulator.detect     : Emulator { kind: Left(Kitty), version: "kitty(0.47.1)", light: true, csi_16t: (14, 32), force_16t: false }

Adapter
    Adapter.matches    : Kgp
    Dimension.available: Dimension { cols: 89, rows: 47, width: 1246, height: 1504 }

Desktop
    XDG_SESSION_TYPE           : Some("wayland")
    WAYLAND_DISPLAY            : Some("wayland-1")
    DISPLAY                    : Some(":0")
    NIRI_SOCKET                : Some("/run/user/1000/niri.wayland-1.2558.sock")
    SWAYSOCK                   : None
    HYPRLAND_INSTANCE_SIGNATURE: None
    WAYFIRE_SOCKET             : None

SSH
    shared.in_ssh_connection: false

WSL
    WSL: false

Variables
    SHELL              : Some("/bin/bash")
    EDITOR             : Some("nvim")
    VISUAL             : Some("nvim")
    YAZI_FILE_ONE      : None
[...]

PR #4100, opened 2026-07-04
fix: input icon spacing/create --dir detection, and add show_icons/truncate_paths options

GrzegorzKozub:
## Summary

Follow-ups to #4080 (context-aware icons for inputs) and #4096 (icons on trash/delete/overwrite confirmations):

- The icon rendered in the input popup's bottom title always had a trailing space baked into its text (`format!("{} ", i.text)`), which made it visually inconsistent with the title text above it, which has no trailing space. Removed it.

image

- `create --dir` (the dedicated "create directory" action) always showed the file icon while typing. The icon lookup only distinguishes file vs. dir by checking whether the currently typed value ends with a path separator — a convention that only applies to the plain `create` action (file-or-dir depending on a trailing `/`). Since `create`'s `InputOpt` didn't carry the `dir` flag anywhere past choosing the title, the icon code had no way to know `--dir` was passed. Gave it its own `create-dir` id (mirroring the existing `rename-file`/`rename-dir` split) so it always resolves to the directory icon.
- Added a `show_icons` option under `[input]` (enabled by default) so the icon feature can be turned off for input popups specifically, without having to clear all `[icon]` rules and lose file-type icons everywhere else.
- Added `show_icons` (enabled by default) and `truncate_paths` (disabled by default) options under `[confirm]`:
  - `show_icons` independently toggles icons on the trash/delete/overwrite confirmation dialogs.
  - `truncate_paths` RTL-truncates each listed path (with a leading `…`) to fit the popup's own configured width instead of relying on the list's word-wrap — a path has no internal spaces to break on, so a long path previously wrapped onto its own line right after the icon whenever it didn't fit, which reads like a stray newline. Off by default to preserve current behavior.

Defaults:

image

With `confirm.show_icons = false`:

image

With `confirm.truncate_paths = true`:

image

And with both the above flags set as said above: 

image

## AI disclaimer

This PR was developed with AI assistance (Claude Code). I reviewed, tested, and am responsible for all of the changes.

## Test plan

- [x] `cargo check -p yazi-fm -p yazi-config` passes
- [x] `cargo fmt --check` shows no diff for the changed files
- [x] Manually verify: `create` (file/dir input, e.g. bound to `A`) still switches icon live based on a trailing `/`
- [x] Manually verify: `create --dir` (e.g. bound to `a`) shows the directory icon immediately, with no trailing space after any input icon
- [x] Manually verify: setting `show_icons = false` under `[input]` hides icons on all input popups (cd/create/rename/shell), and the default (`true`) preserves current behavior
- [x] Manually verify: setting `show_icons = false` under `[confirm]` hides icons on trash/delete/overwrite dialogs
- [x] Manually verify: setting `truncate_paths = true` under `[confirm]` shows a truncated `…`-prefixed path on one line for long paths instead of wrapping to a second line

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/):
 - Make help menu a command palette ([#4074])
 - H/M/L Vim-like motion for moving cursor relative to viewport ([#3970])
 - Context-aware icons for inputs ([#4080])
+- New `show_icons` option under `[input]` to toggle icons on input popups ([#4100])
 - Show file icons in trash/delete/overwrite confirmations ([#4096])
 - Dynamic keymap Lua API ([#4031])
 - New `ui.Input` element ([#4040])
@@ -1768,3 +1769,4 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/):
 [#4074]: https://github.com/sxyazi/yazi/pull/4074
 [#4080]: https://github.com/sxyazi/yazi/pull/4080
 [#4096]: https://github.com/sxyazi/yazi/pull/4096
+[#4100]: https://github.com/sxyazi/yazi/pull/4100
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/):
 - H/M/L Vim-like motion for moving cursor relative to viewport ([#3970])
 - Context-aware icons for inputs ([#4080])
 - New `show_icons` option under `[input]` to toggle icons on input popups ([#4100])
+- New `show_icons` and `truncate_paths` options under `[confirm]` to toggle icons and RTL-ellipsis-truncate long paths on trash/delete/overwrite confirmations ([#4100])
 - Show file icons in trash/delete/overwrite confirmations ([#4096])
 - Dynamic keymap Lua API ([#4031])
 - New `ui.Input` element ([#4040])
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4882,6 +4882,7 @@ dependencies = [
  "serde_with",
  "toml",
  "tracing",
+ "unicode-width",
  "yazi-binding",
  "yazi-codegen",
  "yazi-fs",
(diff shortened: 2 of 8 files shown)


sxyazi:
This PR bundles several features, and some of them like `truncate_paths`, `show_icons` I'm not sure if I want to support or whether they're reasonable design-wise yet. 

I'm still on the fence about whether to surface an "icon" concept or just expose the whole subtitle, and that'll affect how these options are named too. There's more discussion about it in https://github.com/sxyazi/yazi/issues/4082

It's better to discuss those features separately in issues first.

> - The icon rendered in the input popup's bottom title always had a trailing space baked into its text (`format!("{} ", i.text)`), which made it visually inconsistent with the title text above it, which has no trailing space. Removed it.

That was to comply kitty's symbol requirement, https://sw.kovidgoyal.net/kitty/faq/#some-special-symbols-are-rendered-small-truncated-in-kitty, otherwise those icons get rendered tiny and look off next to other icons (like the file list ones).

> - `create --dir` (the dedicated "create directory" action) always showed the file icon while typing. The icon lookup only distinguishes file vs. dir by checking whether the currently typed value ends with a path separator — a convention that only applies to the plain `create` action (file-or-dir depending on a trailing `/`). Since `create`'s `InputOpt` didn't carry the `dir` flag anywhere past choosing the title, the icon code had no way to know `--dir` was passed. Gave it its own `create-dir` id (mirroring the existing `rename-file`/`rename-dir` split) so it always resolves to the directory icon.

Should be fixed in https://github.com/sxyazi/yazi/commit/c92c4aba884c16d318a92f6bc0517ee0d2bcc414.

GrzegorzKozub:
Thanks @sxyazi. No problems at all. Let me know if you like any of these changes and how you like me to provide them. At the minimum, I would really appreciate an ability to disable the icons. Maybe it's worth considering a toggle for any upcoming feature so that Yazi can be even more customizable 😃

sxyazi:
Indeed, I did consider that when adding icon support for the input and confirm - if the user [has icons disabled](https://yazi-rs.github.io/docs/faq#dont-like-nerd-fonts), the input won't show either. But from my understanding, that doesn't quite do what you need and it seems like you're looking to disable icons for the input and confirm locally but keep anything else enabled right? 

Not sure how common that is yet, but it might be worth making it a feature request to see if others also need it, and explore if there's a more flexible way to achieve it there.

Either way, I'm closing this PR for now in favor of discussing them in separate issues. Thanks for your effort!

PR #4105, opened 2026-07-07
Add a read-only rclone VFS provider (`rclone://`)

odfalik:

Addresses the "Integrate rclone" item in #3395 (partial — it doesn't fully
resolve the issue).

## Rationale of this PR

Adds an `rclone://<remote>/<path>` VFS scheme that shells out to the `rclone`
CLI, so yazi can browse any remote rclone knows about (S3, GCS, Azure, B2, SFTP,
WebDAV, Google Drive, local, and the rest) with no FUSE/NFS mount.

It reuses the existing VFS layer, the same plumbing behind the built-in SFTP
provider. A new `SchemeKind::Rclone` threads through the scheme/URL types, mostly
folded into the existing `Sftp` arms since rclone remotes share the same traits
(remote, virtual, Unix-style byte paths).

Configure a service against an existing rclone remote:

```toml
[services.gcs]
type   = "rclone"
remote = "gcs"        # any remote from `rclone listremotes`
```

Then browse `rclone://gcs//my-bucket/some/dir`. An empty path lists the remote's
buckets.

This is a draft, opened to check the direction before I polish. It's read-only
for now.

### How it works

rclone has no persistent open-file handle, so `Provider::File` (which needs
`AsyncRead + AsyncSeek`) streams `rclone cat` from the current position: a seek
kills the child and the next read respawns `rclone cat --offset <n>`. That's
enough for the previewer and for the parallel chunked copier, which opens several
handles and seeks each to its 8 MiB slice.

- Listing: `rclone lsjson`
- Metadata: `rclone lsjson --stat` (POSIX mode is synthesized, since object
  stores have no permissions)
- `remote:/path` keeps its leading slash, so filesystem-like backends (local,
  sftp) work next to object stores

### Tests

- `tests/rclone_local.rs`, hermetic. Points an rclone `local` remote at a temp
  dir, so it needs no cloud account or credentials, and skips itself if `rclone`
  isn't on PATH (CI without rclone stays green). Covers listing, read, seek,
  missing-object, and a >8 MiB parallel multi-chunk copy with position-varying
  bytes.
- `tests/rclone_read.rs`, live and `#[ignore]`d, pointed at a real remote via env
  vars. Nothing hard-coded.

The Test workflow installs rclone on the Unix runners for the hermetic tests.
Verified end-to-end against live S3 and GCS: byte-exact reads, mid-file seeks,
and both copy paths.

### Questions before I take this further

1. **Is a shell-out-to-rclone provider a direction you'd take upstream at all**,
   or would you rather see native backend crates? That decides whether this is
   worth polishing.
2. **Per-op CLI vs a long-lived `rclone rcd` daemon.** Right now each op spawns a
   process, which keeps the code simple. A daemon (living only while yazi is
   open) would cut per-op latency and pool connections the way the SFTP provider
   does. I'm happy either way and wanted your read first.
3. Read-only felt like the right first cut, matching #3843. Writes via
   `rclone rcat`/`copyto` are a clean follow-up once the shape is settled.

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,6 +26,18 @@ jobs:
       - name: Setup Rust toolchain
         run: rustup toolchain install stable --profile minimal
 
+      # For the rclone VFS provider's hermetic tests (Unix only; they drive an
+      # rclone `local` remote, so no cloud account or credentials are needed).
+      - name: Install rclone
+        if: runner.os != 'Windows'
+        shell: bash
+        run: |
+          if [ "$RUNNER_OS" = "macOS" ]; then
+            brew install rclone
+          else
+            sudo apt-get update && sudo apt-get install -y rclone
+          fi
+
       - name: Setup sccache
         uses: mozilla-actions/sccache-action@v0.0.10
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5357,6 +5357,7 @@ dependencies = [
  "parking_lot",
  "russh",
  "serde",
+ "serde_json",
  "tokio",
  "toml",
  "tracing",
--- a/RCLONE_VFS.md
+++ b/RCLONE_VFS.md
@@ -1,0 +1,94 @@
+# rclone VFS provider for yazi
+
+Browse **any of rclone's 100+ remotes** — S3, GCS, Azure, Backblaze, SFTP,
+WebDAV, Google Drive, … — natively inside [yazi](https://github.com/sxyazi/yazi),
+with no FUSE/NFS mount. It rides on yazi's existing VFS layer (the same
+infrastructure behind the built-in SFTP provider) and shells out to the `rclone`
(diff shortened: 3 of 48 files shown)


odfalik:
Closing this in favor of the plugin approach.

#4118 answers the main design question I raised here. Now that VFS engines are Lua plugins, a native `SchemeKind::Rclone` in core is the wrong shape. This branch also conflicts with the `provider/` → `engine/` refactor, so a native PR isn't the way to go.

I'll build `rclone://` as a `scope` plugin instead, following `vfs-demo.yazi` and your "what's next" note. The read path from this branch carries straight over: `lsjson` for listing and metadata, streaming `rclone cat --offset` with seek implemented by respawning the child, and the `remote:/path` leading-slash handling so filesystem-like backends work alongside object stores. Thanks for the direction, will follow up in the plugins repo.

PR #4126, opened 2026-07-14
ci: pin winget-releaser to a full commit SHA

kobihikri:
## What
Pin `vedantmgoyal9/winget-releaser` in `publish.yml` from `@main` to the **v2** release commit (`4ffc788`).

## Why
The `winget` job hands this action `token: ${{ secrets.WINGET_TOKEN }}` (a publishing credential) on every release. Because `@main` is a moving branch, the exact code that receives that token can change without any change here. Pinning to the v2 commit fixes it to a reviewed version; behaviour is unchanged.

I used AI assistance to help identify this and draft the change; I verified every line against the live workflow myself and take responsibility for it.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -16,7 +16,7 @@ jobs:
           merge-multiple: true
 
       - name: Publish to Winget
-        uses: vedantmgoyal9/winget-releaser@main
+        uses: vedantmgoyal9/winget-releaser@4ffc7888bffd451b357355dc214d43bb9f23917e # v2
         with:
           identifier: sxyazi.yazi
           installers-regex: 'yazi-(x86_64|aarch64)-pc-windows-msvc\.zip$'


kobihikri:
Closing this one myself.

The checklist needs me to tick "I confirm this PR follows the AI Policy". Clause 1 says descriptions have to be written by a human, and clause 2 says I have to name the model I used. Mine was AI-drafted and didn't name it. I could go back and rewrite the description now so the tick becomes true, but that felt like working around the policy rather than following it, so I'd rather just close.

Leaving the finding here in case it's useful to anyone: `publish.yml` passes `WINGET_TOKEN` to `vedantmgoyal9/winget-releaser@main`, so whatever happens to be on that branch at release time is what receives the token. Pinning it to the v2 commit `4ffc788` fixes that. Anyone's welcome to pick it up, no credit needed.

Thanks for having a clear AI policy — it's the reason I stopped and read it instead of guessing.

PR #4204, opened 2026-08-03
feat: new `trash` spotter to show original and backing paths of trashed items

sxyazi:
Follow-up for https://github.com/sxyazi/yazi/pull/4144

VHx6AKL3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/):
 ### Added
 
 - Drag and drop ([#4005])
-- Trash bin ([#4144])
+- Trash bin ([#4144], [#4204])
 - Bulk create ([#3793])
 - Make help menu a command palette ([#4074])
 - Input history ([#4104])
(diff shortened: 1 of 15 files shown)


apprenti-sorcier:
Thank you for the recycle bin implementation in yazi that was missing until now; it makes it now the only necessary file manager on your computer.

However, one thing I find missing is the ability to sort the files in the order (or rather reverse order) of their deletion. If you have files that have accumulated for weeks or months, you are more likely to want to restore a newly deleted file rather than one that you haven’t touched in a while. This is a constant need when I used other file managers and all of them that I know on every OS can sort the most recent deleted files. Unfortunately none of the existing sort possibilities yet implemented in yazi (by file birth or modification timestamp) works the same in the recycle bin. Especially if you have several files with the same name, you are more likely to want to restore the last deleted one (or maybe the oldest); and the other timestamps might be misleading.

Maybe it can be also added as a spotter information, but the main need is to have it as a sortable column.

P.S. As for sorting files… randomly, I don’t find any case where it’s useful, unless you are bored. Maybe it was an Easter egg implemented in the software?

PR #4206, opened 2026-08-04
feat: add OSC 7

### LINKED ISSUE #3110: CWD on ghostty new tab/split

Linux Wayland

Ghostty 1.1.3

### `yazi --debug` output

```Shell
Yazi
    Version: 25.6.11 (Crates.io 2025-08-26)
    Debug  : false
    Rustc  : 1.89.0 (29483883 2025-08-04)

Ya
    Version: 25.6.11 (Crates.io 2025-08-26)

Emulator
    TERM                : Some("screen-256color")
    TERM_PROGRAM        : Some("ghostty")
    TERM_PROGRAM_VERSION: Some("1.1.3")
    Brand.from_env      : Some(Ghostty)
    Emulator.detect     : Emulator { kind: Left(Ghostty), light: false, csi_16t: (10, 22), force_16t: false }

Adapter
    Adapter.matches    : Kgp
    Dimension.available: Dimension { rows: 61, columns: 255, width: 2554, height: 1358 }

Desktop
    WAYLAND_DISPLAY            : Some("wayland-0")

SSH

WSL
    WSL: false

Variables
    EDITOR          : Some("hx")
    VISUAL          : None
    YAZI_FILE_ONE   : None
    YAZI_CONFIG_HOME: None
    YAZI_ZOXIDE_OPTS: None
    FZF_DEFAULT_OPTS: None

Text Opener
    default     : Some(OpenerRule { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for: None, spread: true })
    block-create: Some(OpenerRule { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for: None, spread: true })
    block-rename: Some(OpenerRule { run: "${EDITOR:-vi} \"$@\"", block: true, orphan: false, desc: "$EDITOR", for: None, spread: true })

Multiplexers
    TMUX               : false
    ZELLIJ_SESSION_NAME: None

Dependencies
    file          : 5.45
    ffmpeg/ffprobe: 7.1.1-1 / 7.1.1-1
    pdftoppm      : 25.03.0
    magick        : 7.1.1-43
    fzf           : 0.61.1
    fd/fdfind     : 10.2.0 / 10.2.0
    rg            : 14.1.1
    zoxide        : 0.9.8
    jq            : 1.7

Clipboard
    wl-copy/paste: 2.2.1 / 2.2.1
    xsel         : 1.2.1

Routine


```

### Describe the bug

It's pretty much the same bug as https://github.com/sxyazi/yazi/issues/2042 as far as I can tell... - I'd like to open `yazi`, navigate to some folder and then open a new tab/split in `ghostty`, but the `CWD` is the root directory where I first started `yazi` in. This doesn't happen in `kitty` for example. Don't know if it's a `ghostty` or `yazi` issue either.

### Minimal reproducer

Start `yazi` in `ghostty`, navigate to some sub-folder, open a new tab/split in `ghostty`. Expected behavior is to have the new tab/split opened in the sub-folder, but instead it's the root folder where I first started `yazi`.

### Anything else?

_No response_

### Checklist

- [x] I can reproduce it after disabling all custom configs/plugins (`mv ~/.config/yazi ~/.config/yazi-backup`)

Pivok7:

Resolves: #3110 

## Rationale of this PR

Ghostty and some other terminals don't support chdir detection and rely on OSC 7 for that. I checked that it works with ghostty on my linux machine. I'm not very familiar with the codebase so if there is a better place to put that logic I will try to do that.

## Checklist

### DIFF
--- a/yazi-actor/src/app/title.rs
+++ b/yazi-actor/src/app/title.rs
@@ -2,8 +2,8 @@ use anyhow::Result;
 use yazi_actor::Ctx;
 use yazi_macro::{succ, writef};
 use yazi_parser::{app::TitleForm, spark::SparkKind};
-use yazi_shared::{Source, data::Data};
-use yazi_tty::{TTY, sequence::SetTitle};
+use yazi_shared::{Source, data::Data, url::UrlLike};
+use yazi_tty::{TTY, sequence::{SetTitle, SetWorkingDir}};
 use yazi_tui::RatermState;
 
 use crate::Actor;
@@ -18,6 +18,9 @@ impl Actor for Title {
 	fn act(cx: &mut Ctx, form: Self::Form) -> Result<Data> {
 		let s = form.value.unwrap_or_else(|| format!("Yazi: {}", cx.tab().name()).into());
 		writef!(TTY.writer(), "{}", SetTitle(&s))?;
+
+		let cwd = cx.cwd().as_local().unwrap_or(std::path::Path::new("/"));
+		writef!(TTY.writer(), "{}", SetWorkingDir(cwd))?;
 
 		yazi_tui::STATE.set(RatermState { title: !s.is_empty(), ..yazi_tui::STATE.get() });
 		succ!()
--- a/yazi-tty/src/sequence/cursor.rs
+++ b/yazi-tty/src/sequence/cursor.rs
@@ -44,6 +44,15 @@ impl Display for SetTitle<'_> {
 	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result { write!(f, "\x1b]2;{}\x1b\\", self.0) }
 }
 
+/// Set current working directory via OSC 7
+pub struct SetWorkingDir<'a>(pub &'a std::path::Path);
+
+impl Display for SetWorkingDir<'_> {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		write!(f, "\x1b]7;file://localhost{}\x07", self.0.display())
+	}
+}
+
 /// Set cursor style
 #[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
 pub enum SetCursorStyle {


sxyazi:
Thanks for the PR.

I'm not sure I want to support this. Last time I [tested](https://github.com/sxyazi/yazi/issues/3110#issuecomment-3224791315) it, only Ghostty *mandatorily* supported it, while other terminals either don't support it at all, or support both it and `chdir` at the same time, so it's unclear to me:

- Is there anything else besides Ghostty that *strictly* requires it right now?
- How's this protocol supposed to behave for Windows paths? Since Windows paths can't be fully represented as a URI.
- There's also a competing protocol `OSC 9;9` - why go with `OSC 7` instead of that?
- Right now Yazi does an unconditional `chdir`, would there be any issues if `chdir` and `OSC 7` are both used at the same time?

Also, Yazi supports plugins, so you can already easily implement this yourself with just a few lines - just put this into your `init.lua`:

```lua
ps.sub("cd", function()
  local cwd = tostring(cx.active.current.cwd)
  io.write("\x1b]7;file://localhost" .. cwd .. "\x07")
  io.flush()
end)
```

Pivok7:
It's a shame that terminal world is so messy. I also checked that zellij terminal multiplexer doesn't detect chdir.

I took a look at nushell (shell made in rust) and they use OSC 9;9 on windows and OSC 7 otherwise to fix this issue:

```nu
# shell_integration.osc7 (bool): Report current directory to terminal using OSC 7.
# Enables opening new tabs/windows in the same directory.
# Default: true (false on Windows)
$env.config.shell_integration.osc7 = ($nu.os-info.name != windows)
```

```
# shell_integration.osc9_9 (bool): Enable OSC 9;9 support (ConEmu/Windows Terminal feature).
# Alternative to OSC 7 for communicating current path.
# Default: false (true on Windows)
$env.config.shell_integration.osc9_9 = ($nu.os-info.name == windows)
```

Seems like there is no straightforward way of ensuring the terminal reads cwd. I suppose using plugin is a good fix if someone needs this feature

Pivok7:
also this code doesn't seem to work :

```lua
ps.sub("cd", function()
  local cwd = tostring(cx.active.current.cwd)
  io.write("\027]7;file://" .. cwd .. "\027\\")
  io.flush()
end)
```

this one does the trick

```lua
ps.sub("cd", function()
  local cwd = tostring(cx.active.current.cwd)
  io.write("\x1b]7;file://localhost" .. cwd .. "\x1b\\")
  io.flush()
end)
```

sxyazi:
Yeah, that's exactly why I don't really want to bake these into Yazi directly - these protocols are pretty platform/terminal-specific right now, and there still doesn't seem to be a widely-accepted, well-designed protocol out there.

Last time I tested Zellij it supported neither `OSC 7` nor `chdir`, not sure if that's changed by now.

Glad that works for you - I've updated the https://github.com/sxyazi/yazi/pull/4206#issuecomment-5182492721 with your code, thanks! Might be worth publishing it as a small plugin, maybe drop it into [yazi-rs/plugins](https://github.com/yazi-rs/plugins) as an experimental thing, where we could try supporting both `OSC 7` and `OSC 9;9`.

Pivok7:
Sure, I will try to submit a plugin and also test if it works on windows

PR #4268, opened 2026-08-20
fix: correct open rule matching for trashed directories

olibrks:
## Context
This PR builds upon the folder mime-types updates in #4212 and the trash bin features #4144  to fix mismatch between folder and trash rules in yazi.toml.

* **Trash Folder mime-type** - Changed `folder/trash` mime-type  to `trash/folder`. This ensures trash directories are correctly caught by the `mime = "trash/**"` rules introduced in #4144.
* **Default Folder Rules** - Replaced `url = "*/"` folder rules with `mime = "folder/*"  to avoid mismatch between trash rules and folder rules.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/):
 ### Fixed
 
 - Assertion failure on expanding URLs with variables containing absolute paths ([#4256])
+- Correct open rule matching for trashed directories ([#4268])
 - Wait for terminal probe echo back before stopping instance ([#4271])
 - Avoid flicker caused by screen clear on final response from terminal ([#4250])
 - Fall back when `vergen` cannot determine Git SHA ([#4252])
@@ -1835,4 +1836,5 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/):
 [#4256]: https://github.com/sxyazi/yazi/pull/4256
 [#4260]: https://github.com/sxyazi/yazi/pull/4260
 [#4265]: https://github.com/sxyazi/yazi/pull/4265
+[#4268]: https://github.com/sxyazi/yazi/pull/4268
 [#4271]: https://github.com/sxyazi/yazi/pull/4271
--- a/yazi-config/preset/yazi-default.toml
+++ b/yazi-config/preset/yazi-default.toml
@@ -71,7 +71,7 @@ trash = [
 [open]
 rules = [
 	# Folder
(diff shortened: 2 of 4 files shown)


sxyazi:
Thanks for the patch, but the change doesn't make any sence - all folder's MIME type should starts with `folder/` including VFSs, see https://github.com/sxyazi/yazi/pull/3222

olibrks:
mine-trash.lua plugin adds "trash/" in front of mime-type. Folders are not following that rule (making it folder/trash) preventing the rules mine=trash/* to include folders. So, no trash operations like restore for folders.

sxyazi:
Oh I see what you're saying now, but Yazi checks if the MIME type is prefixed with "folder/" to determine if it's a directory internally and changing this would cause trashed folders to no longer be recognized.

https://github.com/sxyazi/yazi/blob/25d6bd19a9b30acbef84c03110e30a0a631c0737/yazi-config/src/open/open.rs#L54

 A better way may be needed to fix this.

olibrks:
I would argue that a folder in trash doesn't need to follow normal folder rules anymore. This is already the case for files in trash that don't follow mime-type rules anymore(for example, an audio file in trash won't be caught by the rule "mime=audio/*").

sxyazi:
Then how do you tell if a trashed file is a directory or not by its MIME type?

olibrks:
The premise is that trashed files and folders should be targeted differently than normal files and folders:

Normal Files
``` mime = audio/*, video/*, text/*, etc ```
Normal Folders
``` mime = folder/* ```  (identical to ```url = "*/" ```)

Trashed Files & Folders - Global rules for both files and folders to allow multi-selection restore
``` mime = trash/* ```
Trashed Folders - If someone wants to target trashed folders for any reason
``` mime = trash/folder ```
Trashed Files - If someone wants to target trashed files for any reason
``` mime = trash/(audio/*, text/*, etc} ```

If I'm not mistaken, open.rs uses mime-type in case a `url` rule ends with `/`. This is still working for normal folders.
Both `mime="folder/*"` and the identical` url="*/"` rules won't target trashed folders anymore with this PR.

Note: As suggested in this PR, using `mime=folder/*` sounds more intuitive and makes all the open rules use mime-types (except the catch-all `url="*"`). This is however just a style choice.

sxyazi:
That's right, but there's another promise that one can tell if a file is a directory or not by its MIME type (whether it's in VFS or not) and we'll need a way to fix the problem without breaking another. I think changing `open.rs` might be the only sound way here, reopening.

sxyazi (approved):
ty!

